PR #5, opened 2017-11-15
two new functions

Gertian:
* added a zero element on vectorspaces (similar to the eye function) 
* added a function that returns the vectorspace living on some index as if that index were in the codomain


src/tensors/tensor.jl
```diff
@@ -138,6 +147,7 @@ Tensor(P::TensorSpace{S}) where {S<:IndexSpace} = TensorMap(P, one(P))
 # Special purpose constructors
 #------------------------------
 Base.zero(t::AbstractTensorMap) = fill!(similar(t), 0)
+Base.zero(P::Union{IndexSpace,TensorSpace}) = fill!(similar(TensorMap(eye, P←P)),0)
```
Jutho, line 150: That's anyway a suboptimal implementation since you'll first allocate a Tensor `eye`, and than create/allocate another one that is `similar`, and fill that one with zeros.

PR #6, opened 2017-11-29
included LQ using the column permutation method

Gertian:


src/auxiliary/linalg.jl
```diff
@@ -14,7 +14,10 @@ struct QRpos <: OrthogonalFactorizationAlgorithm
 end
 struct QR <: OrthogonalFactorizationAlgorithm
 end
-# TODO: QL and QLpos ?
+struct QL <: OrthogonalFactorizationAlgorithm
+end
+struct LQ <: OrthogonalFactorizationAlgorithm
+end
```
Jutho: LQ seems to be defined twice?

src/auxiliary/linalg.jl
```diff
@@ -33,28 +36,48 @@ end
...
+        Q, R = leftorth!(A, isa(alg, QL) ? QR() : QRpos() )
+        #swap comumns in Q
+        @inbounds for j = 1:nhalf, i = 1:m
+            Q[i,j], Q[i,n+1-j] = Q[i,n+1-j], Q[i,j]
+        end
+        #swap rows and columns in R
+        @inbounds for j = 1:nhalf, i = 1:m
+            R[i,j], R[n+1-i,n+1-j] = R[n+1-i,n+1-j], R[i,j]
+        end
+        return Q, R
```
Jutho: The `isodd(n)` part is missing:
```julia
        if isodd(n)
            j = nhalf+1
            @inbounds for i = 1:nhalf
                R[i,j], R[m+1-i,j] = R[m+1-i,j], R[i,j]
            end
        end
```

src/auxiliary/linalg.jl
Gertian: Oh, that the second one should have been QLpos.

PR #14, opened 2019-04-17
Update representationspace.jl

maartenvd:


Jutho:
? Doesn't this work like it is now? However, I agree that I should certainly look into simplifying the number of different constructors.

maartenvd:
No it doesn't, because the constructor does not expect a keyword argument
(making the benchmark fail). Either we globally use keywords or we use
optional arguments.

Jutho:
But both of these work, no?
```julia
julia> V1 = SU2Space((i=>i+1 for i=0:3); dual = true)
SU₂Space(0=>1, 1=>2, 2=>3, 3=>4)'

julia> V2 = SU2Space((i=>i+1 for i=0:3))'
SU₂Space(0=>1, 1=>2, 2=>3, 3=>4)'

julia> V1 == V2
true
```

maartenvd:
that's odd, because I got:

```
ERROR: LoadError: MethodError: no method matching U₁Space(::Pair{U₁,Float64}, ::Pair{U₁,Float64}, ::Pair{U₁,Float64}, ::Pair{U₁,Float64}, ::Pair{U₁,Float64}, ::Pair{U₁,Float64}, ::Pair{U₁,Float64}; dual=false)
Closest candidates are:
  U₁Space(::Any, ::Any) where G<:Sector at /home/maavdamm/.julia/packages/TensorKit/Y6AiJ/src/spaces/representationspace.jl:9 got unsupported keyword argument "dual"

```

when running the benchmark, and assumed that the constructor had a problem

Jutho:
It is the `Float64` in the pairs. You probably want to replace
```
            degeneracies = ceil.(DD[jj]*exp.(-log(DD[jj])*(-ii:ii).^2/(ii^2)))
```
with
```
            degeneracies = ceil.(Int, DD[jj]*exp.(-log(DD[jj])*(-ii:ii).^2/(ii^2)))
```

maartenvd:
ah; makes sense :p

Jutho:
Maybe the constructor could try to do the conversion. I'll remember to take a look at all the different constructors.

PR #16, opened 2019-07-27
Typos and typography

jagot:
I fixed some typos and typography, while reading the manual. Please check that I did not mess something up. In particular, in `sectors.md`, I was not sure if it should be ``\mathsf{U}_1`` or ``\mathsf{SU}_1``.


Jutho:
Thanks, this is very helpful. The group is `\mathsf{U}_1`, there is no `\mathsf{SU}_1` (or technically, `\mathsf{SU}_1` would be equivalent to `\mathbb{Z}_2`).

PR #24, opened 2020-01-29
Fix a bug with svd of AdjointTensorMap

mhauru:
In the current master, observe the following:
```
julia> using TensorKit
julia> t = TensorMap(randn, Float64, ℂ^2 ← ℂ^2)'
AdjointTensorMap(ProductSpace(ℂ^2) ← ProductSpace(ℂ^2)):
 -0.44957311339859585  0.574946008374467 
 -1.0585667067396305   0.8048926380999573
julia> svd(t, (1,), (2,))
(0.0, AdjointTensorMap(ProductSpace(ℂ^2) ← ProductSpace(ℂ^2)):
 0.47443890076071277   0.8802884353693206 
 0.8802884353693206   -0.47443890076071277
, AdjointTensorMap(ProductSpace(ℂ^2) ← ProductSpace(ℂ^2)):
 1.508085327421287  0.0                
 0.0                0.16362510057556934
, AdjointTensorMap(ProductSpace(ℂ^2) ← ProductSpace(ℂ^2)):
 -0.7593330317060951  0.650702195293692
  0.650702195293692   0.759333031706095
)
```

The truncation error has been moved to being the first return value. Fixing this is of course a breaking change, but I really hope noone has code that does things like `err, U, S, Vt = svd(t', indsout, indsin)`.

By the way, this only comes up when specifying the index lists. Just calling `svd(t)` incurs a `copy(t)`, which removes the adjoint structure, and thus
```
julia> svd(t)
(TensorMap(ProductSpace(ℂ^2) ← ProductSpace(ℂ^2)):
 -0.47443890076071293  -0.8802884353693207 
 -0.8802884353693206    0.47443890076071293
, TensorMap(ProductSpace(ℂ^2) ← ProductSpace(ℂ^2)):
 1.5080853274212869  0.0                
 0.0                 0.16362510057556925
, TensorMap(ProductSpace(ℂ^2) ← ProductSpace(ℂ^2)):
  0.7593330317060952  -0.6507021952936921
 -0.6507021952936921  -0.7593330317060952
, 0.0)
```

I also included a test that catches this, by trying all the decomposition tests also on `AdjointTensorMap`s.

### DIFF
--- a/src/tensors/factorizations.jl
+++ b/src/tensors/factorizations.jl
@@ -221,7 +221,8 @@ function rightnull!(t::AdjointTensorMap{S}; alg::OrthogonalFactorizationAlgorith
     return adjoint(leftnull!(adjoint(t); alg = alg'))
 end
 function LinearAlgebra.svd!(t::AdjointTensorMap{S}; trunc::TruncationScheme = NoTruncation(), p::Real = 2, alg::Union{SVD,SDD} = SDD()) where {S<:EuclideanSpace}
-    return map(adjoint, reverse(svd!(adjoint(t); trunc = trunc, p = p, alg = alg)))
+    u, s, vt, err = svd!(adjoint(t); trunc = trunc, p = p, alg = alg)
+    return map(adjoint, (vt, s, u, err))
 end
 
 function leftorth!(t::TensorMap{S}; alg::OrthogonalFactorizationAlgorithm = QRpos()) where {S<:EuclideanSpace}
--- a/test/tensors.jl
+++ b/test/tensors.jl
@@ -214,38 +214,41 @@ for (G,V) in ((Trivial, Vtr), (ℤ₂, Vℤ₂), (ℤ₃, Vℤ₃), (U₁, VU₁
     @testset "Factorization" begin
         W = V1 ⊗ V2 ⊗ V3 ⊗ V4 ⊗ V5
         for T in (Float32, Float64, ComplexF32, ComplexF64)
-            t = Tensor(rand, T, W)
-            @testset "leftorth with $alg" for alg in (TensorKit.QR(), TensorKit.QRpos(), TensorKit.QL(), TensorKit.QLpos(), TensorKit.Polar(), TensorKit.SVD(), TensorKit.SDD())
-                Q, R = @inferred leftorth(t, (3,4,2),(1,5); alg = alg)
-                QdQ = Q'*Q
-                @test QdQ ≈ one(QdQ)
-                @test Q*R ≈ permuteind(t, (3,4,2),(1,5))
-            end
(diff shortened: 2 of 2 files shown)


src/tensors/factorizations.jl
```diff
@@ -221,7 +221,8 @@ function rightnull!(t::AdjointTensorMap{S}; alg::OrthogonalFactorizationAlgorith
     return adjoint(leftnull!(adjoint(t); alg = alg'))
 end
 function LinearAlgebra.svd!(t::AdjointTensorMap{S}; trunc::TruncationScheme = NoTruncation(), p::Real = 2, alg::Union{SVD,SDD} = SDD()) where {S<:EuclideanSpace}
-    return map(adjoint, reverse(svd!(adjoint(t); trunc = trunc, p = p, alg = alg)))
+    u, s, vt, err = svd!(adjoint(t); trunc = trunc, p = p, alg = alg)
+    return map(adjoint, (vt, s, u, err))
```
Jutho, line 225: Maybe `map(adjoint, (vt, s, u))..., err`. There is no harm in taking the `adjoint` of `err`, but also no need. 

Also, while writing the documentation, I started to second guess my decision of returning the arguments as `u`, `s` and `vt`, while for Julia Base (and Matlab),  `svd` returns `u`, `s` and `v`. While typically `vt` is more useful, it does seem against the style / guidelines to overload a Julia Base function and then have it do something (slightly) different.

Jutho:
Great change; thanks for catching this.

src/tensors/factorizations.jl
mhauru: Yeah, not adjointing `err` would indeed be cleaner.

I see the point in following Julia Base, even though I wish they had gone with a different convention (the NumPy one) themselves. I'm happy to accommodate my code if you decide to make the change, but it could cause some nasty silent breakage for some people, if they are SVDing matrices and `v` and `vt` have the same shape and thus you can replace one with the other without anything exploding. Then again, I don't know how much stability you even want to guarantee at this time.

Jutho: Hi Markus, I already merged your PR so don't bother. I might change this later. 

I didn't know that Python did return `vt` or, as they called (better), `vh`. That's great. I guess there are more people coming from Python than from Matlab. So maybe the right course of action here is for `TensorKit` to export `svd` as an independent method from `LinearAlgebra.svd`, and make it unnecessary for people to ever do `using LinearAlgebra` in combination with `using TensorKit`. This than matches with how  TensorKit exports `eig` and `eigh`, aside from also overloading the type-unstable method `LinearAlgebra.eigen`. The narrative is then that TensorKit pretends to be closer to Numpy than to Matlab or to Julia's `LinearAlgebra` module.

PR #27, opened 2020-04-10
Add Dict and Pair constructors for Cartesian and Complex spaces

mhauru:
I changed the code a bit from what we discussed earlier, to not have the method signatures be too restricted.


Jutho:
Ok looks good; my apologies for not merging last week. I forgot about it.

PR #30, opened 2020-04-15
Add `real` and `imag`.

mhauru:


src/tensors/tensor.jl
```diff
@@ -410,6 +410,34 @@ function Base.show(io::IO, t::TensorMap{S}) where {S<:IndexSpace}
...
+function Base.real(t::AbstractTensorMap{S}) where {S}
+    # `isreal` for a `Sector` returns true iff the F and R symbols are real. This guarantees
+    # that the real/imaginary part of a tensor `t` can be obtained by just taking
+    # real/imaginary part of the degeneracy data.
+    if isreal(sectortype(S))
+        realdata = Dict(k => real(v) for (k, v) in blocks(t))
+        return TensorMap(realdata, codomain(t), domain(t))
+    else
+        msg = "`real` has not been implemented for `AbstractTensorMap{$(S)}`."
+        throw(msg)
```
Jutho: I think this needs to be `throw(ArgumentError(msg))`; also in `imag`

mhauru:
The way the tests are set up will fail if run on a sector type that `!isreal`. Not sure if this is desirable.

src/tensors/tensor.jl
mhauru: Changed now.

PR #33, opened 2020-05-28
update tensoroperations.jl

maartenvd:
use the new threads.spawn?

### DIFF
--- a/src/tensors/tensoroperations.jl
+++ b/src/tensors/tensoroperations.jl
@@ -130,15 +130,9 @@ function add!(α, tsrc::AbstractTensorMap{S}, β, tdst::AbstractTensorMap{S,N₁
         K = Threads.nthreads()
         if K > 1
             let iterator = fusiontrees(tsrc)
-                Threads.@threads for k = 1:K
-                    counter = 0
-                    for (f1,f2) in iterator
-                        counter += 1
-                        if mod1(counter, K) == k
-                            _addabelianblock!(α, tsrc, β, tdst, p1, p2, f1, f2)
-                        end
-                    end
-                end
+                fetch.(map(iterator) do (f1,f2)
+                    @Threads.spawn _addabelianblock!(α, tsrc, β, tdst, p1, p2, f1, f2)
+                end)
             end
         else # debugging is easier this way
             for (f1,f2) in fusiontrees(tsrc)


src/tensors/tensoroperations.jl
```diff
@@ -130,15 +130,9 @@ function add!(α, tsrc::AbstractTensorMap{S}, β, tdst::AbstractTensorMap{S,N₁
...
-                    for (f1,f2) in iterator
-                        counter += 1
-                        if mod1(counter, K) == k
-                            _addabelianblock!(α, tsrc, β, tdst, p1, p2, f1, f2)
-                        end
-                    end
-                end
+                fetch.(map(iterator) do (f1,f2)
+                    @Threads.spawn _addabelianblock!(α, tsrc, β, tdst, p1, p2, f1, f2)
+                end)
```
Jutho, line 135: probably 
```julia
wait.(map(iterator) do (f1,f2)
                     @Threads.spawn _addabelianblock!(α, tsrc, β, tdst, p1, p2, f1, f2)
                 end)
```
or
```julia
for (f1,f2) in fusiontrees(tsrc)
                     @Threads.spawn _addabelianblock!(α, tsrc, β, tdst, p1, p2, f1, f2)
end
@sync
```

Jutho: But I also have no experience with this...

maartenvd: ah yes, wait would've been better

maartenvd: it is very plausible that fetch..(map) or wait.map() gets optimized to wait(@spawn) for every block, making it effectively sequential. I'm testing that now

Jutho: I think the `@sync` after the `@spawn` in the loop is a common paradigm, so that should hopefully work.

maartenvd: @sync was what I tried, and it failed. I think the correct syntax is @sync map() do .... end

Jutho: Yes, I also have misremembered this. I think just
```julia
@sync for (f1,f2) in fusiontrees(tsrc)
@Threads.spawn _addabelianblock!(α, tsrc, β, tdst, p1, p2, f1, f2)
end
```
should be ok. The `let iterator = ` was to avoid the penalty of the anonymous function with captured variable created by the `@threads` macro. With `@sync` and `@spawn` this should not be a problem I believe.

maartenvd: Indeed that works, but maybe this pull request isn't all that necessary, in julia 1.6 @threads will use the @spawn interface anyway

PR #36, opened 2020-06-25
added a flag so that truncbelow includes first value(ie sector) below treshhold

Gertian:


Jutho:
Very clean; looks great!
I'll await CI and then merge.

PR #39, opened 2020-06-30
can now add any number of additional svds after truncbelow

Gertian:
This pull request should be the right one :) !

### DIFF
--- a/src/tensors/truncation.jl
+++ b/src/tensors/truncation.jl
@@ -24,9 +24,9 @@ truncspace(space::ElementarySpace) = TruncationSpace(space)
 
 struct TruncationCutoff{T<:Real} <: TruncationScheme
     ϵ::T
-    addone::Bool
+    add_back::Int
 end
-truncbelow(epsilon::Real, addone::Bool=false) = TruncationCutoff(epsilon, addone)
+truncbelow(epsilon::Real, add_back::Int=0) = TruncationCutoff(epsilon, add_back)
 
 
 # For a single vector
@@ -72,7 +72,9 @@ function _truncate!(v::AbstractVector, trunc::TruncationCutoff, p::Real = 2)
     if dtrunc === nothing
         dtrunc = 0
     end
-    dtrunc < length(v) && trunc.addone && (dtrunc+=1)
+    for i in 1:trunc.add_back
+	dtrunc < length(v) && (dtrunc+=1)
+    end
     truncerr = norm(view(v, dtrunc+1:length(v)), p)
     resize!(v, dtrunc)
     return v, truncerr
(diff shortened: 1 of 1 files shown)


src/tensors/truncation.jl
```diff
@@ -72,7 +72,9 @@ function _truncate!(v::AbstractVector, trunc::TruncationCutoff, p::Real = 2)
     if dtrunc === nothing
         dtrunc = 0
     end
-    dtrunc < length(v) && trunc.addone && (dtrunc+=1)
+    for i in 1:trunc.add_back
+	dtrunc < length(v) && (dtrunc+=1)
+    end
```
Jutho: `dtrunc = min(dtrunc + trunc.add_back, length(v))` as one-liner?

src/tensors/truncation.jl
```diff
@@ -166,21 +168,24 @@ end
...
-	    next_val[c] = v[newdim+1]
+	    next_val[c] = v
 	else
 	    truncdim[c] = newdim
+            next_val[c] = v[newdim+1:end]
         end
     end
-    trunc.addone && (truncdim[findmax(next_val)[2]]+=1)
+    for i in 1:trunc.add_back
+        key_max = findmax(next_val)[2]
```
Jutho: Did you test this? Is this not looking for the maximum of a bunch of vectors, which I think is not defined?

src/tensors/truncation.jl
```diff
@@ -166,21 +168,24 @@ end
...
+	    next_val[c] = v
 	else
 	    truncdim[c] = newdim
+            next_val[c] = v[newdim+1:end]
         end
     end
-    trunc.addone && (truncdim[findmax(next_val)[2]]+=1)
+    for i in 1:trunc.add_back
+        key_max = findmax(next_val)[2] 
+	length(next_val[key_max]) == 0 && break
```
Jutho, line 183: Small alignment issue.

src/tensors/truncation.jl
Jutho: There is also `argmax` as alternative to the second argument of `findmax`

Gertian: Yes I tested this. 
Also expected it to fail but turns out that it it searching for the vector that contains the largest element.

src/tensors/truncation.jl
Gertian: Yes this would be cleaner !

Gertian:
I made and tested your suggestions.

Gertian:
I pushed your suggestions.

src/tensors/truncation.jl
Jutho: Ok, it indeed searches for the one which is largest according to a semantic order. So this works because the singular values are sorted in descending order. Nice. That makes it very easy indeed.

Jutho:
I'll merge once CI has finished.

PR #46, opened 2020-11-05
Fixes to twist and docstrings

mhauru:
I made `twist` check that the index being twisted actually exists (it used to just silently do nothing), and to do nothing for bosonic tensors (it used to error).

I also fixed the docstring for `permute!` (it was just the docstring for `permute`) and for `braid`. Please double check the new docstring for `braid` to make sure I got it right.


src/fusiontrees/manipulations.jl
```diff
@@ -121,12 +121,13 @@ end
...
-applied as braids such that if `i` and `j` cross, ``τ_{i,j}`` is applied if
-`levels[i] < levels[j]` and ``τ_{j,i}^{-1}`` if `levels[i] > levels[j]`. This does not
-allow to encode the most general braid, but a general braid can be obtained by
+specified by specifying that the new sector at position `i` was originally at position
+`p[i]` and assinging to every index a distinct level or depth `levels[p[i]]` (in other
+words, `levels` lists depths of the indices in the order they are in before braiding). This
+permutation is then decomposed into elementary swaps between neighbouring indices, where the
+swaps are applied as braids such that if `i` and `j` cross, ``τ_{i,j}`` is applied if
+`levels[p[i]] < levels[p[j]]` and ``τ_{j,i}^{-1}`` if `levels[p[i]] > levels[p[j]]`. This
+does not allow to encode the most general braid, but a general braid can be obtained by
```
Jutho: This whole explanation is not very accurate (not in the original formulation, but still not in the current formulation I think). It depends on how you interpret `i` and `j`, are they labels for the indices which remain associated with particular indices during the process of elementary swaps, or are they a counting of the indices as they appear in the order of the tensor at any point in time. I mixed both conventions, but I don't think it's completely resolved with the new formulation either. Do you think it is more clear? Maybe something like:

```
The braiding is determined by specifying that the new sector at position `k` corresponds to the sector  that was originally at the position `i = p[k]`, and assigning to every index `i` of the original fusion tree a distinct level or depth `levels[i]`. This permutation is
then decomposed into elementary swaps between neighbouring indices, where the swaps are
applied as braids such that if `i` and `j` cross, ``τ_{i,j}`` is applied if
`levels[i] < levels[j]` and ``τ_{j,i}^{-1}`` if `levels[i] > levels[j]`. This does not
allow to encode the most general braid, but a general braid can be obtained by
combining such operations.
```
```

mhauru: Sure, that formulation work as well, and avoids the ugliness of `levels[p[i]]`. As long as it makes clear whether `levels` is in the order of the indices before or after braiding I'm happy, since I previously couldn't figure that out/got the wrong impression until I just tried it out.

mhauru: See commit below

PR #47, opened 2020-11-05
planar_trace

mhauru:
I wrote a quick version of `planar_trace`. Feel free to discard if you had a different implementation in mind.

### DIFF
--- a/src/tensors/planar.jl
+++ b/src/tensors/planar.jl
@@ -219,7 +219,7 @@ function _add_modules(ex::Expr)
                         (ex.args[i] for i in argind)...)
     elseif ex.head == :call && ex.args[1] == :trace!
         @assert ex.args[4] == :(:N)
-        argind = [2,4,5,6,7,8,9,10]
+        argind = [2,3,4,5,6,7,8,9,10]
         return Expr(ex.head, GlobalRef(TensorKit, Symbol(:planar_trace!)),
                         (ex.args[i] for i in argind)...)
     elseif ex.head == :call && ex.args[1] == :contract!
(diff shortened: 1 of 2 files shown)


src/tensors/planar.jl
```diff
@@ -219,7 +219,7 @@ function _add_modules(ex::Expr)
                         (ex.args[i] for i in argind)...)
     elseif ex.head == :call && ex.args[1] == :trace!
         @assert ex.args[4] == :(:N)
-        argind = [2,4,5,6,7,8,9,10]
+        argind = [2,3,4,5,6,7,8,9,10]
```
Jutho: Change this to `[2,3,5,6,7,8,9,10]`. The `@planar` macro will take out `conj` statements and replace them with adjoints, that's why there is `@assert ex.args[4] == :(:N)`.

This then affects the `planar_trace!` function below, it should not have the conjugation flag as an argument.

src/tensors/planar.jl
```diff
@@ -322,3 +322,67 @@ function planar_contract!(α, A::AbstractTensorMap{S}, B::AbstractTensorMap{S},
...
+        _planar_trace!(α, tsrc, β, tdst, pl, pr, q1, q2)
+    else
+        p = adjointtensorindices(tsrc, (p1..., p2...))
+        pl = TupleTools.getindices(p, codomainind(tdst))
+        pr = TupleTools.getindices(p, domainind(tdst))
+        q1 = adjointtensorindices(tsrc, q1)
+        q2 = adjointtensorindices(tsrc, q2)
+        _planar_trace!(α, adjoint(tsrc), β, tdst, pl, pr, q1, q2)
+    end
+end
```
Jutho: Most of this can go, and `_planar_trace!` can just become `planar_trace!`. Come to think about it, the macro preprocessing does however not check whether traces are compatible with planarity, so this I will need to add.

src/tensors/planar.jl
mhauru: Ah, I was wondering about whether this flag fill actually ever do anything, but I guess I should have read the line above. :)

I'll then remove the split between `planar_trace!` and `_planar_trace!`, because it's ugly and I only did it to deal with `AdjointTensorMap` vs `TensorMap`.

PR #49, opened 2021-01-16
allow for empty blocks in left/right-orth

maartenvd:
at the moment, the following is inconsistent:

```
    t = TensorMap(zeros,ComplexF64,Rep[SU₂](1=>10)*Rep[SU₂](1=>2),Rep[SU₂](1=>10));
    #(l,q) = rightorth(t,alg=LQpos()); #fails

    data = convert(Array,t);
    tt = TensorMap(data,ℂ^size(data,1)*ℂ^size(data,2),ℂ^size(data,3))
    (l,q) = rightorth(tt,alg=LQpos()); #works
```

In this pull request they now work the same way - by making l rank-defficient, while q remains unitary.


Jutho:
Look's good; thanks.

PR #51, opened 2021-05-22
Update planar.jl

maartenvd:


maartenvd:
There are other problems with planar, for example this fails:

```
t1 = TensorMap(rand,ComplexF64,𝔹^2,𝔹^3);
t2 = TensorMap(rand,ComplexF64,(𝔹^2)',(𝔹^3)');
@planar C[] := t1[1;2]*t2[1;2]

```

I don't yet fully understand how planar.jl works, so I'm not yet sure how to fix it

PR #53, opened 2021-08-05
foldright frobeniusschur

lkdvos:
Possibly does not take into account quantum dimension and Frobenius Schur indicator in the case of UniqueFusion, coefficient hardcoded to 1 (end of line 307). This might be generally true for UniqueFusion style, however then probably no need for calculating factor (line 298-301)?


Jutho:
Hi Lukas; you're probably correct. I'll merge as soon as I've completed with the planar and braiding stuff for Maarten.

Jutho:
Thanks for spotting this!

PR #56, opened 2021-08-13
braidingtensor special cases

maartenvd:
some special cases for the braidingtensor action (+ some braidingtensor bugfixes)

also includes special case code for artin_braid, when the leg that is being braided is trivial.


src/tensors/braidingtensor.jl
```diff
@@ -101,6 +107,11 @@ function Base.copy!(t::TensorMap, b::BraidingTensor)
     space(t) == space(b) || throw(SectorMismatch())
     fill!(t, zero(eltype(t)))
     for (f1, f2) in fusiontrees(t)
+        if f1 == nothing || f2 == nothing
+            copy!(t.data,one(t.data))
```
Jutho: can you replace this line with `_one!(t.data)` to avoid allocating a `one(t.data)` first.

Jutho:
Thanks; there were still quite a few bugs in my part apparently :-)

maartenvd:
How would you avoid the inference failures in artin_braid?

src/fusiontrees/manipulations.jl
```diff
@@ -766,7 +768,32 @@ function artin_braid(f::FusionTree{I, N}, i; inv::Bool = false) where {I<:Sector
...
+                inner = TupleTools.setindex(inner,incharges[i-1],i-1);
+            end
+        end
+
+        f′ = FusionTree{I}(uncoupled′, coupled′, isdual′, inner, vertices′)
+        return fusiontreedict(I)(f′ => true)
+    end
+
     oneT = one(eltype(Rsymbol(u,u,u))) * one(eltype(Fsymbol(u,u,u,u,u,u)))
+
```
Jutho, line 802: Move the line of `oneT` to above your `if` clause, and make line 792 into
```julia
fusiontreedict(I)(f′ => oneT)
```

maartenvd: but then Rsymbol(u,u,u) needs to be defined for trivial irreps

Jutho: Ah no, that's the point, it does not exist for NoBraiding()

Jutho: Just make it 
```julia
if BraidingStyle(I) isa NoBraiding
oneT = one(eltype(Fsymbol(u,u,u,u,u,u)))
else
oneT = one(eltype(Rsymbol(u,u,u))) * one(eltype(Fsymbol(u,u,u,u,u,u)))
end
```
above your `if`. Then 
```julia
fusiontreedict(I)(f′ => oneT)
```
and below your `if` clause maybe add
```julia
BraidingStyle(I) isa NoBraiding && throw(SectorMismatch("cannot braid sector "*type_repr(I)))
```
so that the compiler knows that it can never go below that line

Jutho:
Hi Maarten; I did not have the time to review in detail the contraction code and the resolution of resolving all taus by `@planar2`, but a loop is also what I thought. I will merge so that you can continue while I am gone.

maartenvd:
ah, unfortunately I still had an unpushed commit, containing another special case where scalars did not play well with planar, but also a bugfix for some special case braidingtensor actions. It should be the last of it, as I now have a fully planar copy of mpskit, which I guess is a fairly extensive test.

In any case, I don't particularly mind working with a forked copy of tensorkit for the time being - enjoy your holiday first and foremost!

PR #59, opened 2021-09-15
plansor

maartenvd:
An attempt at plansor. Depending on the braiding properties of the contracted tensors, it will either use @tensor or @planar. If @tensor is used, it will remove all braidingtensors from the expression.


Jutho:
Thanks!

PR #60, opened 2021-09-29
typo in man/sectors.md

Jakob-Unfried:
Stumbled upon this small typo while trying to understand your work

### DIFF
--- a/docs/src/man/sectors.md
+++ b/docs/src/man/sectors.md
@@ -110,7 +110,7 @@ There is a graphical representation associated with the fusion tensors and their
 
 As always, we refer to the subsection on [topological data of a unitary fusion category](@ref ss_topologicalfusion) for further details.
 
-Finally, for the implementation, it will be useful to distinguish between an number of
+Finally, for the implementation, it will be useful to distinguish between a number of
 different possibilities regarding the fusion rules. If, for every ``a`` and ``b``, there is
 a unique ``c`` such that ``a ⊗ b = c`` (i.e. ``N^{ab}_{c} = 1`` and ``N^{ab}_{c′} = 0`` for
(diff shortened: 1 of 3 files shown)


Jutho:
Thanks for those. Should I keep this PR open while you continue reading and correcting? 

Note that the manual has become a bit outdated over the past year. I think most of the information it contains is still correct, but recent changes have not been documented. Feel free to ask me any questions you may have.

Jakob-Unfried:
I will read more in the next few days, and will push any further typos here, so keeping it open might be good.

For now, i want to understand the concepts.
Eventually, and provided that i can make time for it, i want to implement either general groups or at least SU(2)
for python (not necessarily IN python, but i will need to understand the data structures better to make such design choices)

Jakob-Unfried:
I am done reading and don't expect to find anything else.
Thanks for making this freely available, it was quite helpful to me.
Feel free to merge, edit or cherrypick.

One more thing i noticed:
I was a bit confused by [this paragraph](https://github.com/Jutho/TensorKit.jl/blob/0b05d1b165019e642062e8998bb65e441152b65c/docs/src/man/sectors.md#L226), which seems contradictory (saying that braiding twice does nothing by definition *and* that it can be non-trivial)

Jutho:
Thanks; they all look great so I no need to cherrypick. The link to the confusing paragraph does not seem to work; could you just copy paste the confusing sentences?

Jakob-Unfried:
I guess the link to the code line doesnt work if github shows you the rendered markdown...

I mean this part
```
By definition, applying this operation twice leads us back to the original
situation. If that is the case, the braiding is said to be symmetric. For more general
fusion categories, associated with the physics of anyonic particles, this is generally not
the case and, as a result, permutations of tensor indices are not unambiguously defined.
```

PR #70, opened 2022-06-07
truncated eigh function

philihps:
Hi Jutho, I've come across the necessity of using a truncated eigen decomposition, e.g. in this [time evolution algorithm](https://arxiv.org/pdf/2005.06104.pdf) or in the multiplication of MPO * MPS using the [density matrix algorithm](https://tensornetwork.org/mps/algorithms/denmat_mpo_mps/). Since TensorKit implements truncations only for SVDs, I thought I give it a try to implement a truncated eigh function.

In the code I reversed the order of the eigen values to be able to use the `_truncate!` function, which now makes the output of `eigh(t)` and `teigh(t, trunc = NoTruncation())` inconsistent. I don't know what your preference for sorting is, but this would need some further changes.

If this is a useful function for TensorKit, I would be happy to help finalize an implementation.

### DIFF
--- a/src/TensorKit.jl
+++ b/src/TensorKit.jl
@@ -65,7 +65,7 @@ export dot, norm, normalize, normalize!, tr
 export mul!, lmul!, rmul!, adjoint!, pinv, axpy!, axpby!
 export leftorth, rightorth, leftnull, rightnull,
         leftorth!, rightorth!, leftnull!, rightnull!,
-        tsvd!, tsvd, eigen, eigen!, eig, eig!, eigh, eigh!, exp, exp!,
+        tsvd!, tsvd, eigen, eigen!, eig, eig!, eigh, eigh!, teigh, teigh!, exp, exp!,
         isposdef, isposdef!, ishermitian, sylvester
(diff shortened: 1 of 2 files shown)


philihps:
I've just realized, that a truncated SVD is equally useful in the applications I mentioned. So I will close the PR.

Jutho:
Feel free to leave this open; I will try to make some time to review this soon, and it might still be a valuable addition.

Jutho:
Although I do indeed think that truncation on eigenvalues typically only makes sense if you have additional information, such as when they are all positive, because you have a positive definite matrix, in which case the eigenvalue and singular value decomposition coincide. I actually don't know if there is a computational advantage of one over the other for positive definite matrices.

philihps:
Yes, that is what I realised, too. Since the singular value decomposition is equivalent for most cases where an eigen decomposition is used (e.g. for density matrices), I guess it’s superfluous to add a truncated version. Although it might be nice to have it, for instance to throw away tiny eigen values when computing entanglement entropies, for instance. If you think it’s a useful addition to TensorKit, I will reopen the pull request.

PR #72, opened 2023-02-12
spacetype abstract types to traits

lkdvos:
* Removes `InnerProductSpace` and `EuclideanSpace` as abstract types.
* Adds `InnerProductStyle` as a trait based mechanism to query if a space has an inner product.
* Implements `NoInnerProduct`, `HasInnerProduct` and `EuclideanInnerProduct` as traits

Remark:
It might be clearer to remove this distinction altogether. I don't know of a current use of `GeneralSpace`, which is the only non-Euclidean spacetype to begin with. If we are happy with this design I'll update the docs as well

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 *-old
 __pycache__
 .ipynb*
+Manifest.toml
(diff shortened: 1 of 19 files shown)


lkdvos:
Cannot really seem to figure out why CI fails on windows with julia 1.6?

PR #74, opened 2023-07-24
Update for TensorOperations v4 

Jutho:


Jutho:
Ok, merging as is and continuing from there ... Thanks for all the work so far.

PR #75, opened 2023-07-25
Formatter

lkdvos:


Jutho:
I will merge at some point :-). But I first want to do one or two major updates, to make planar/plansor work again.

PR #77, opened 2023-08-10
Overhaul of implementation and interface of index manipulations, tensor operations and planar operations

Jutho:


Jutho:
Ok, I am merging it like this. The `contract!` implementation needs a revision (listed as TODO), in order to be able to use different backends. The plan would be:
* If `sectortype isa Trivial`, simply pas tensors `.data` to `TO.tensorcontract!`
* Otherwise, replicate the `mul!` implementation, but call `TO.tensorcontract!` on the matrix blocks instead of `LinearAlgebra.mul!`. `TO.tensorcontract!` has a dedicated implementation for the case where everything are matrices, but this way one could still have the backend select different matrix multiplication algorithms (e.g. BLAS vs Octavian vs ...)

PR #81, opened 2023-09-25
Fermion Refactor

lkdvos:
This PR changes the Fermion implementation such that only `FermionParity` remains, all other fermionic charges are just `ProductCharge` with fermionparity as a factor.


lkdvos:
It's mainly because I have had this locally for a while, the first commit predates quite a bit of the current history. The easiest solution is to squash and merge into a single commit, which keeps the linear history

PR #82, opened 2023-09-25
Automatic Differentiation

lkdvos:
This PR adds rrules and tests for most linear algebra methods.

Note that the tests for `tsvd` are disabled when working with `Complex`, as I have not been able to figure out how to make the lack of uniqueness of `U` and `V` work with finite differences and or the ChainRulesTestUtils interface.


ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -0,0 +1,337 @@
...
+_repartition(p::Index2Tuple, N₁::Int) = _repartition(linearize(p), N₁)
+function _repartition(p::Union{IndexTuple,Index2Tuple}, ::Index2Tuple{N₁}) where {N₁}
+    return _repartition(p, N₁)
+end
+function _repartition(p::Union{IndexTuple,Index2Tuple},
+                      ::AbstractTensorMap{<:Any,N₁}) where {N₁}
+    return _repartition(p, N₁)
+end
+
+TensorKit.block(t::ZeroTangent, c::Sector) = t
```
Jutho, line 29: This seems a bit suspicious? Why is this needed?

ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -0,0 +1,337 @@
...
+
+# Constructors
+# ------------
+
+@non_differentiable TensorKit.TensorMap(f::Function, storagetype, cod, dom)
```
Jutho, line 34: I guess that is good for now, but I could imagine that at some point we fill a tensor with a function at the very top of the program, and the gradient should be pushed back into that function. But I also wouldn't know how to deal with this from the top of my head.

lkdvos: It seems like this will become a little more clear with some concrete examples, but for now this mostly ensures that `f = zeros, randn, ones` works as expected. For functions that do have some internal data to fill in the tensormap, in principle we could pull back the tangents by somehow summing over the tangents from the blocks, but it's hard to imagine that this works in a completely black-box manner

ext/TensorKitChainRulesCoreExt.jl
lkdvos: This is mostly to avoid having to manually deal with the `ZeroTangent` type. For example, a QR decomposition where the final result does not depend on R, would generate a `ZeroTangent` for dR, which is just an abstract representation that behaves as the zero vector in any (co)vectorspace. As some of the `rrules` are implemented "blockwise", this would either require manually checking if a tangent is a `ZeroTangent`, or, which is what I chose to do, rely on the hope that the compiler would recognize that the blockwise operation results in ZeroTangents anyways, and thus automatically takes care of this

ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -0,0 +1,337 @@
...
+function ChainRulesCore.rrule(::typeof(scalar), t::AbstractTensorMap)
+    scalar_pullback(Δc) = NoTangent(), fill!(similar(t), unthunk(Δc))
+    return scalar(t), scalar_pullback
+end
+
+# LinearAlgebra
+# -------------
+
+function ChainRulesCore.rrule(::typeof(tr), A::AbstractTensorMap)
+    tr_pullback(Δtr) = NoTangent(), @thunk(Δtr * id(domain(A)))
```
lkdvos: This probably should not be thunk'd, as only a single tangent is returned, this should only be called when the tangent is actually needed

PR #83, opened 2023-09-27
Remove field type-parameter for vectorspaces

lkdvos:
This PR changes out the type-parameter-based approach to defining the underlying fields, in favor of accessing the field through a trait-based system. Specifically for the abstract vectorspace types, this allows for easier subtyping, and is warranted because these never really got used. CartesianSpace, ComplexSpace and GradedSpace all have fixed fields, which can easily be implemented case-by-case, and GeneralSpace still has the field parameter but this does not affect the other types.

Additionally, this also includes some promotion rules within the vector spaces, to allow operations between `ProductSpace` and `ElementarySpace` to be defined in a more conventional way, which is also more easily extended to new spacetypes.

Warning: this also includes some commits from PR #82, and should only be merged after that one is merged.


src/spaces/deligne.jl
```diff
@@ -15,46 +15,59 @@ natural representation spaces of the direct product of two groups.
...
     I₂ = sectortype(P₀)
     S = Vect[I₁ ⊠ I₂]
     N = length(P)
     return ProductSpace{S,N}(map(V -> V ⊠ P₀, tuple(P...)))
 end
 
-function ⊠(P₀::ProductSpace{<:ElementarySpace{ℂ},0}, P::ProductSpace{<:ElementarySpace{ℂ}})
+function ⊠(P₀::ProductSpace{<:ElementarySpace,0}, P::ProductSpace{<:ElementarySpace})
+    field(P) == field(P₀) ||
+        throw(ArgumentError("Deligne products require spaces over the same field"))
```
Jutho: Probably could be function
```julia
@noinline function check_fields_deligne(P_1,P_2)
    field(P_1) == field(P_2) || throw(ArgumentError("Deligne products require spaces over the same field"))
end
```
or maybe just
```
const _DELIGNE_ERROR = ArgumentError("Deligne products require spaces over the same field")
```
and then everywhere
```
field(P) == field(P₀) || throw(_DELIGNE_ERROR)
```

src/spaces/productspace.jl
```diff
@@ -255,3 +245,16 @@ Base.eltype(P::ProductSpace) = eltype(typeof(P))
...
+
+# Promotion and conversion
+# ------------------------
+function Base.promote_rule(::Type{S}, ::Type{<:ProductSpace{S}}) where {S<:ElementarySpace}
+    return ProductSpace{S}
+end
+
+# ProductSpace to ElementarySpace
+Base.convert(::Type{S}, P::ProductSpace{S,0}) where {S<:ElementarySpace} = oneunit(S)
+Base.convert(::Type{S}, P::ProductSpace{S}) where {S<:ElementarySpace} = fuse(P.spaces...)
```
Jutho, line 257: Is that something you want to have happen automatically?

Jutho: Although, I guess since the promote is giving you `ProductSpace` this would never be called "implicitly".

lkdvos: I think this is mostly a convenience thing, in principle `ProductSpace` is the "larger" type, so this should indeed never be called

src/spaces/deligne.jl
```diff
@@ -12,52 +12,68 @@ The Deligne tensor product also works in the type domain and for sectors and ten
...
     I₂ = sectortype(P)
     S = Vect[I₁ ⊠ I₂]
     N = length(P)
     return ProductSpace{S,N}(map(V -> P₀ ⊠ V, tuple(P...)))
 end
+
+@noinline function check_fields_deligne(P₁, P₂)
+    return field(P₁) == field(P₂) ||
+           throw(ArgumentError("Deligne products require spaces over the same field"))
+end
```
Jutho, line 77: I think I was too quick with this suggestion. The point is that you want to have that, if the compiler can infer that the check is true, that you don't create a string (as this allocates memory). This can be obtained by having the check, and then having the creation of the error message in a separate function that has a `@noinline` decorator. As the body of this function is then not executed if the check is inferably true, no string is being allocated. Check some of the `@noinline` use cases in Base: https://github.com/search?q=repo%3AJuliaLang%2Fjulia+%40noinline&type=code

The current implementation (which I suggested) is not good. It has the check together with the error creation, so when this function is executed, the string is always allocated, even if the check is true. The `@noinline` in this case is also not usefu; it just means that the check itself is also never inlined, and thus there is always a small function call overhead for evaluating the check.

lkdvos: Ah, I missed this. I think this should then fix it, I'll also do this for the inner product style checks.

lkdvos:
I think this can be merged?

src/spaces/productspace.jl
```diff
@@ -31,6 +31,8 @@ Base.axes(P::ProductSpace, n::Int) = axes(P.spaces[n])
 
 dual(P::ProductSpace{<:ElementarySpace,0}) = P
 dual(P::ProductSpace) = ProductSpace(map(dual, reverse(P.spaces)))
+flip(P::ProductSpace{<:ElementarySpace,0}) = P
+flip(P::ProductSpace) = ProductSpace(map(flip, reverse(P.spaces)))
```
Jutho: Do we want reverse in this case. Not sure where this is coming up?

Jutho:
I think this looks good, aside from my final question about `flip(ProductSpace)`

src/spaces/productspace.jl
lkdvos: I apparently got confused because of the MATLAB syntax for flip. In BraidingTensor, there is a check somewhere that is meant to verify if the input homspace is `V1 ⊗ V2 ← V2 ⊗ V1`, and somehow I used `flip(codomain(W)) == domain(W)` for this. I changed it to use `Base.reverse(ProductSpace)`, which should be fine?

Jutho: Yes I think the current version is better. This is the doc string for flip:

```
Return a single vector space of type `S` that has the same value of [`isdual`](@ref) as
`dual(V)`, but yet is isomorphic to `V` rather than to `dual(V)`.
```

PR #89, opened 2023-10-19
Add list of publications using TensorKit

lkdvos:


lkdvos:
List has been added to QuantumGhent website

PR #94, opened 2023-11-28
Rewrite the deduction of braidingtensor spaces

### LINKED ISSUE #93: Confusing/bugged? index convention for braiding tensors.
In the following code, which contracts two tensors in various ways depicted below, I'd expect all diagrams to be equivalent.

Nevertheless I get a ```spacemismatch``` for d-type diagram for reasons that totally evade me. I'm not sure if this is intended behavior or a bug but either way I think it deserves clarification.

```julia
using TensorKit, TensorOperations, Test

sp = ℂ^2
T = TensorMap(rand, ComplexF64, sp, sp)

#the normal diagram
a = @planar opt=true T[a;b]*T[b;a]

#the diagram with one winding, both conventions for the labeling of the braiding
b = @planar opt=true T[d;a]*T[b;c]*τ[c b;a d]
c = @planar opt=true T[d;a]*T[b;c]*τ[a c;d b]
@test a ≈ b ≈ c

#the diagrams with one pulling trough --> two taus --> both conventions lead to 4 diagrams
d = @planar opt=true T[f;a]*T[c;d]*τ[d b;c e]*τ[e b;a f]   #this one gives a space mismatch for reasons that evade me !
e = @planar opt=true T[f;a]*T[c;d]*τ[d b;c e]*τ[a e;f b]
f = @planar opt=true T[f;a]*T[c;d]*τ[c d;e b]*τ[e b;a f]
g = @planar opt=true T[f;a]*T[c;d]*τ[c d;e b]*τ[a e;f b]
```

potential_bug

Resulting error for the d-type diagram : 
```julia
ERROR: SpaceMismatch("(ℂ^2 ⊗ ℂ^2) ← ProductSpace{ComplexSpace, 0}() ≠ permute((ℂ^2 ⊗ ℂ^2) ← (ℂ^2 ⊗ ℂ^2)[(1, 2), (3, 4)] * ℂ^2 ← ℂ^2[(2, 1), ()], ((1, 2), ())")
Stacktrace:
 [1] planarcontract!(::TensorMap{ComplexSpace, 2, 0, Trivial, Matrix{ComplexF64}, Nothing, Nothing}, ::TensorKit.BraidingTensor{ComplexSpace, Matrix{Float64}}, ::Tuple{Tuple{Int64, Int64}, Tuple{Int64, Int64}}, ::TrivialTensorMap{ComplexSpace, 1, 1, Matrix{ComplexF64}}, ::Tuple{Tuple{Int64, Int64}, Tuple{}}, ::Tuple{Tuple{Int64, Int64}, Tuple{}}, ::One, ::Zero)
   @ TensorKit ~/.julia/dev/TensorKit/src/tensors/braidingtensor.jl:232
 [2] _planarcontract!(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any)
   @ TensorKit ~/.julia/dev/TensorKit/src/planar/postprocessors.jl:57
 [3] top-level scope
   @ REPL[652]:1
```

lkdvos:
Previously, this constructed an indexmap, but this is ambiguous if an index appears on multiple braidingtensors.
This should now be fixed by contructing braidingtensors independently. 
Also adds a test for the case that was not caught.
Fixes #93

### DIFF
--- a/src/planar/macros.jl
+++ b/src/planar/macros.jl
@@ -23,7 +23,7 @@ function planarparser(planarexpr, kwargs...)
 
     # braiding tensors need to be instantiated before kwargs are processed
     push!(parser.preprocessors, _construct_braidingtensors)
-    
+
     for (name, val) in kwargs
         if name == :order
             isexpr(val, :tuple) ||
@@ -62,7 +62,7 @@ function planarparser(planarexpr, kwargs...)
             throw(ArgumentError("Unknown keyword argument `name`."))
         end
     end
-    
+
     treebuilder = parser.contractiontreebuilder
     treesorter = parser.contractiontreesorter
     costcheck = parser.contractioncostcheck
--- a/src/planar/preprocessors.jl
+++ b/src/planar/preprocessors.jl
@@ -89,6 +89,7 @@ function _construct_braidingtensors(ex::Expr)
         return Expr(ex.head, map(_construct_braidingtensors, ex.args)...)
     end
 
+    # create a mapping of tensor_expr => braiding_constructor_expr
     i = 1
     translatebraidings = Dict{Any,Any}()
     while i <= length(list)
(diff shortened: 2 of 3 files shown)


Gertian:
MPSKit.jl doesn't compile against this.

I think the reason is that contractions such as :
```julia
@planar contractcheck = true opt=true d[f;g] := T[f;a]*B[c;d]*τ[d b;c e]*τ[e b;a g] 
```
result in ```ERROR: AssertionError: could not figure out all spaces ```

maartenvd:
Can you explain this bit?

> Previously, this constructed an indexmap, but this is ambiguous if an index appears on multiple braidingtensors.

The indexmap is always ambiguous, even when having a single braidingtensor between two regular tensors. You can find out the space by calling space(t,...) on either on the tensors, but that is not a real problem. I don't follow why chaining multiple braidingtensors together would make the indexmap more ambiguous.

lkdvos:
The problem is mostly that you need to know if the connection will be made between the codomain of the braidingtensor and the domain of the regular tensor, or the other way around, in order to determine if the space needs to be dual or not. However, keeping a single indexmap for multiple braidingtensors can lead to mismatches in that regards.

if you imagine an expression tau[a b; c d] * tau[b d; e f], you would construct two braidingtensors with space V in the codomain, which leads to a space mismatch when connecting them. (V \= V')

In other words, you cannot create a global indexmap, but should really create one per braidingtensor. The way I get around this is by just going through, creating as many braidingtensors as possible, and then really instantiating them into the expression as if they were tensors, and repeating this until all indices have been resolved

lkdvos:
Closed in favor of #95

PR #95, opened 2023-12-01
rewrite constructbraidingtensor preprocessor

Jutho:
This is an alternative to [PR#94](https://github.com/Jutho/TensorKit.jl/pull/94), that should also be correct if you have a tensor expression composed of a sum of different tensor network contractions. I've included the test cases from PR#94. 

I think also `_remove_braidingtensors` will suffer from the same bugs as the old `_construct_braidingtensors`, so I probably will also rewrite this one before merging.

Also, it needs some more testing and it contains a branch that should not active and currently prints out a "huh". So I should not forget to remove this.

Maybe you can already test with your code and with MPSKit: @lkdvos and @Gertian .


Jutho:
It is still easy to break `@planar` by doing things like adding `/2` or other scalar expressions in tensor expressions, so I will need to debug/rewrite further.

src/planar/planaroperations.jl
```diff
@@ -61,6 +61,7 @@ function planarcontract!(C::AbstractTensorMap{S,N₁,N₂},
     codB, domB = codomainind(B), domainind(B)
     oindA, cindA = pA
     cindB, oindB = pB
+    # @show codA, domA, codB, domB, oindA, cindA, oindB, cindB, pAB
```
lkdvos: Probably can remove this debug statement

src/planar/planaroperations.jl
```diff
@@ -91,6 +92,7 @@ _cyclicpermute(t::Tuple{}) = ()
 function reorder_indices(codA, domA, codB, domB, oindA, oindB, p1, p2)
     N₁ = length(oindA)
     N₂ = length(oindB)
+    # @show codA, domA, codB, domB, oindA, oindB, p1, p2
```
lkdvos: Probably can remove this debug statement

src/planar/planaroperations.jl
lkdvos:
```suggestion
```

src/planar/planaroperations.jl
lkdvos:
```suggestion
```

src/planar/preprocessors.jl
```diff
@@ -68,259 +68,314 @@ _is_adjoint(ex) = isexpr(ex, TO.prime)
...
+        return newex, unchanged
+    elseif isexpr(ex, :call) && ex.args[1] == :/ && length(ex.args) == 3
+        newarg, unchanged = _remove_braidingtensors!(ex.args[2], indexmap)
+        return Expr(:call, :/, newarg, ex.args[3]), unchanged
+    elseif isexpr(ex, :call) && ex.args[1] == :\ && length(ex.args) == 3
+        newarg, unchanged = _remove_braidingtensors!(ex.args[3], indexmap)
+        return Expr(:call, :\, ex.args[2], newarg), unchanged
     else
-        return Expr(ex.head, map(_purge_braidingtensors, args)...)
+        error("unexpected expression $ex")
```
lkdvos: in line 547 this is an `ArgumentError`, maybe switch this as well to be consistent?
```suggestion
        throw(ArgumentError("unexpected expression $ex"))
```

Jutho:
When the test turn green, I think this is now ready. I have locally tested the current master of MPSKit with this on Julia 1.10, and all seems fine. Maybe there are additional tests from PEPSKit that @Gertian wants to try?

Jutho:
Ok I will merge this. If there would still be issues with this in PEPSKit.jl @Gertian , let me know.

Gertian:
@Jutho  I'll see if my old code for contraction still works.

What is the current status for AD of planar fermionic diagrams ?

lkdvos:
Somewhere near the top of my ever larger to-do list :wink: I'm a little hesitant to promise due dates, but maybe end of January-ish? Maybe @Jutho has more time?

Gertian:
Great :)

I'm happy to hear it's still on your menu !

Jutho:
Definitely still on the menu. I have a lot of grading to do the coming weeks, but if I need some distraction or procrastination, the AD rules for the planar operations might be a good candidate. However, you are also very welcome to give it a try yourself if you want :-).

Gertian:
To be honest I have no idea where to get started with this.

Gertian:
Thanks, I'll play around with it today.
I'll let you know what I found (or didn't find)

PR #97, opened 2024-01-01
Sectors improvements

lkdvos:
Adds some small quality of life improvements:

* `ProductSector`s can now use `getindex` to access the component sectors, and `iterate` to allow easy destructuring. This removes the need for knowing the internal structure/fieldnames of these objects.
* Fermionic sectors are now printed with their unicode names. This makes their representations shorter, and brings them in line with the unicode names used for the groups.
* Fermionic tensors can now be converted to arrays, and users are warned that these don't preserve all of the fermionic properties.
* It is now possible to use any projective representation of `U1`, by using `PU1`/`PU1Irrep`. This is useful in the context of quantum systems that use `U1` in relation to particle number conservation, and allows fractional occupation numbers.

Some remarks/questions:
* The current implementation of `PU1Irrep` iterates through its values by only taking charges that are an integer value away from the starting point, which defaults to 0, thus only giving integer values. It is possible to iterate through all rational numbers in a somewhat reasonable fashion, but this is not compatible with ordering them based on their magnitude. I don't see a very elegant solution to this, and will just assume that iterating through all values of `PU1Irrep` is not something that is used. The only other solution I can think of consists of defining a "rescaled/shifted U1", which would just take two parametric values to denote the shift and scale, which corresponds to what numerators and denominators are allowed. This would then be mostly equivalent to regular U1. While this might be somewhat reasonable from the physics point of view, I am hesitant to implement this as shifting the unit feels weird in the context of group theory.
* The current implementation of `U1Irrep` allows half-integer charges, which would technically fall under the projective representations. We could remove this, but that would constitute a (unnecessary?) breaking change.
* printing product sector types could be slightly more pretty, if we don't mind violating type piracy slightly. I might be in favor of this, as this greatly shortens the output of sectors in for example Hubbard-like symmetries. We could consider a global switch to enable/disable pretty-printing (maybe including the use of unicode) through a preferences-based system, to have both options available


Jutho:
I can agree with most of these changes, but we might want to think about projective representations a bit more carefully. Strictly speaking, Zn and U1 don't have (nontrivial) projective representations. Furthermore, projective representations, when they exist, are necessarily higherdimensional.

For the case of U1 specifically, it should always be possible to shift the fractional charge to an integer charge, although I can see why this can sometimes be "convenient". True projective representations exist for O2 = CU1, but there only half-integer charges make sense. For convenience, I've also allowed half-integer U1 irreps, but I think already that one has some flaws.

The problems start with the categorical interpretation as the set of (representatives of) isomorphism classes of simple objects, and thus in particular, with iterating over all of them. For U1, I iterate over both the integer and half-(odd)-integer values, but that might not be what is intended. For the PU1Irrep implementation here, you iterate over integer differences with respect to the starting charge, but that means that if the initial charge is fractional, the set over which you iterate does not contain the zero charge, and so it does not represent the simple objects of a tensor category (as there needs to be a unit object).

Let's discuss this at the office.

Jutho:
Looks good to me, so I have merged.

PR #98, opened 2024-01-01
Small fixes

lkdvos:
* Fixes an issue with using `argmax(f, domain)` which is not defined in versions below 1.7
* Temporarily restricts TensorOperations to not be version 4.0.8 as this breaks @planar

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,22 +14,19 @@ TupleTools = "9d95972d-f1c8-5527-a6e0-b4b365fa01f6"
 VectorInterface = "409d34a3-91d5-4945-b6ec-7529ddf182d8"
 WignerSymbols = "9f57e263-0b3d-5e2e-b1be-24f2bb48858b"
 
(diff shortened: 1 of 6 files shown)


Jutho:
Looks good, although it seems one AD test still did not pass on Julia 1.6. But it seems an accuracy issue again, at least all the values that are printed seem to match with a large number of significant digits.

lkdvos:
I'll keep playing around a bit to see if I can get it to behave, it's probably not that important, but the green check is nice

Jutho:
Especially since this will affect all further CI runs once merged.

Project.toml
Jutho, line 1: Was there.a reason to reorganise the position of weakdeps and extensions in the Project.toml? I thought Aqua made me make exactly the opposite move in TensorOperations.jl

Jutho:
Ok I'll merge this and revert Project.toml changes in the final commits to #95

PR #101, opened 2024-01-17
Docs update I

lkdvos:
This is a first step in a rewrite of the documentation. The goal is to add docstrings to all exported methods and have a correctly structured lib and index part, before rethinking how to structure the manual part of the documentation.

### DIFF
--- a/src/auxiliary/random.jl
+++ b/src/auxiliary/random.jl
@@ -1,9 +1,32 @@
+"""
+    randuniform([::Type{T}=Float64], dims::Dims{N}) -> Array{T,N}
+
+Create an array of size `dims` with random entries uniformly distributed in the allowed
+values of `T`.
+
+See also [`randnormal`](@ref), [`randisometry`](@ref) and[`randhaar`](@ref).
+"""
 randuniform(dims::Base.Dims) = randuniform(Float64, dims)
 randuniform(::Type{T}, dims::Base.Dims) where {T<:Number} = rand(T, dims)
 
+"""
+    randnormal([::Type{T}=Float64], dims::Dims{N}) -> Array{T,N}
+
+Create an array of size `dims` with random entries normally distributed.
+
+See also [`randuniform`](@ref), [`randisometry`](@ref) and[`randhaar`](@ref).
+"""
 randnormal(dims::Base.Dims) = randnormal(Float64, dims)
 randnormal(::Type{T}, dims::Base.Dims) where {T<:Number} = randn(T, dims)
 
+"""
+    randisometry([::Type{T}=Float64], dims::Dims{2}) -> Array{T,2}
+    randhaar([::Type{T}=Float64], dims::Dims{2}) -> Array{T,2}
+
+Create a random isometry of size `dims`.
+
+See also [`randuniform`](@ref) and [`randnormal`](@ref).
+"""
 randisometry(dims::Base.Dims{2}) = randisometry(Float64, dims)
 function randisometry(::Type{T}, dims::Base.Dims{2}) where {T<:Number}
     return dims[1] >= dims[2] ? _leftorth!(randnormal(T, dims), QRpos(), 0)[1] :
--- a/src/auxiliary/random.jl
+++ b/src/auxiliary/random.jl
@@ -12,7 +12,7 @@ randuniform(::Type{T}, dims::Base.Dims) where {T<:Number} = rand(T, dims)
 """
     randnormal([::Type{T}=Float64], dims::Dims{N}) -> Array{T,N}
 
-Create an array of size `dims` with random entries normally distributed.
+Create an array of size `dims` with random entries distributed according to the standard normal distribution.
 
 See also [`randuniform`](@ref), [`randisometry`](@ref) and[`randhaar`](@ref).
 """
@@ -23,7 +23,7 @@ randnormal(::Type{T}, dims::Base.Dims) where {T<:Number} = randn(T, dims)
     randisometry([::Type{T}=Float64], dims::Dims{2}) -> Array{T,2}
     randhaar([::Type{T}=Float64], dims::Dims{2}) -> Array{T,2}
 
-Create a random isometry of size `dims`.
+Create a random isometry of size `dims`, uniformly distributed according to the Haar measure.
 
 See also [`randuniform`](@ref) and [`randnormal`](@ref).
 """
--- a/src/fusiontrees/manipulations.jl
+++ b/src/fusiontrees/manipulations.jl
@@ -605,6 +605,14 @@ function planar_trace(f₁::FusionTree{I}, f₂::FusionTree{I},
     return newtrees
 end
 
+"""
+    planar_trace(f::FusionTree{I,N}, q1::IndexTuple{N₃}, q2::IndexTuple{N₃}) where {I<:Sector,N,N₃}
+        -> <:AbstractDict{FusionTree{I,N-2*N₃}, <:Number}
+
+Perform a planar trace of the uncoupled indices of the fusion tree `f` at `q1` with those at
+`q2`, which are required to be pairwise neighbouring. The result is returned as a dictionary
+of output trees and corresponding coefficients.
+"""
 function planar_trace(f::FusionTree{I,N},
                       q1::IndexTuple{N₃}, q2::IndexTuple{N₃}) where {I<:Sector,N,N₃}
     u = one(I)
@@ -652,6 +660,13 @@ function planar_trace(f::FusionTree{I,N},
 end
 
 # trace two neighbouring indices of a single fusion tree
+"""
+    elementary_trace(f::FusionTree{I,N}, i) where {I<:Sector,N} -> <:AbstractDict{FusionTree{I,N-2}, <:Number}
+
+Perform an elementary trace of neighbouring uncoupled indices `i` and
+`i+1` on a fusion tree `f`, and returns the result as a dictionary of output trees and
+corresponding coefficients.
+"""
 function elementary_trace(f::FusionTree{I,N}, i) where {I<:Sector,N}
     (N > 1 && 1 <= i <= N) ||
         throw(ArgumentError("Cannot trace outputs i=$i and i+1 out of only $N outputs"))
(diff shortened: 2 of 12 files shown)


src/auxiliary/random.jl
```diff
@@ -1,9 +1,32 @@
+"""
+    randuniform([::Type{T}=Float64], dims::Dims{N}) -> Array{T,N}
+
+Create an array of size `dims` with random entries uniformly distributed in the allowed
+values of `T`.
```
Jutho, line 5: uniform is typically understood to be the uniform distribution on [0,1]. It is definitely not uniform over all values of `T` (which would not even be well defined).

src/auxiliary/random.jl
```diff
@@ -1,9 +1,32 @@
...
+
+See also [`randnormal`](@ref), [`randisometry`](@ref) and[`randhaar`](@ref).
+"""
 randuniform(dims::Base.Dims) = randuniform(Float64, dims)
 randuniform(::Type{T}, dims::Base.Dims) where {T<:Number} = rand(T, dims)
 
+"""
+    randnormal([::Type{T}=Float64], dims::Dims{N}) -> Array{T,N}
+
+Create an array of size `dims` with random entries normally distributed.
```
Jutho: The normal distribution with mean zero and standard deviation one is typically called the standard normal distribution, so maybe we should try to get this in.

src/auxiliary/random.jl
```diff
@@ -1,9 +1,32 @@
...
+See also [`randuniform`](@ref), [`randisometry`](@ref) and[`randhaar`](@ref).
+"""
 randnormal(dims::Base.Dims) = randnormal(Float64, dims)
 randnormal(::Type{T}, dims::Base.Dims) where {T<:Number} = randn(T, dims)
 
+"""
+    randisometry([::Type{T}=Float64], dims::Dims{2}) -> Array{T,2}
+    randhaar([::Type{T}=Float64], dims::Dims{2}) -> Array{T,2}
+
+Create a random isometry of size `dims`.
```
Jutho: I guess here you can mention the Haar measure or uniform measure over the compact manifold of isometric matrices.

src/fusiontrees/manipulations.jl
```diff
@@ -605,6 +605,14 @@ function planar_trace(f₁::FusionTree{I}, f₂::FusionTree{I},
     return newtrees
 end
 
+"""
+    planar_trace(f::FusionTree{I,N}, q1::IndexTuple{N₃}, q2::IndexTuple{N₃}) where {I<:Sector,N,N₃}
+        -> <:AbstractDict{FusionTree{I,N-2*N₃}, <:Number}
+
+Perform a planar trace of the uncoupled indices of the fusion tree `f` at `q1` with those at
+`q2`, which are required to be pairwise neighbouring. The result is returned as a dictionary
```
Jutho: pairwise neighbouring might be a bit confusing, as there are different ways in which traces of several indices can be planar:

It could for example be that q2[1] = q1[1]+1 and q2[2] = q1[2]+1, but also that q2[2] = q1[2]-3 , namely with q2[2] = q2[1] +1 = q1[1]+2 = q1[2] + 3

src/sectors/sectors.jl
```diff
@@ -62,16 +62,20 @@ Base.eltype(::Type{SectorValues{I}}) where {I<:Sector} = I
 Base.values(::Type{I}) where {I<:Sector} = SectorValues{I}()
 
 # Define a sector for ungraded vector spaces
-struct Trivial <: Sector
-end
+"""
+    Trivial
+
+Singleton type to represent the trivial sector, i.e. the unit element of the group
```
Jutho: This can get a bit confusing. The trivial group has only a unit element, but `Trivial()` typically (in our use case) refers to the only irrep of the trivial group, which is of course the trivial irrep and acts as the unit element in the category Rep[trivial group]. But I also don't immediately know the best way to phrase this.

src/tensors/abstracttensor.jl
```diff
@@ -23,15 +23,59 @@ i.e. a tensor map with only a non-trivial output space.
...
 sectortype(::Type{<:AbstractTensorMap{S}}) where {S<:IndexSpace} = sectortype(S)
+
 function InnerProductStyle(::Type{<:AbstractTensorMap{S}}) where {S<:IndexSpace}
     return InnerProductStyle(S)
 end
+
+"""
+    field(::Union{T,Type{T}}) where {T<:AbstractTensorMap} -> Type{𝕂<:Field}
+
+Return the type of field `𝕂` of a tensor.
```
Jutho, line 50: I am wondering if this method definition is really useful? It is the field of the vector spaces over which the tensormap is defined. But this could easily be confused with the scalartype of the tensormap, which can be different. Maybe we should consider deprecating this method?

src/tensors/abstracttensor.jl
```diff
@@ -23,15 +23,59 @@ i.e. a tensor map with only a non-trivial output space.
...
+"""
 field(::Type{<:AbstractTensorMap{S}}) where {S<:IndexSpace} = field(S)
+
+"""
+    numout(::Union{T,Type{T}}) where {T<:AbstractTensorMap} -> Int
+
+Return the number of output spaces of a tensor. This should be equivalent to `length(codomain(t))`.
```
Jutho: The second sentence is a bit strange since you did not name the argument in the method definition, so `t` is undefined here. This also only works if indeed the argument is a TensorMap instance (not the TensorMap type).

src/tensors/abstracttensor.jl
```diff
@@ -23,15 +23,59 @@ i.e. a tensor map with only a non-trivial output space.
...
+
+See also [`numin`](@ref) and [`numind`](@ref).
+"""
 numout(::Type{<:AbstractTensorMap{<:IndexSpace,N₁,N₂}}) where {N₁,N₂} = N₁
+
+"""
+    numin(::Union{T,Type{T}}) where {T<:AbstractTensorMap} -> Int
+
+Return the number of input spaces of a tensor. This should be equivalent to `length(domain(t))`.
```
Jutho: Same remark about `t`

src/tensors/abstracttensor.jl
```diff
@@ -23,15 +23,59 @@ i.e. a tensor map with only a non-trivial output space.
...
+
+See also [`numout`](@ref) and [`numind`](@ref).
+"""
 numin(::Type{<:AbstractTensorMap{<:IndexSpace,N₁,N₂}}) where {N₁,N₂} = N₂
+
+"""
+    numind(::Union{T,Type{T}}) where {T<:AbstractTensorMap} -> Int
+
+Return the total number of input and output spaces of a tensor. This should be equivalent to `numout(t) + numin(t)`.
```
Jutho: Same remark about `t`

src/tensors/abstracttensor.jl
```diff
@@ -51,28 +95,80 @@ similarstoragetype(t::AbstractTensorMap, T) = similarstoragetype(typeof(t), T)
...
+The index information of a tensor, i.e. the `HomSpace` of its domain and codomain. If `i` is specified, return the `i`-th index space.
+"""
 space(t::AbstractTensorMap) = HomSpace(codomain(t), domain(t))
 space(t::AbstractTensorMap, i::Int) = space(t)[i]
+
+"""
+    dim(t::AbstractTensorMap) -> Int
+
+The total number of free parameters of a tensor, discounting the entries that are
+fixed by symmetry.
```
Jutho: I think it could be mentioned that this is the dimension of the hom space on which the tensormap is defined.

src/tensors/abstracttensor.jl
```diff
@@ -51,28 +95,80 @@ similarstoragetype(t::AbstractTensorMap, T) = similarstoragetype(typeof(t), T)
...
+"""
 function codomainind(::Type{<:AbstractTensorMap{<:IndexSpace,N₁,N₂}}) where {N₁,N₂}
     return ntuple(n -> n, N₁)
 end
+codomainind(t::AbstractTensorMap) = codomainind(typeof(t))
+
+"""
+    domainind(::Union{T,Type{T}}) where {T<:AbstractTensorMap} -> Tuple{Int}
+
+Provide all indices of the domain of a tensor.
```
Jutho: I find `provide` a strange word here, in contrast to `return`. Not sure why, could be subjective. For one of the other reason, it feels like `Provide` requires the user to provide something.

src/tensors/abstracttensor.jl
```diff
@@ -87,6 +183,46 @@ function adjointtensorindices(t::AbstractTensorMap, p::Index2Tuple)
...
+@doc """
+    blockdim(t::AbstractTensorMap, c::Sector) -> Base.Dims
+
+Return the dimensions of the block of a tensor corresponding to a coupled sector `c`.
+""" blockdim
+
+@doc """
+    fusiontrees(t::AbstractTensorMap)
+
+Return an iterator over all fusion trees of a tensor.
```
Jutho: I don't recall if on other places I have called this "splitting-fusion-tree pairs".

src/sectors/sectors.jl
Jutho: Maybe it could also be mentioned that this is the unit object of the category Vect of ordinary vector spaces (not as the first sentence).

Jutho (approved):
Thanks for this, doc updates are always very welcome.

src/auxiliary/random.jl
lkdvos: Ah I got confused by not reading the docstring of `rand` far enough:
```julia-repl
help?> rand
search: rand randn transcode macroexpand @macroexpand @macroexpand1

  rand([rng=default_rng()], [S], [dims...])

  Pick a random element or array of random elements from the set of values
  specified by S; S can be

    •  an indexable collection (for example 1:9 or ('x', "y", :z)),

    •  an AbstractDict or AbstractSet object,

    •  a string (considered as a collection of characters), or

    •  a type: the set of values to pick from is then equivalent to
       typemin(S):typemax(S) for integers (this is not applicable to
       BigInt), to [0, 1) for floating point numbers and to [0, 1)+i[0,
       1) for complex floating point numbers;
```

I am wondering though if this method really makes sense, as it is just a copy of `rand`

Jutho: Maybe not, but this also applies to `randnormal` which is just equal to `randn`.

src/tensors/abstracttensor.jl
lkdvos: It seems like fusion-splitting tree pairs is mentioned in the documentation somewhere, as well as splitting-fusion tree pairs. I am completely fine with either, but I think the current naming is definitely confusing.

src/tensors/abstracttensor.jl
lkdvos: I am definitely in favor of doing this. I think it is now only being used in checking if the deligne product is allowed, i.e. it will not allow you to take the deligne product of `CartesianSpace` and `ComplexSpace`, but I am not sure if this check is really all that useful.

PR #103, opened 2024-02-01
Some docs fixes

leburgel:
Some fixes for docstrings that appeared as missing in `lib/tensors`. Some of them were actually missing and I only added the ones I was confident about adding, and some of them weren't appearing because they were already referenced somewhere else so I had to mess with the type specifiers a bit.

The only thing here I'm not too sure about are the references to `blocksectors` in `lib/spaces`. I found two relevant documented methods so I added these both explicitly, but it feels like maybe there should just be one docstring for an overarching `blocksectors(::VectorSpace)`?

### DIFF
--- a/src/tensors/tensor.jl
+++ b/src/tensors/tensor.jl
@@ -37,7 +37,7 @@ end
 #! format: on
 
 """
-    Tensor{S<:IndexSpace, N, I<:Sector, A, F₁, F₂} = TensorMap{S, N, 0, I, A, F₁, F₂}
+    Tensor{S, N, I, A, F₁, F₂} = TensorMap{S, N, 0, I, A, F₁, F₂}
 
 Specific subtype of [`AbstractTensor`](@ref) for representing tensors whose data is stored
 in blocks of some subtype of `DenseMatrix`.
@@ -45,7 +45,7 @@ in blocks of some subtype of `DenseMatrix`.
 A `Tensor{S, N, I, A, F₁, F₂}` is actually a special case `TensorMap{S, N, 0, I, A, F₁, F₂}`,
 i.e. a tensor map with only a non-trivial output space.
 """
-const Tensor{S<:IndexSpace,N,I<:Sector,A,F₁,F₂} = TensorMap{S,N,0,I,A,F₁,F₂}
+const Tensor{S,N,I,A,F₁,F₂} = TensorMap{S,N,0,I,A,F₁,F₂}
 """
     TrivialTensorMap{S<:IndexSpace, N₁, N₂, A<:DenseMatrix} = TensorMap{S, N₁, N₂, Trivial, 
                                                                         A, Nothing, Nothing}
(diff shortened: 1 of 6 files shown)


src/tensors/tensor.jl
```diff
@@ -36,9 +36,32 @@ struct TensorMap{S<:IndexSpace, N₁, N₂, I<:Sector, A<:Union{<:DenseMatrix,Se
...
+"""
 const Tensor{S<:IndexSpace,N,I<:Sector,A,F₁,F₂} = TensorMap{S,N,0,I,A,F₁,F₂}
+"""
+    TrivialTensorMap{S<:IndexSpace, N₁, N₂, A<:DenseMatrix} = TensorMap{S, N₁, N₂, Trivial, 
+                                                                        A, Nothing, Nothing}
+
+A special case of [`TensorMap`](@ref) for representing tensor maps with trivial symmetry,
+i.e., whose `sectortype` is `Trivial`.
+"""
 const TrivialTensorMap{S<:IndexSpace,N₁,N₂,A<:DenseMatrix} = TensorMap{S,N₁,N₂,Trivial,A,
```
Jutho: It seems like my code is not very consistent with which subtype relations are imposed in these constants/aliases. In this one, `A<:DenseMatrix`, but in the `Tensor` definition it is not.

src/tensors/tensor.jl
```diff
@@ -36,9 +36,32 @@ struct TensorMap{S<:IndexSpace, N₁, N₂, I<:Sector, A<:Union{<:DenseMatrix,Se
...
                                                                        Nothing,Nothing}
+"""
+    TrivialTensor{S<:IndexSpace, N, I<:Sector, A, F₁, F₂} = TrivialTensorMap{S, N, 0, A}
```
Jutho: There is no `I<:Sector` in `TrivialTensor`.

Jutho:
Thanks. Looks good, up to one comment.

src/tensors/tensor.jl
lkdvos: I think it might be the most convenient to not impose anything for the aliases. The restrictions will still apply, as the original type has them, and if ever they need to change, this needs to only happen at a single point in the code.

leburgel: Makes sense. I dropped the restrictions that are just duplicates of the original `TensorMap` ones, and it seems all I'm left with is the `A<:DenseMatrix` in `TrivialTensorMap`. As far as I understand the storagetype for a `Tensor` can still be a `SectorDict{<:Sector,<:DenseMatrix}` (with a single entry?), so I don't think this one should have an additional subtype restriction?

leburgel:
Also snuck in a small typo fix in the Deligne product docstring while I"m at it.

src/tensors/tensor.jl
Jutho: Thanks. This looks good now. As for the `A` parameter, we are considering to revise the `TensorMap` structure quite drastically, so this will probably change anyway in the (hopefully near) future.

lkdvos (approved):

PR #105, opened 2024-03-05
Docs additions related to constructing a given symmetric `TensorMap`

leburgel:
Some docs additions that should make it easier for users to find out how to construct symmetric `TensorMap`s corresponding to specific physical systems. In summary I:

- Explicitly documented the fields of the `FusionTree` type, since these need to be accessed explicitly when assigning data slices labeled by splitting-fusion-tree pairs.
- Documented the `fusiontrees` iterator generator, both in lib/sectors and lib/tensors.
- Explicitly documented the fields of all exported `Sector` subtypes, since again these need to be accessed explicitly when assigning data slices.
- Added docs for `FermionNumber` and `FermionSpin`.
- Added docstrings for the `TensorMap` constructors to lib/tensors.
- Documented the different flavors of `getindex` and `setindex!` for `TensorMap`s in lib/tensors.
- Expanded the section on "tensor maps from existing data" in man/tensors to be a bit more explicit about the Heisenberg example.
- Added a section in man/tensors on how to assign data slices after initializing a `TensorMap`, both by using `block(::TensorMap, ::Sector)` to assign the full matrix blocks, and using the `setindex!` syntax `t[f₁::FusionTree, f₂::FusionTree] = v` to assign slices of `t` labeled by splitting-fusion-tree pairs `(f₁, f₂)`.

The changes to man/tensors might make it a bit physics-oriented, but since this example was already being used I didn't see much harm in making it more explicit like this. I anyway wouldn't know how to explain the splitting-fusion-tree slice assignment in any other way.

### DIFF
--- a/docs/src/lib/tensors.md
+++ b/docs/src/lib/tensors.md
@@ -24,21 +24,36 @@ TrivialTensorMap
 TrivialTensor
 ```
 
-## Specific `TensorMap` constructors
+## `TensorMap` constructors
 
+### General constructors
+
+A general `TensorMap` can be constructed by specifying its data, codmain and domain in one
+of the following ways:
+```@docs
+TensorMap(::AbstractDict{<:Sector,<:DenseMatrix}, ::ProductSpace{S,N₁},
+                   ::ProductSpace{S,N₂}) where {S<:IndexSpace,N₁,N₂}
+TensorMap(::Any, ::Type{T}, codom::ProductSpace{S},
+                   dom::ProductSpace{S}) where {S<:IndexSpace,T<:Number}
+TensorMap(::DenseArray, ::ProductSpace{S,N₁}, ::ProductSpace{S,N₂};
+                   tol) where {S<:IndexSpace,N₁,N₂}
+```
+
+Several special-purpose methods exist to generate data according to specific distributions:
+```@docs
+randuniform
+randnormal
+randisometry
+```
+
+### Specific constructors
+
+Additionally, several special-purpose constructors exist to generate data according to specific distributions:
 ```@docs
 id
 isomorphism
 unitary
 isometry
-```
-
-Additionally, several special-purpose methods exist to generate data according to specific distributions:
-
-```@docs
-randuniform
-randnormal
-randisometry
 ```
 
 ## Accessing properties and data
@@ -71,8 +86,26 @@ blocksectors(::AbstractTensorMap)
 blockdim(::AbstractTensorMap, ::Sector)
 block
 blocks
-fusiontrees
+fusiontrees(::AbstractTensorMap)
 hasblock
+```
+
+For `TensorMap`s with `Trivial` `sectortype`, the data can be directly accessed and
+manipulated in a straightforward way:
+```@docs
+Base.getindex(t::TrivialTensorMap)
+Base.getindex(t::TrivialTensorMap, indices::Vararg{Int})
+Base.setindex!(t::TrivialTensorMap, ::Any, indices::Vararg{Int})
+```
+
+For general `TensorMap`s, this can be done using custom `getindex` and `setindex!` methods:
+```@docs
+Base.getindex(t::TensorMap{<:IndexSpace,N₁,N₂,I},
+              sectors::Tuple{Vararg{I}}) where {N₁,N₂,I<:Sector}
+Base.getindex(t::TensorMap{<:IndexSpace,N₁,N₂,I},
+              f₁::FusionTree{I,N₁},
+              f₂::FusionTree{I,N₂}) where {N₁,N₂,I<:Sector}
+Base.setindex!(::TensorMap{<:IndexSpace,N₁,N₂,I}, ::Any, ::FusionTree{I,N₁}, ::FusionTree{I,N₂}) where {N₁,N₂,I<:Sector}
 ```
 
 ## `TensorMap` operations
--- a/docs/src/man/tensors.md
+++ b/docs/src/man/tensors.md
@@ -313,8 +313,8 @@ e.g. for `CartesianSpace` or `ComplexSpace`. Then the `data` array is just resha
 matrix form and referred to as such in the resulting `TensorMap` instance. When `spacetype`
 is `GradedSpace`, the `TensorMap` constructor will try to reconstruct the tensor data such
 that the resulting tensor `t` satisfies `data == convert(Array, t)`. This might not be
-possible, if the data does not respect the symmetry structure. Let's sketch procedure using
-a simple physical example, namely the SWAP gate on two qubits,
+possible, if the data does not respect the symmetry structure. This procedure can be
+sketched using a simple physical example, namely the SWAP gate on two qubits,
 ```math
 \begin{align*}
 \mathrm{SWAP}: \mathbb{C}^2 \otimes \mathbb{C}^2 & \to \mathbb{C}^2 \otimes \mathbb{C}^2\\
--- a/docs/src/man/tensors.md
+++ b/docs/src/man/tensors.md
@@ -344,7 +344,7 @@ V3 = U1Space(1/2=>1,-1/2=>1) # restricted space that only uses the `σ_z` rotati
 t3 = TensorMap(data, V3 ⊗ V3, V3 ⊗ V3)
 for (c,b) in blocks(t3)
     println("Data for block $c :")
-    b |> disp
+    disp(b)
     println()
 end
 ```
@@ -390,7 +390,7 @@ block(t4, U1Irrep(1)) .= [1;;];
 block(t4, U1Irrep(-1)) .= [1;;];
 for (c, b) in blocks(t4)
     println("Data for block $c :")
-    b |> disp
+    disp(b)
     println()
 end
 ```
@@ -429,7 +429,7 @@ for (f₁, f₂) in fusiontrees(t5)
 end
 for (c, b) in blocks(t5)
     println("Data for block $c :")
-    b |> disp
+    disp(b)
     println()
 end
 ```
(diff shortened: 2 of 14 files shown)


docs/src/man/tensors.md
```diff
@@ -313,13 +313,28 @@ e.g. for `CartesianSpace` or `ComplexSpace`. Then the `data` array is just resha
...
+```math
+\begin{align*}
+\mathrm{SWAP}: \mathbb{C}^2 \otimes \mathbb{C}^2 & \to \mathbb{C}^2 \otimes \mathbb{C}^2\\
+|i\rangle \otimes |j\rangle &\mapsto |j\rangle \otimes |i\rangle.
+\end{align*}
+```
+This operator can be rewritten in terms of the familiar Heisenberg exchange interaction
+``\vec{S}_i \cdot \vec{S}_j`` as
+```math
+\mathrm{SWAP} = 2 \vec{S}_i \cdot \vec{S}_j + \frac{1}{2} 𝟙,
```
lkdvos, line 327: in Math environments, maybe replacing 𝟙 with the latex equivalent is nicer?

docs/src/lib/tensors.md
```diff
@@ -24,23 +24,38 @@ TrivialTensorMap
 TrivialTensor
 ```
 
-## Specific `TensorMap` constructors
+## `TensorMap` constructors
 
+### General constructors
+
+A general `TensorMap` can be constructed by specifying its data, codmain and domain in one
+of the following ways:
```
Gertian, line 32: Perhaps highlight that the datastructure that the user gives depends on the symmetries of the codomain and domain ?

In particular 
Complexspaces -> single dense array
Symmetric spaces -> fusion tree

leburgel: I don't explain it in the .md file directly, but I think this should be clear from the docstrings I added (which is indeed not obvious since there's no docs preview (yet)). In particular, the docstring for the constructor from a dense array states how this works with and without symmetries,
https://github.com/Jutho/TensorKit.jl/blob/d671313509d90b634fa1609a1f497c9597af7644/src/tensors/tensor.jl#L313-L344
and the assignment using fusion trees (if this is what you meant) is in the `setindex!` docstring
https://github.com/Jutho/TensorKit.jl/blob/d671313509d90b634fa1609a1f497c9597af7644/src/tensors/tensor.jl#L616-L627

docs/src/man/tensors.md
```diff
@@ -357,6 +372,68 @@ axes(P, (SU2Irrep(1), SU2Irrep(0), SU2Irrep(2)))
...
+A first option is to directly set the full matrix block for each coupled sector in the
+`TensorMap`. For the example with U₁ symmetry, this can be done as
+```@repl tensors
+t4 = TensorMap(zeros, V3 ⊗ V3, V3 ⊗ V3);
+block(t4, U1Irrep(0)) .= [1 0; 0 1];
+block(t4, U1Irrep(1)) .= [1;;];
+block(t4, U1Irrep(-1)) .= [1;;];
+for (c, b) in blocks(t4)
+    println("Data for block $c :")
+    b |> disp
```
lkdvos: Is there a reason this is not just `disp(b)`?

src/sectors/fermions.jl
```diff
@@ -74,9 +86,18 @@ type_repr(::Type{FermionNumber}) = "FermionNumber"
 # convenience default converter -> allows Vect[FermionNumber](1 => 1)
 Base.convert(::Type{FermionNumber}, a::Int) = FermionNumber(a)
 
+"""
+    const FermionSpin = SU2Irrep ⊠ FermionParity
+    FermionSpin(j::Real)
+
+Represents the fermion spin as the direct product of a ``SU₂`` irrep `j` and a fermion
+parity, with the restriction that the fermion parity is odd if and only if ` 2 * j` is odd.
```
lkdvos:
```suggestion
parity, with the restriction that the fermion parity is odd iff ` 2 * j` is odd.
```

src/tensors/abstracttensor.jl
```diff
@@ -185,15 +185,16 @@ function adjointtensorindices(t::AbstractTensorMap, p::Index2Tuple)
 end
 
 @doc """
-    blocks(t::AbstractTensorMap)
+    blocks(t::AbstractTensorMap) -> Dict{<:Sector,<:DenseMatrix}
```
lkdvos:
```suggestion
    blocks(t::AbstractTensorMap) -> SectorDict{<:Sector,<:DenseMatrix}
```
I think this actually returns a `SortedVectorDict`, which is aliased by `SectorDict`. The name is probably clear enough to warrant specifying this detail?

src/tensors/tensor.jl
```diff
@@ -199,6 +219,32 @@ function _buildblockstructure(P::ProductSpace{S,N}, blocksectors) where {S<:Inde
...
+## Arguments
+- `f`: callable object that returns a `DenseMatrix`, or `UndefInitializer`.
+- `eltype::Type{<:Number}`: element type of the data.
+- `codomain::ProductSpace{S,N₁}`: the codomain as a `ProductSpace` of `N₁` spaces of type
+  `S<:ElementarySpace`.
+- `domain::ProductSpace{S,N₂}`: the domain as a `ProductSpace` of `N₂` spaces of type
+  `S<:ElementarySpace`.
+
+If `eltype` is left unspecified, `f` should support the calling syntax `f(::Tuple{Int,Int})`
+such that `f((m, n))` returns a `DenseMatrix` with `size(f(m, n)) == (m, n)`. If `eltype` is
```
```suggestion
such that `f((m, n))` returns a `DenseMatrix` with `size(f((m, n))) == (m, n)`. If `eltype` is
```

src/tensors/tensor.jl
```diff
@@ -466,6 +546,21 @@ blocks(t::TensorMap) = t.data
...
+    t[sectors]
+
+Return a view into the data slice of `t` corresponding to the splitting - fusion tree pair
+with combined uncoupled charges `sectors`. In particular, if `sectors == (s1..., s2...)`
+where `s1` and `s2` correspond to the coupled charges in the codomain and domain
+respectively, then a `StridedViews.StridedView` of size
+`(dims(codomain(t), s1)..., dims(domain(t), s2))` is returned.
+
+This method is only available for the case where `FusionStyle(I) isa UniqueFusion`,
+since it assumes a well-defined coupled charge.
```
```suggestion
since it assumes a uniquely defined coupled charge.
```

src/tensors/tensor.jl
```diff
@@ -466,6 +546,21 @@ blocks(t::TensorMap) = t.data
 fusiontrees(t::TrivialTensorMap) = ((nothing, nothing),)
 fusiontrees(t::TensorMap) = TensorKeyIterator(t.rowr, t.colr)
 
+"""
+    Base.getindex(t::TensorMap{<:IndexSpace,N₁,N₂,I},
+                  sectors::Tuple{Vararg{I}}) where {N₁,N₂,I<:Sector}
```
```suggestion
                  sectors::NTuple{N₁ + N₂, I}) where {N₁,N₂,I<:Sector} 
```
This pattern does not work for julia syntax, but maybe we can just use it in the docstring?

lkdvos (approved):

docs/src/man/tensors.md
leburgel: That's what I tried first, but `\mathbb{1}` (which I think should just be this one) didn't render correctly in my local build. I'll try to fix this.

docs/src/man/tensors.md
leburgel: No, just straight copied from above without thinking about it. I'll change it.

src/tensors/tensor.jl
leburgel: Certainly fine by me, looks more clear, but I don't really have the means to assess whether this might be bad practice...

lkdvos (approved):

docs/src/man/tensors.md
```diff
@@ -313,13 +313,28 @@ e.g. for `CartesianSpace` or `ComplexSpace`. Then the `data` array is just resha
 matrix form and referred to as such in the resulting `TensorMap` instance. When `spacetype`
 is `GradedSpace`, the `TensorMap` constructor will try to reconstruct the tensor data such
 that the resulting tensor `t` satisfies `data == convert(Array, t)`. This might not be
-possible, if the data does not respect the symmetry structure. Let's sketch this with a
-simple example
+possible, if the data does not respect the symmetry structure. Let's sketch procedure using
```
Jutho: This procedure can be sketched using ...

src/fusiontrees/fusiontrees.jl
```diff
@@ -4,13 +4,18 @@
...
-has `M=max(0, N-2)` inner lines. Furthermore, for `FusionStyle(I) isa GenericFusion`,
-the `L=max(0, N-1)` corresponding vertices carry a label of type `T`. If `FusionStyle(I)
-isa MultiplicityFreeFusion, `T = Nothing`.
+sectors to a coupled sector. It actually represents a splitting tree, but fusion tree
+is a more common term.
+
+## Fields
+- `uncoupled::NTuple{N,I}`: the uncoupled sectors coming out of the splitting tree, before the possible 𝑍 isomorphism.
+- `coupled::I`: the coupled sector.
+- `isdual::NTuple{N,Bool}`: indicates whether the an isomorphism is present (`true`) or not
```
Jutho: "the an isomorphism". The old text just had "an isomorphism". But that is not very descriptive. These are called Z isomorphisms in my convention, as present in the description of the `uncoupled` field. Maybe there you can add (see `isdual`).

Jutho:
Thanks; this looks great. I've added only a few minor comments.

lkdvos:
I think everything is addressed now, shall I merge?

Jutho:
Yes, I had planned to do this yesterday but wanted to check with Lander whether this was ready and then somehow forgot.

PR #106, opened 2024-03-15
Add scalartype parameter to `AbstractTensorMap`

lkdvos:
This PR promotes the scalartype of a tensor to be present in the `AbstractTensorMap` supertype.
In order to make this work, some minor changes were made in many method signatures. Wherever possible, I chose to remove as many of the type parameters as possible, to write the code as generic as possible.

I chose for `AbstractTensorMap{E,S,N1,N2}`, having the eltype/scalartype first in mimicking of `AbstractArray{T,N}`. It has some advantages, but also some disadvantages as opposed to `AbstractTensorMap{S,N1,N2,E}`. In the end, I don't have a very strong opinion on this, so if you do, I'll just turn it around. 

Importantly, this is very much a breaking change, so it might not be a bad idea to start a v1.0 branch somewhere and point this PR to that, and only merging into master when we are ready for that.


lkdvos:
Closing this PR to bundle it in a big v1.0 update

PR #107, opened 2024-03-20
Fix typo

leburgel:
I noticed I missed a backtick in the previous update, and couldn't let it go...


lkdvos:
Forever in the commit history...

Jutho:
?

PR #109, opened 2024-03-27
Add unicode alternatives: `otimes` and `oplus`

### LINKED ISSUE #108: \otimes+TAB in PyJulia
Hi,

I am using PyJulia to call Julia code from Python 3, and I have found an issue when trying to type the tensor product. Namely, \otimes+TAB doesn't work, neither it does copy pasting from Julia (both things tried on terminal on an Ubuntu machine). Is there any solution for this? 

I found [this](https://pyjulia.readthedocs.io/en/latest/limitations.html) in the PyJulia documentation suggesting that it might not be possible, but a workaround would be highly appreciated.

Thanks!

lkdvos:
Adds a non-unicode alternative to `\otimes` for sectors: `fusionproduct`.

Fixes #108 .

### DIFF
--- a/src/TensorKit.jl
+++ b/src/TensorKit.jl
@@ -37,7 +37,7 @@ export infimum, supremum, isisomorphic, ismonomorphic, isepimorphic
 
 # methods for sectors and properties thereof
 export sectortype, sectors, hassector, Nsymbol, Fsymbol, Rsymbol, Bsymbol,
-       frobeniusschur, twist
+       frobeniusschur, twist, otimes
 export fusiontrees, braid, permute, transpose
 export ZNSpace, SU2Irrep, U1Irrep, CU1Irrep
 # other fusion tree manipulations, should not be exported:
--- a/src/TensorKit.jl
+++ b/src/TensorKit.jl
@@ -30,7 +30,7 @@ export TruncationScheme
 export SpaceMismatch, SectorMismatch, IndexError # error types
 
 # general vector space methods
-export space, field, dual, dim, dims, fuse, flip, isdual, insertunit
+export space, field, dual, dim, dims, fuse, flip, isdual, insertunit, oplus
 
 # partial order for vector spaces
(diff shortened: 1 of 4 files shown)


Jutho:
Probably needs some coverage :-). I am also wondering whether it is not better to really just have a unicode alternative to `\otimes`, that does not depend on the argument types (which is probably what you had suggested at first). Otherwise, we further complicate the interface. I would be fine with, `otimes`, `tensorproduct`, `tensorprod`. In TensorCore.jl, they simply call this `tensor`, but this is maybe a bit confusing (although one does read it like "a tensor b").

lkdvos:
I definitely also prefer having a fixed unicode alternative for this. It also creates less confusion when implementing the interface, as with the current implementation you would have to implement `\otimes` and not `fusionproduct`. 

I do think `tensor` might be confusing, so I would vote `otimes`, and then use `const ⊗ = otimes`.

Jutho:
Ok sounds good. However, I would keep using the actual unicode in all of the method definitions; I do like it more than the ascii `otimes`.

lkdvos:
I'll implement when I find some time.

Just for completeness: I double-checked and the alias thing works how you would expect: they are completely equivalent and defining methods for either implements them for both, so we can definitely just keep using the unicode everywhere.

```julia-repl
julia> otimes(a, b) = a * b
otimes (generic function with 1 method)

julia> const ⊗ = otimes
otimes (generic function with 1 method)

julia> ⊗(a::Int, b::Int) = 2 * a * b
otimes (generic function with 2 methods)

julia> otimes(1, 2)
4

julia> ⊗(1, 2)
4

julia> 1 ⊗ 2
4

julia> ⊗(1.0, 2.0)
2.0
```

lkdvos:
I implemented the changes, and in the meantime also added a non-unicode alternative for `oplus`. As these are now true aliases, I don't think I need to add coverage?

PR #112, opened 2024-03-31
Update ProductSpace constructors with promote/convert

lkdvos:


src/spaces/productspace.jl
```diff
@@ -22,11 +22,27 @@ ProductSpace{S}(spaces::Vararg{S}) where {S<:ElementarySpace} = ProductSpace{S}(
 function ProductSpace(spaces::Tuple{S,Vararg{S}}) where {S<:ElementarySpace}
     return ProductSpace{S,length(spaces)}(spaces)
 end
-function ProductSpace(space1::S, rspaces::Vararg{S}) where {S<:ElementarySpace}
+function ProductSpace(space1::ElementarySpace, rspaces::Vararg{ElementarySpace})
     return ProductSpace((space1, rspaces...))
 end
 
-ProductSpace(P::ProductSpace) = P
```
Jutho, line 29: Was it ok to just remove this one? This one is not replaced by any of the new constructors below I believe.

lkdvos: You're probably right, I'm not sure why I removed it in the first place actually

PR #115, opened 2024-04-07
Specify eltype for CU1Irrep product iterator

lkdvos:
I am not sure if this actually has some performance implications in the code, but at the very least it changes the following:
Before:
```julia-repl
julia> collect(CU1Irrep(0) ⊗ CU1Irrep(1))
1-element Vector{Any}:
 Irrep[CU₁](1, 2)
```
---
After:
```julia-repl
julia> collect(CU1Irrep(0) ⊗ CU1Irrep(1))
1-element Vector{CU1Irrep}:
 (1, 2)
```

### DIFF
--- a/src/sectors/irreps.jl
+++ b/src/sectors/irreps.jl
@@ -334,6 +334,7 @@ function Base.length(p::CU1ProdIterator)
         return 2
     end
 end
+Base.eltype(::Type{CU1ProdIterator}) = CU1Irrep
 
 ⊗(a::CU1Irrep, b::CU1Irrep) = CU1ProdIterator(a, b)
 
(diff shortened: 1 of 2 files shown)


Jutho:
Maybe there can be some extra tests in test/sectors.jl:
e.g. below line 21
```julia
@test eltype(it) == I
@constinferred collect(it)
```

lkdvos:
I was looking for something to test, but `collect(it)` does not actually help, the inferred type is `Vector{Any}` and this is what it returns, so no error is raised. Similarly I got stuck with trying to test `iterate(it)`, this is also correctly inferred, it is just the iterator interface of julia Base that is a bit strange, as it defaults `eltype` to `Any`.
Testing the `eltype` directly is definitely better, I really should have thought of that :grin:

Jutho:
Maybe `typeof(collect(it)) == Vector{I}` as extra test?

Jutho:
Looks great; thanks. I'll merge when the tests are complete.

PR #116, opened 2024-04-09
Add `repartition(!)(::AbstractTensorMap)`

lkdvos:


lkdvos:
@Jutho, I am not actually sure how to make `repartition(t, N1, N2` inferrable. I've tried `Base.@constprop :aggressive`, which does not seem to help, and without resorting to `val(N)` type things, I don't think I can make it work. So maybe the next question is, does it matter? In principle, this is just a wrapper around `transpose`, so in important cases you can use that instead?

Jutho:
Can you try with `@constinferred repartition(t, $k)` (but maybe first locally)? Constant propagation can only work if the argument is really a constant. So either you need to explicitly have `@constinferred repartition(t, 3)` or something, or you need to splice in the value using `$k`.

lkdvos:
> Can you try with `@constinferred repartition(t, $k)` (but maybe first locally)? Constant propagation can only work if the argument is really a constant. So either you need to explicitly have `@constinferred repartition(t, 3)` or something, or you need to splice in the value using `$k`.

Thanks! Indeed, combining the `@constprop` and the corrected `@constinferred` fixes everything, at least on my machine. I can't seem to find a way of doing this in julia v1.6 though...

lkdvos:
So now, for Julia v1.6, there is a slight type-instability for `repartition`, but everything else should be fine. I think if the tests pass, this is ready to go.

Jutho:
Ok I will merge. Given that this is an addition to the public interface, I guess this should become TensorKit v0.13? Let's hope this doesn't bring bad luck :-).

lkdvos:
We might want to tag the latest version (v0.12.3) before this then, I think this didn't happen yet.

test/tensors.jl
```diff
@@ -201,22 +201,36 @@ for V in spacelist
...
                 t = Tensor(rand, ComplexF64, W)
+                a = convert(Array, t)
                 for k in 0:5
                     for p in permutations(1:5)
                         p1 = ntuple(n -> p[n], k)
                         p2 = ntuple(n -> p[k + n], 5 - k)
                         t2 = permute(t, (p1, p2))
                         a2 = convert(Array, t2)
-                        @test a2 ≈ permutedims(convert(Array, t), (p1..., p2...))
+                        @test a2 ≈ permutedims(a, (p1..., p2...))
```
Jutho, line 224: That's a nice improvement!

PR #117, opened 2024-04-18
Multithreaded implementations using OhMyThreads

lkdvos:
This is an attempt at providing a unifying interface for various multi-threaded approaches to looping over blocksectors, and looping over fusiontrees.

It can be controlled through the `scheduler` kwarg in various places, upon which OhMyThreads dispatches, such that there is a possibility of creating and experimenting with scheduler types if one so pleases.
By default, there is `default_scheduler(tensortype))=SerialScheduler()`, which just mimicks the non-multithreaded approaches at minimal overhead.
Various built-in schedulers exist for OhMyThreads, with Greedy, Dynamic or Static schemes, each with up- and downsides depending on the specific case as well as spacetypes. The hope is that this allows a flexible approach that enables users to tweak these settings if they so please, without having to deal with this when this is not necessary.

## To do:
- [ ] Documentation
- [ ] Proper benchmarks
- [ ] Abstraction layer for looping over blocksectors, or looping over fusiontrees
- [ ] "Atomic Int" based scheduler + task sorting


Jutho:
Maybe the `default_scheduler` should also take in the specific function as an argument, so that one can define

`default_scheduler(::Type{<:TensorMap}, ::typeof(tsvd!))`

In principle, the scheduler keyword argument should also be added to `tsvd` etc, unless we require people that want to use this, to use the more expert-mode mutating methods.

PR #120, opened 2024-05-05
Utility and Quality of life for VectorSpaces

lkdvos:
This PR adds some missing constructors for ComplexSpace and CartesianSpace with generators, and refactors some small pieces of code that were being used for permuting and manipulating spaces to be in one place.

Additionally, it cleans up some of the space checks, and adds missing `@inbounds`  annotations for the `add_***_kernels`, which avoids a couple hash calls in the `getindex` for the fusion trees I think.

### DIFF
--- a/src/spaces/homspace.jl
+++ b/src/spaces/homspace.jl
@@ -115,3 +115,16 @@ function dim(W::HomSpace)
     end
     return d
 end
+
+# Operations on HomSpaces
+# -----------------------
+function permute(W::HomSpace{S}, (p₁, p₂)::Index2Tuple{N₁,N₂}) where {S,N₁,N₂}
+    cod = ProductSpace{S,N₁}(map(n -> W[n], p₁))
+    dom = ProductSpace{S,N₂}(map(n -> dual(W[n]), p₂))
+    return cod ← dom
+end
+
+function Base.:*(W::HomSpace{S}, V::HomSpace{S}) where {S}
+    domain(W) == codomain(V) || throw(SpaceMismatch("$(domain(W)) ≠ $(codomain(V))"))
+    return HomSpace(codomain(W), domain(V))
+end
--- a/src/spaces/homspace.jl
+++ b/src/spaces/homspace.jl
@@ -124,7 +124,13 @@ function permute(W::HomSpace{S}, (p₁, p₂)::Index2Tuple{N₁,N₂}) where {S,
     return cod ← dom
 end
 
-function Base.:*(W::HomSpace{S}, V::HomSpace{S}) where {S}
+"""
+    compose(W::HomSpace, V::HomSpace)
+
+Obtain the HomSpace that is obtained from composing the morphisms in `W` and `V`. For this
+to be possible, the domain of `W` must match the codomain of `V`.
+"""
+function compose(W::HomSpace{S}, V::HomSpace{S}) where {S}
     domain(W) == codomain(V) || throw(SpaceMismatch("$(domain(W)) ≠ $(codomain(V))"))
     return HomSpace(codomain(W), domain(V))
 end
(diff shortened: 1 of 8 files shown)


src/tensors/indexmanipulations.jl
```diff
@@ -313,23 +311,23 @@ function add_transform!(tdst::AbstractTensorMap{S,N₁,N₂},
...
             dest = $(codomain(tdst))←$(domain(tdst)), p₁ = $(p₁), p₂ = $(p₂)"))
     end
 
     I = sectortype(S)
     if p₁ == codomainind(tsrc) && p₂ == domainind(tsrc)
-        add!(tdst, tsrc, α, β)
+        @inbounds add!(tdst, tsrc, α, β)
     elseif I === Trivial
-        _add_trivial_kernel!(tdst, tsrc, (p₁, p₂), fusiontreetransform, α, β, backend...)
+        @inbounds _add_trivial_kernel!(tdst, tsrc, (p₁, p₂), fusiontreetransform, α, β,
```
Jutho: I don't think these `@inbounds` do anything. I use the `@inbounds` and `@boundscheck` to actually do the space checking. The `_add_...` methods don't have additional space checks, so there is not really anything to disable I think. Only `add!` has an additional space check, but that one is not indicated with `@boundscheck`, so it cannot be disabled.

src/spaces/homspace.jl
```diff
@@ -115,3 +115,16 @@ function dim(W::HomSpace)
...
+
+# Operations on HomSpaces
+# -----------------------
+function permute(W::HomSpace{S}, (p₁, p₂)::Index2Tuple{N₁,N₂}) where {S,N₁,N₂}
+    cod = ProductSpace{S,N₁}(map(n -> W[n], p₁))
+    dom = ProductSpace{S,N₂}(map(n -> dual(W[n]), p₂))
+    return cod ← dom
+end
+
+function Base.:*(W::HomSpace{S}, V::HomSpace{S}) where {S}
```
Jutho: I am not sure about this one. `HomSpace` is perhaps not a subtype of `VectorSpace`, but mathematically it kind of is, and on `VectorSpace` instances, `*` acts very differently.

lkdvos: Ah, I forgot that we use `*` as a non-unicode otimes... What do you think about `\circ` instead? I don't really mind the name, it's just a simple convenience thing

src/tensors/indexmanipulations.jl
lkdvos: I actually did not mean to elide the space checks, but rather the checks from https://github.com/Jutho/TensorKit.jl/blob/370dd92a3a17958bd028f8211ee2464b87040511/src/tensors/tensor.jl#L604-L608
I did however definitely do this wrong, so let me try again ;)

src/spaces/homspace.jl
lkdvos: I changed it to just be `compose`. If we want, we can still add the unicode alternative, but I think this does not interfere with anything.

Jutho: I like compose; it might conflict with other packages but for now we don't export it I assume? I actually wouldn't mind also defining compose for `TensorMap`s. I think of `*` as a convenience symbol that does certain functionality depending on the type, i.e.
* `*` on vector spaces is `otimes`
* `*` on tensor maps is `compose`
* `*` on ... is ...

src/tensors/indexmanipulations.jl
Jutho: Ok I see. Another solution would have been to add `@propagate_inbounds` to the `_add` functions.

lkdvos: I don't think it matters too much, should I still change it?

Jutho: No this is perfect.

PR #123, opened 2024-05-12
small fix in backward rule for leftorth, rightorth

tangwei94:
Previously, the backwards for `leftorth` and `rightorth` occasionally give wrong results when `R` or `L` are not used in forward computations.
For example, 
```julia 
    function f(C1)
        U, _ = leftorth(C1; alg=QRpos())
        return norm(tr(U))
    end
    C = TensorMap(rand, ComplexF64,  ℂ^4,  ℂ^4);
    f'(C) # this gradient may be wrong
```

### DIFF
--- a/ext/TensorKitChainRulesCoreExt.jl
+++ b/ext/TensorKitChainRulesCoreExt.jl
@@ -539,9 +539,9 @@ function qr_pullback!(ΔA::AbstractMatrix, Q::AbstractMatrix, R::AbstractMatrix,
     ΔR1 = view(ΔR, 1:p, :)
     ΔR11 = view(ΔR, 1:p, 1:p)
 
-    M = zeros(eltype(R), (p, p))
+    M = similar(R, (p, p))
     ΔR isa AbstractZero || mul!(M, ΔR1, R1')
-    ΔQ isa AbstractZero || mul!(M, Q1', ΔQ1, -1, +1)
+    ΔQ isa AbstractZero || mul!(M, Q1', ΔQ1, -1, !(ΔR isa AbstractZero))
     view(M, lowertriangularind(M)) .= conj.(view(M, uppertriangularind(M)))
     if eltype(M) <: Complex
         Md = view(M, diagind(M))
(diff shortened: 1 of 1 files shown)


ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -539,7 +539,7 @@ function qr_pullback!(ΔA::AbstractMatrix, Q::AbstractMatrix, R::AbstractMatrix,
     ΔR1 = view(ΔR, 1:p, :)
     ΔR11 = view(ΔR, 1:p, 1:p)
 
-    M = similar(R, (p, p))
+    M = zeros(eltype(R), (p, p))
     ΔR isa AbstractZero || mul!(M, ΔR1, R1')
     ΔQ isa AbstractZero || mul!(M, Q1', ΔQ1, -1, +1)
```
lkdvos: Great catch!
I think something like this might even be able to avoid the extra zeroing out:
```suggestion
    M = similar(R, (p, p))
    ΔR isa AbstractZero || mul!(M, ΔR1, R1')
    ΔQ isa AbstractZero || mul!(M, Q1', ΔQ1, -1, !(ΔR isa AbstractZero))
```

ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -593,7 +593,7 @@ function lq_pullback!(ΔA::AbstractMatrix, L::AbstractMatrix, Q::AbstractMatrix,
     ΔL1 = view(ΔL, :, 1:p)
     ΔR11 = view(ΔL, 1:p, 1:p)
 
-    M = similar(L, (p, p))
+    M = zeros(eltype(L), (p, p))
     ΔL isa AbstractZero || mul!(M, L1', ΔL1)
     ΔQ isa AbstractZero || mul!(M, ΔQ1, Q1', -1, +1)
```
tangwei94:
```suggestion
    M = similar(L, (p, p))
    ΔL isa AbstractZero || mul!(M, L1', ΔL1)
    ΔQ isa AbstractZero || mul!(M, ΔQ1, Q1', -1, !(ΔL isa AbstractZero))
```

ext/TensorKitChainRulesCoreExt.jl
tangwei94: Indeed. I similarly modified `lq_pullback!`

Jutho:
Well spotted, I can imagine that this was a tough one to debug. I will merge as soon as CI finished running.

PR #126, opened 2024-05-28
Fix rrules for Fermionic symmetries

qmortier:
rrules for tensorcontract etc for fermions


lkdvos:
I rebased this on the current master and added a few changes that simplify the implementations.

Most notably I added a method for twisting multiple tensor indices at the same time. I'll finish this up tomorrow, but I would like to add:
- [ ] some tests for the vectorized twist
- [ ] some simplifications in the tensortrace implementation

test/ad.jl
```diff
@@ -111,20 +123,25 @@ Vlist = ((ℂ^2, (ℂ^3)', ℂ^3, ℂ^2, (ℂ^2)'),
...
+          ℂ[SU2Irrep](0 => 1, 1 => 1),
           ℂ[SU2Irrep](1 // 2 => 1, 1 => 1)',
-          ℂ[SU2Irrep](0 => 2, 1 // 2 => 2),
+          ℂ[SU2Irrep](1 // 2 => 2),
           ℂ[SU2Irrep](0 => 1, 1 // 2 => 1, 3 // 2 => 1)'))
 
-@testset "Automatic Differentiation with spacetype $(TensorKit.type_repr(eltype(V)))" verbose = true for V in
-                                                                                                         Vlist
-    @testset "Basic Linear Algebra with scalartype $T" for T in (Float64, ComplexF64)
+@timedtestset "Automatic Differentiation with spacetype $(TensorKit.type_repr(eltype(V)))" verbose = true for V in
```
Jutho, line 130: What is currently the minimal Julia version? Since Julia 1.8, there is no more benefit from using `@timedtestset` over Julia's `Tests.@testset`. (In fact, the former is a copy of the latter).

test/ad.jl
```diff
@@ -157,71 +174,100 @@ Vlist = ((ℂ^2, (ℂ^3)', ℂ^3, ℂ^2, (ℂ^2)'),
...
+                d = 0
+                local V1, V2, V3
+                # retry a couple times to make sure there is at least some nonzero elements
+                for _ in 1:10
+                    k1 = rand(1:3)
+                    k2 = rand(1:2)
+                    k3 = rand(1:2)
+                    V1 = map(v -> rand(Bool) ? v' : v, rand(V, k1))
+                    V2 = map(v -> rand(Bool) ? v' : v, rand(V, k2))
+                    V3 = map(v -> rand(Bool) ? v' : v, rand(V, k3))
```
Jutho: you probably want mapreduce with `otimes` as reduction?

test/ad.jl
lkdvos: Currently, TensorKit still supports 1.6. I have nothing against bumping this, or even getting rid of the `@timedtestset`, I just noticed that all other tests still use `@timedtestset`, so I went for uniformity. Let's keep them all as `@timedtestset` for now, and consider removing that dependency in a separate PR?

Jutho:
I guess something is off with the random number generator on Julia 1.6 :-)

lkdvos:
I am investigating now, but I have no clue what is going on here

lkdvos:
So, as far as I can tell, the issue arises whenever the SVD is truncating between two singular values that are close together, such that the FiniteDifferences calls yield different sizes of tensors, which then ofcourse cannot be added. 
I changed some of the parameters around, and hope that it works now, but it's definitely not "fixed".

lkdvos:
Everything seems to turn green again, ready to go for me!

ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -111,19 +121,17 @@ function ChainRulesCore.rrule(::typeof(⊗), A::AbstractTensorMap, B::AbstractTe
         dA_ = @thunk begin
             ipA = (codomainind(A), domainind(A))
             pB = (allind(B), ())
-            dA = zerovector(A,
-                            TensorOperations.promote_contract(scalartype(ΔC),
-                                                              scalartype(B)))
-            dA = tensorcontract!(dA, ipA, ΔC, pΔC, :N, B, pB, :C)
+            dA = zerovector(A, promote_contract(scalartype(ΔC), scalartype(B)))
+            tB = twist(B, filter(x -> isdual(space(B, x)), allind(B)))
+            dA = tensorcontract!(dA, ipA, ΔC, pΔC, :N, tB, pB, :C)
```
Jutho, line 128: I am wondering whether it would not be better to write these rules using `planarcontract!`. The tensor product method `⊗` works for any two tensors of the same spacetype.

lkdvos: I agree, but currently `planarcontract!` does not have support for conj flags (which we could add manually, but that feels a bit annoying since we should have this in the near future anyways), and I think there are some assumptions on `pAB` which we'd have to explicitly check here. Maybe a TODO comment for after the planar updates would be better?

Jutho: I agree that a TODO will do for now. But maybe using the (hermitian) adjoint of `A` and of `B` in the definition for the (pullback) adjoints of `B` and and `A` would be very natural.

src/tensors/indexmanipulations.jl
```diff
@@ -248,17 +248,31 @@ function twist!(t::AbstractTensorMap, i::Int; inv::Bool=false)
...
+end
 
 """
-    twist(t::AbstractTensorMap, i::Int; inv::Bool=false)
-        -> t
+    twist(tsrc::AbstractTensorMap, i::Int; inv::Bool=false) -> tdst
+    twist(tsrc::AbstractTensorMap, is; inv::Bool=false) -> tdst
 
-Apply a twist to the `i`th index of `t` and return the result as a new tensor.
+Apply a twist to the `i`th index of `tsrc` and return the result as a new tensor.
```
Jutho, line 270: I forgot to mention that the doc string needed an update to reflect the `is` argument here.

PR #127, opened 2024-06-03
Fix undefined `tol` and remove unused variable in QR and LQ pullback

leburgel:
The gauge dependency checks in `qr_pullback!` and `lq_pullback!` threw an error because the check `tol` was defined in a let block. Also, the `ΔR11` variable was unused in both cases, so I removed it.

### DIFF
--- a/ext/TensorKitChainRulesCoreExt.jl
+++ b/ext/TensorKitChainRulesCoreExt.jl
@@ -537,7 +537,7 @@ function qr_pullback!(ΔA::AbstractMatrix, Q::AbstractMatrix, R::AbstractMatrix,
                       rtol::Real=atol > 0 ? 0 : eps(real(eltype(R)))^(3 / 4))
     Rd = view(R, diagind(R))
     tol = atol > 0 ? atol : rtol * maximum(abs, Rd)
-    p = findlast(x -> abs(x) >= tol, Rd)
+    p = findlast(>=(tol) ∘ abs, Rd)
     m, n = size(R)
 
     Q1 = view(Q, :, 1:p)
@@ -590,7 +590,7 @@ function lq_pullback!(ΔA::AbstractMatrix, L::AbstractMatrix, Q::AbstractMatrix,
                       rtol::Real=atol > 0 ? 0 : eps(real(eltype(L)))^(3 / 4))
     Ld = view(L, diagind(L))
     tol = atol > 0 ? atol : rtol * maximum(abs, Ld)
-    p = findlast(x -> abs(x) >= tol, Ld)
+    p = findlast(>=(tol) ∘ abs, Ld)
     m, n = size(L)
 
     L1 = view(L, :, 1:p)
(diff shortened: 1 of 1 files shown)


lkdvos:
I don't think this matters for performance all that much, but in principle you could check if this leads to a "boxed" variable inside of the closure used by `findlast`. The let block avoids this which could lead to more type stability, but it probably does not make a huge difference. You could check this by descending into it via Cthulhu.jl, which should show some warnings.

leburgel:
I could check, but if it's anyway safer I could also just put back the let block and redefine the tolerance later when it's used for the check (potentially again in a scoped block, though I don't know if it makes any difference for that specific use). I definitely wasn't thinking about this, just wanted it to run.

In any case, the way I changed it here is exactly the same as what was already happening in the SVD pullback. So if this is something we should be careful with, I can also change it there.

ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -536,9 +536,8 @@ function qr_pullback!(ΔA::AbstractMatrix, Q::AbstractMatrix, R::AbstractMatrix,
                       atol::Real=0,
                       rtol::Real=atol > 0 ? 0 : eps(real(eltype(R)))^(3 / 4))
     Rd = view(R, diagind(R))
-    p = let tol = atol > 0 ? atol : rtol * maximum(abs, Rd)
-        findlast(x -> abs(x) >= tol, Rd)
-    end
+    tol = atol > 0 ? atol : rtol * maximum(abs, Rd)
+    p = findlast(x -> abs(x) >= tol, Rd)
```
Jutho: You can avoid the closure problem by using `findlast(Base.Fix2(>=, tol) ∘ abs, Rd)`

ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -591,9 +589,8 @@ function lq_pullback!(ΔA::AbstractMatrix, L::AbstractMatrix, Q::AbstractMatrix,
                       atol::Real=0,
                       rtol::Real=atol > 0 ? 0 : eps(real(eltype(L)))^(3 / 4))
     Ld = view(L, diagind(L))
-    p = let tol = atol > 0 ? atol : rtol * maximum(abs, Ld)
-        findlast(x -> abs(x) >= tol, Ld)
-    end
+    tol = atol > 0 ? atol : rtol * maximum(abs, Ld)
+    p = findlast(x -> abs(x) >= tol, Ld)
```
Jutho: Same here. Also, if you noticed a similar pattern in other functions like svd pullback, feel free to also update accordingly :-)

ext/TensorKitChainRulesCoreExt.jl
leburgel: Thanks!

ext/TensorKitChainRulesCoreExt.jl
leburgel: The svd pullback already handled it this way, so no need for further changes as far as I can tell. It seems things that look "exactly the same" to my untrained eye can still be quite different :)

Jutho:
Thanks.

PR #131, opened 2024-06-13
small fix in the backward rule of `norm`

tangwei94:
Previously, the backward of `norm` will become `NAN` if the norm is zero.

### DIFF
--- a/ext/TensorKitChainRulesCoreExt.jl
+++ b/ext/TensorKitChainRulesCoreExt.jl
@@ -173,7 +173,7 @@ function ChainRulesCore.rrule(::typeof(norm), a::AbstractTensorMap, p::Real=2)
     p == 2 || error("currently only implemented for p = 2")
     n = norm(a, p)
     function norm_pullback(Δn)
-        return NoTangent(), a * (Δn' + Δn) / (n * 2 + eps(real(eltype(a)))), NoTangent()
+        return NoTangent(), a * (Δn' + Δn) / 2 / hypot(n, eps(one(n))), NoTangent()
     end
     return n, norm_pullback
 end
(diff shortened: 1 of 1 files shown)


ext/TensorKitChainRulesCoreExt.jl
```diff
@@ -172,7 +172,9 @@ end
 function ChainRulesCore.rrule(::typeof(norm), a::AbstractTensorMap, p::Real=2)
     p == 2 || error("currently only implemented for p = 2")
     n = norm(a, p)
-    norm_pullback(Δn) = NoTangent(), a * (Δn' + Δn) / (n * 2), NoTangent()
+    function norm_pullback(Δn)
+        return NoTangent(), a * (Δn' + Δn) / (n * 2 + eps(real(eltype(a)))), NoTangent()
```
Jutho: Could you change this to `a * (Δn' + Δn) / 2 / hypot(n, eps(one(n)))` ? I think that is slightly nicer, in that, if `n==1.`, then `n+eps()` is no longer exactly one, but `hypot(1.,eps())` is still exactly `1.` due to machine precision.

Jutho:
Thanks; that's an important fix. I made one suggestion in the code.

ext/TensorKitChainRulesCoreExt.jl
tangwei94:
```suggestion
        return NoTangent(), a * (Δn' + Δn) / 2 / hypot(n, eps(one(n))), NoTangent()
```

tangwei94: Yes, this is indeed better. I have modified this line according to the suggestion.

PR #132, opened 2024-06-23
Move FiniteDifferences support to package extension

lkdvos:
This is mostly to avoid having to redefine this functionality in downstream packages, i.e. in PEPSKit.jl.
See https://github.com/QuantumKitHub/PEPSKit.jl/pull/47

### DIFF
--- a/ext/TensorKitFiniteDifferencesExt.jl
+++ b/ext/TensorKitFiniteDifferencesExt.jl
@@ -1,0 +1,31 @@
+module TensorKitFiniteDifferencesExt
+
+using TensorKit
+using FiniteDifferences
+
+function FiniteDifferences.to_vec(t::T) where {T<:TensorKit.TrivialTensorMap}
+    vec, from_vec = to_vec(t.data)
+    return vec, x -> T(from_vec(x), codomain(t), domain(t))
+end
+function FiniteDifferences.to_vec(t::AbstractTensorMap)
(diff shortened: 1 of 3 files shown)


ext/TensorKitFiniteDifferencesExt.jl
```diff
@@ -0,0 +1,31 @@
...
+function FiniteDifferences.to_vec(t::AbstractTensorMap)
+    vec = mapreduce(vcat, blocks(t); init=scalartype(t)[]) do (c, b)
+        return reshape(b, :) .* sqrt(dim(c))
+    end
+    vec_real = scalartype(t) <: Real ? vec : collect(reinterpret(real(scalartype(t)), vec))
+
+    function from_vec(x_real)
+        x = scalartype(t) <: Real ? x_real : reinterpret(scalartype(t), x_real)
+        t′ = similar(t)
+        ctr = 0
```
Jutho: I actually don't like `ctr` very much :-). Leaving out vowels is not a proper abbreviation. I'd prefer `counter` or just `i` or `pos`.

lkdvos: I refactored to avoid it :)

Jutho: Thanks. That is a nice implementation.

PR #133, opened 2024-06-27
Refactor ChainRulesCoreExt into separate files

lkdvos:
This refactors the chainrules extension into separate files.
In the meantime, I improved the test coverage a bit, and ironed out some type instabilities.


ext/TensorKitChainRulesCoreExt/constructors.jl
```diff
@@ -17,8 +17,10 @@ function ChainRulesCore.rrule(::typeof(Base.copy), t::AbstractTensorMap)
     return copy(t), copy_pullback
 end
 
+# this rule does not work for generic symmetries, as we currently have no way to
+# project back onto the symmetric subspace
```
Jutho: What do you mean by that? The TensorMap constructor does try to project back, right, i.e. I thought that `TensorMap(convert(Array, t), space(t)) == t` should work for the group-based symmetries. Maybe we need to test this somewhere.

Jutho: Now I see, I guess you mean that you need a `ProjectTo` that acts on the array data and would remove the entries where there needs to be zeros for the `TensorMap` constructor.

lkdvos: Exactly. I added some tests for these methods and ran into issues, precisely because of that.
In principle, these rules were mostly defined in the first place to facilitate some testing, so I thought it might be safest to just restrict this to the case where we know it will work, because it wasn't entirely clear to me how to implement this projection for non-abelian symmetries. It's not too hard to project out the entries that have to be zero, but I am not entirely sure how to project to the nearest linear combination of clebsch gordans?
I would also have to think a bit longer to consider if we need to support this projection, or throw an error if the deviation is too large, as there really should not be any contribution to the gradient along the non-symmetric directions

Jutho: Well, for cotangents I think directions that are out of the subspace can just be ignored without warning or error. In the factorisations, for example, I warn for directions that are dual to gauge-dependent directions (e.g. changing the phase of an eigenvector), but I simply ignore cotangent directions that cannot contribute (e.g. changing the norm of an eigenvector).

Jutho: I guess that in principle this error could be made optional:
https://github.com/Jutho/TensorKit.jl/blob/master/src/tensors/tensor.jl#L380

lkdvos: I just used `tol=Inf`, I'm running the tests locally right now, I'll push an update in a bit

ext/TensorKitChainRulesCoreExt/constructors.jl
```diff
@@ -4,34 +4,34 @@
...
     end
     return A, convert_pullback
 end
 
 function ChainRulesCore.rrule(::typeof(Base.convert), ::Type{Dict}, t::AbstractTensorMap)
     out = convert(Dict, t)
-    function convert_pullback(c)
+    function convert_pullback(c′)
+        c = unthunk(c′)
         if haskey(c, :data) # :data is the only thing for which this dual makes sense
```
Jutho, line 35: How about `dual = typeof(out)(:data => c[:data])` ?

src/tensors/tensor.jl
```diff
@@ -350,52 +350,57 @@ function TensorMap(data::DenseArray, codom::ProductSpace{S,N₁}, dom::ProductSp
...
     end
+
+    if !isapprox(data, convert(Array, t); atol=tol)
+        throw(ArgumentError("Data has non-zero elements at incompatible positions"))
+    end
+
+    return t
 end
 
+function _interleave(a::Base.Dims{N}, b::Base.Dims{N}) where {N}
```
Jutho: `Dims` is exported by `Base`, so no need to qualify.

How about a recursive definition?
```julia
_interleave(a::Dims{N}, b::Dims{N}) where {N} = (a[1], b[1], _interleave(tail(a), tail(b))...)
_interleave(::Tuple{}, ::Tuple{}) = ()
```
(`Base.tail` is not exported, but TensorKit imports it explicitly)

that should be simple enough to be inlined and not cause problems with type inference.

src/tensors/tensor.jl
```diff
@@ -350,52 +350,57 @@ function TensorMap(data::DenseArray, codom::ProductSpace{S,N₁}, dom::ProductSp
...
-        end
-        i = 1
-        for (c, b) in blocks(t2)
-            for k in 1:length(b)
-                b[k] = lhs[i]
-                i += 1
-            end
+    end
+
+    t = TensorMap(zeros, eltype(data), codom, dom)
```
Jutho: Maybe it is useful to wrap this in a function on its own, that can be used by the ad rules, and only keep the final `isapprox` check in the actual `TensorMap` constructor. No strong feelings about this one though.

ext/TensorKitChainRulesCoreExt/constructors.jl
lkdvos: I don't think that works, because then the spaces are missing in the dictionary-to-tensormap converter. Maybe the comment is a bit misleading -- all fields in the dictionary are required, but Zygote tends to drop fields that do not contribute (i.e. codomain and domain). Thus, this uses the dictionary from the forwards pass with the data from the backwards pass.

Jutho: Ok. I wanted to avoid copying the `data` from `out`, but probably `copy(out)` is a shallow copy that does not duplicate `out[:data]`?

lkdvos: Indeed, it should really only duplicate the pointer, which I think is acceptable. On top of that, I am not sure this rule is ever used/useful anyways. I think I copied this at some point from maarten, but presumably that was also only introduced for testing purposes, as it also suffers from the weird interplay of "inner product on the parameters" for non-abelian symmetries.

PR #134, opened 2024-06-28
Soften type restrictions on tsvd kwargs

lkdvos:
This implementation ensures that additional algorithms can be defined for the `tsvd` implementation.

### DIFF
--- a/src/tensors/factorizations.jl
+++ b/src/tensors/factorizations.jl
@@ -246,10 +246,7 @@ function LinearAlgebra.isposdef(t::AbstractTensorMap, (p₁, p₂)::Index2Tuple)
     return isposdef!(permute(t, (p₁, p₂); copy=true))
 end
 
-function tsvd(t::AbstractTensorMap; trunc::TruncationScheme=NoTruncation(),
-              p::Real=2, alg::Union{SVD,SDD}=SDD())
-    return tsvd!(copy(t); trunc=trunc, p=p, alg=alg)
-end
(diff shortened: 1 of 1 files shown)


src/tensors/factorizations.jl
```diff
@@ -440,6 +435,8 @@ function tsvd!(t::TensorMap;
     else
         truncerr = abs(zero(scalartype(t)))
         W = S(dims)
+        # TODO: do we really want this behaviour? this changes the arrows of the S legs, but
+        # only sometimes?
```
Jutho: Good question. I think the idea was that if either U or V can be interpreted as simply a unitary basis transform on the domain or the codomain, we make them as such. But this can actually be extended even if U or V are isometries. As soon as either N1 or N2 are 1, S could inherit the arrow from there. Not sure if that is something useful.

lkdvos: Yeah, in general it is a bit of a strange question. I think I would prefer it if we just fix the arrow, i.e. result in a non-dual space. I mostly don't like that the behaviour is dependent on the truncation scheme, which I think might be confusing/inconsistent.

Jutho: That is definitely true. I also don't see why that was necessary. The only case where I could see that it is potentially useful to have other arrows is when either the domain and/or codomain of `t` have just one space, and you want `U` or `V` to represent a linear map on that space. But this works as well in the truncated case. 

But I am certainly fine with just having a fixed convention, so feel free to remove these lines.

PR #135, opened 2024-07-15
Progress towards a version 1

lkdvos:


TensorKitSectors/src/precompile.jl
```diff
@@ -0,0 +1,32 @@
...
+    precompile(⊗, (I, I))
+    precompile(⊗, (I, I, I))
+    precompile(⊗, (I, I, I, I))
+
+    precompile(FusionStyle, (I,))
+    precompile(BraidingStyle, (I,))
+
+    precompile(dim, (I,))
+    precompile(sqrtdim, (I,))
+    precompile(isqrtdim, (I,))
```
Jutho: I don't remember how I came to call this `isqrtdim` with `i` for `inv`(erse). It seems like `Base.isqrt` is for integer square root. I wouldn't be opposed to making this `invsqrtdim` moving forward.

src/fusiontrees/fusiontrees.jl
```diff
@@ -203,8 +196,9 @@ function Base.convert(A::Type{<:AbstractArray}, f::FusionTree{I,N}) where {I,N}
     d1 = size(C1)
     X = similar(C1, (d1[1], d1[2], Base.tail(dtail)...))
     trivialtuple = ntuple(identity, Val(N))
-    return TO.tensorcontract!(X, ((trivialtuple..., N + 1), ()), C1, ((1, 2), (3,)), :N,
-                              Ctail, ((1,), Base.tail(trivialtuple)), :N, true, false)
+    return TO.tensorcontract!(X, C1, ((1, 2), (3,)), false,
+                              Ctail, ((1,), Base.tail(trivialtuple)), false,
+                              ((trivialtuple..., N + 1), ()), One(), Zero())
```
Jutho: I guess the last two arguments can be omitted.

src/planar/postprocessors.jl
```diff
@@ -51,32 +51,26 @@ end
...
+            return Expr(ex.head, GlobalRef(TensorKit, Symbol(:planarcontract!)),
                         map(_insert_planar_operations, ex.args[2:end])...)
         elseif ex.args[1] == GlobalRef(TensorOperations, :tensortrace!)
             conjA = popat!(ex.args, 6)
-            @assert conjA == :(:N) "conj flag should be `:N` ($conjA)"
-            return Expr(ex.head, GlobalRef(TensorKit, Symbol(:_planartrace!)),
+            @assert !conjA "conj flag should be disabled"
+            return Expr(ex.head, GlobalRef(TensorKit, Symbol(:planartrace!)),
                         map(_insert_planar_operations, ex.args[2:end])...)
         elseif ex.args[1] in TensorOperations.tensoroperationsfunctions
```
Jutho, line 75: I am confused by this. Why do `tensoradd!` & friends are already of the `GlobalRef` form (i.e. we check `ex.args[1] == GlobalRef(TensorOperations, :tensoradd!)` but would the other functions in `tensoroperationsfunction` not yet have the `GlobalRef` asssignment, and need to have it inserted by this postprocessor? It's old code, so it's probably a question to myself 😄 .

src/tensors/abstracttensor.jl
```diff
@@ -3,53 +3,53 @@
...
 """
-    abstract type AbstractTensorMap{S<:IndexSpace, N₁, N₂} end
+    abstract type AbstractTensorMap{E<:Number, S<:IndexSpace, N₁, N₂} end
 
-Abstract supertype of all tensor maps, i.e. linear maps between tensor products
-of vector spaces of type `S<:IndexSpace`. An `AbstractTensorMap` maps from
-an input space of type `ProductSpace{S, N₂}` to an output space of type
-`ProductSpace{S, N₁}`.
+Abstract supertype of all tensor maps, i.e. linear maps between tensor products of vector
+spaces of type `S<:IndexSpace`, with element type `E`. An `AbstractTensorMap` maps from an
```
Jutho: I would be up for consistently referring to `E` as the scalar type. I see why you chose `E` for element type, but I guess there is no conflict with calling it `T` as in the `AbstractArray{T,N}` convention?

src/planar/postprocessors.jl
lkdvos: The other functions in `tensoroperations` do already have this, except for the ones that get added by different postprocessors (eg `free_temporaries`). In principle, the postprocessor that adds the `GlobalRef`s should just come last, but it is already part of the `TensorParser()` call so it's a bit inelegant :upside_down_face:

src/tensors/abstracttensor.jl
lkdvos: I think I did not go for `T` as that might be confusing with `TensorMap`, but I am more than happy to consistently change it back to `T`.

src/planar/postprocessors.jl
Jutho: Oh right, I see, postprocessor steps that the planar parser adds after the `addtensoroperations` postprocessor that is by default present in `TensorParser`.

src/planar/postprocessors.jl
```diff
@@ -88,16 +82,24 @@ function _insert_planar_operations(ex)
...
-       ex.args[1].mod == TensorKit &&
-       ex.args[1].name ∈ _PLANAR_OPERATIONS
-        b = Backend{backend}()
-        return Expr(:call, ex.args..., b)
-    elseif isa(ex, Expr)
-        return Expr(ex.head, (insert_operationbackend(e, backend) for e in ex.args)...)
-    else
-        return ex
-    end
+"""
```
Jutho, line 85: These are not yet used currently? This is in preparation of further changes to the planar parser?

src/tensors/abstracttensor.jl
```diff
@@ -3,53 +3,53 @@
...
-Base.eltype(::Union{T,Type{T}}) where {T<:AbstractTensorMap} = scalartype(T)
+Base.eltype(::Type{<:AbstractTensorMap{E}}) where {E} = E
 
 """
     spacetype(::Union{T,Type{T}}) where {T<:AbstractTensorMap} -> Type{S<:IndexSpace}
 
 Return the type of the elementary space `S` of a tensor.
 """
-spacetype(::Type{<:AbstractTensorMap{S}}) where {S<:IndexSpace} = S
+spacetype(::Type{<:AbstractTensorMap{E,S}}) where {E,S<:IndexSpace} = S
```
Jutho: Could also be `spacetype(::Type{<:AbstractTensorMap{<:Any,S}}) where {S<:IndexSpace} = S` I guess, but this is ok

src/tensors/abstracttensor.jl
```diff
@@ -97,7 +97,7 @@ similarstoragetype(t::AbstractTensorMap, T) = similarstoragetype(typeof(t), T)
 const order = numind
 
 @doc """
-    codomain(t::AbstractTensorMap{S,N₁,N₂}, [i::Int]) -> ProductSpace{S,N₁}
+    codomain(t::AbstractTensorMap{E,S,N₁,N₂}, [i::Int]) -> ProductSpace{S,N₁}
```
Jutho: Documentation here is a bit inconsistent: should probably be
```
@doc """
    codomain(t::AbstractTensorMap{E,S,N₁,N₂}) -> ProductSpace{S,N₁}
    codomain(t::AbstractTensorMap{E,S,N₁,N₂}, i::Int) -> S
```

src/tensors/abstracttensor.jl
```diff
@@ -167,13 +167,13 @@ Return all indices of a tensor, i.e. the indices of its domain and codomain.
 
 See also [`codomainind`](@ref) and [`domainind`](@ref).
 """
-function allind(::Type{<:AbstractTensorMap{<:IndexSpace,N₁,N₂}}) where {N₁,N₂}
-    return ntuple(n -> n, N₁ + N₂)
+function allind(::Type{T}) where {T<:AbstractTensorMap}
+    return ntuple(n -> n, numind(T))
```
Jutho: `n->n` is `identity` like above?

src/tensors/abstracttensor.jl
```diff
@@ -226,6 +226,55 @@ Return the dimensions of the block of a tensor corresponding to a coupled sector
...
+# 1 argument
+Base.similar(t::AbstractTensorMap) = similar(t, storagetype(t), space(t))
+
+# generic implementation for AbstractTensorMap -> returns `TensorMap`
+function Base.similar(::AbstractTensorMap, ::Type{TorA},
+                      P::TensorMapSpace{S}) where {TorA<:MatOrNumber,S}
+    N₁ = length(codomain(P))
+    N₂ = length(domain(P))
+    TT = tensormaptype(S, N₁, N₂, TorA)
+    return TT(undef, codomain(P), domain(P))
```
Jutho, line 278: `TT(undef, P)` ? I guess it depends on which constructor is "the most fundamental"

src/tensors/braidingtensor.jl
```diff
@@ -125,9 +125,9 @@ end
     end
 end
 
-Base.similar(::BraidingTensor, T::Type, P::TensorMapSpace) = TensorMap(undef, T, P)
+# efficient copy constructor
+Base.copy(b::BraidingTensor{E,S}) where {E,S} = BraidingTensor{E,S}(b.V1, b.V2, b.adjoint)
```
Jutho: Since all the field values should be immutable, is this not just `copy(b::BraidingTensor) = b` ? Do we use this somewhere?

src/tensors/braidingtensor.jl
```diff
@@ -209,30 +210,40 @@ end
...
 
-function planarcontract!(C::AbstractTensorMap{S,N₁,N₂},
-                         A::BraidingTensor{S},
-                         (oindA, cindA)::Index2Tuple{2,2},
-                         B::AbstractTensorMap{S},
-                         (cindB, oindB)::Index2Tuple{2,N₃},
-                         (p1, p2)::Index2Tuple{N₁,N₂},
+function planarcontract!(C::AbstractTensorMap,
+                         A::BraidingTensor,
+                         (oindA, cindA)::Index2Tuple,
```
Jutho, line 235: I am confused how this can be valid without the specific `(oindA, cindA)::Index2Tuple{2,2}` part? The implementation for `Bosonic` below really assumes that you are exactly permuting two indices, and not have e.g. 3 indices contracted with the braiding tensor (corresponding to a trace operation).

Jutho: Oh I see, you've added some check below. Never mind

src/tensors/indexmanipulations.jl
```diff
@@ -87,17 +85,17 @@ which determines whether they will braid over or under any other index with whic
...
-@propagate_inbounds function braid!(tdst::AbstractTensorMap{S,N₁,N₂},
-                                    tsrc::AbstractTensorMap{S},
-                                    (p₁, p₂)::Index2Tuple{N₁,N₂},
-                                    levels::IndexTuple) where {S,N₁,N₂}
-    return add_braid!(tdst, tsrc, (p₁, p₂), levels, true, false)
+@propagate_inbounds function braid!(tdst::AbstractTensorMap,
+                                    tsrc::AbstractTensorMap,
+                                    p::Index2Tuple,
+                                    levels::IndexTuple)
+    return add_braid!(tdst, tsrc, p, levels, true, false)
```
Jutho: `true, false` -> `One(), Zero()` ?

src/tensors/indexmanipulations.jl
```diff
@@ -1,25 +1,24 @@
...
                 
 See [`permute`](@ref) for creating a new tensor and [`add_permute!`](@ref) for a more general version.
 """
-@propagate_inbounds function Base.permute!(tdst::AbstractTensorMap{S,N₁,N₂},
-                                           tsrc::AbstractTensorMap{S},
-                                           p::Index2Tuple{N₁,N₂}) where {S,N₁,N₂}
+@propagate_inbounds function Base.permute!(tdst::AbstractTensorMap,
+                                           tsrc::AbstractTensorMap,
+                                           p::Index2Tuple)
     return add_permute!(tdst, tsrc, p, true, false)
```
Jutho: `true, false` -> `One(), Zero()` ? It might be that we check for this later anyway.

src/tensors/tensor.jl
```diff
@@ -219,101 +239,120 @@ function _buildblockstructure(P::ProductSpace{S,N}, blocksectors) where {S<:Inde
...
+            fill!(t, $felt(T))
+            return t
+        end
+    end
 end
 
-function TensorMap(dataorf, T::Type{<:Number}, P::TensorMapSpace{S}) where {S<:IndexSpace}
-    return TensorMap(dataorf, T, codomain(P), domain(P))
-end
+for randfun in (:rand, :randn, :randexp)
```
Jutho, line 278: `randhaar`/`randisometry` is lost in this process, but that one is anyway slightly different, since it needs to be interpreted at the level of whole blocks rather than elementswise, and as such is also harder to be applied in place.

src/tensors/abstracttensor.jl
lkdvos: the problem here is that this is a fully specified tensormap type, and I did not add a constructor for this. The "fundamental" one takes domain and codomain as separate arguments, as these are separate fields in the struct as well. I actually would prefer to store the homspace, and to make that one the most fundamental, but I didn't want to single-handedly decide this :)

Jutho: Ok, let's keep that for when we are completely changing and breaking the TensorMap structure.

PR #140, opened 2024-07-29
WIP: redesign tensormap structure

Jutho:
Plan:

- [x] New fusion tree iterator, that accepts sector iterators for outgoing and uses new ordering that sorts fusion trees such that left most values are changing the quickest in the list: outgoing[1], outgoing[2], vertex[1], inner[1], outgoing[3], vertex[2], inner[2], …
- [x] Encode all structure information at the level of the `HomSpace` and experiment with different strategies for caching this information
- [x] Store all the tensor data in one sequential DenseVector
- [ ] (optional): Experiment whether it is beneficial to use a `Memory` object in Julia 1.11 and beyond.


Jutho:
I've been continuing with this PR and will soon push an update. One set of questions that comes up is what we do with indexing, in particular to `TrivialTensorMap`.
* Defining `TrivialTensorMap` as a type alias is already less convenient, since the sectortype is no longer part of the parameter list of `TensorMap`. I am hesitant to make unions such as `S<:Union{ComplexSpace,CartesianSpace,GeneralSpace}` since that is not extensible, so I would just get rid of the type alias.
* Indexing into `TrivialTensorMap` can use no argument `t[]` (return the data as N=N1+N2-dimensional tensor), `t[i1,...,iN]` (returning a scalar, equivalent to t[][i1,...,iN]), `t[(nothing,nothing)]`, `t[(nothing,nothing)]` where `nothing is a substitute for the trivial fusion tree (and thus `t[(nothing,nothing)] == t[]`). 

The question is whether we want to keep all of those. Without a type alias, they would start with `sectortype(t) == Trivial || throw(error...)`.

Jutho:
Another question is what the interface needs to be to get the data vector of a `TensorMap`; do we just use `t.data` or do we want an explicit interface for that. Before the data was the list of blocks, and so `blocks(t)` and `block(t,c)` was the public interface to access the raw data. However, simple vector space methods, in particular everything from VectorInterface.jl, can now be rewritten to act directly on the data vector without needing any of the structure info. So acting on the `data` field is quite common.

Jutho:
I am all in favor of getting rid of the `(nothing, nothing)` indexing.

For the nonsymmetric / trivially symmetric tensors, the identity

`t[scalarindices...] == t[][scalarindices...]`

implies that maybe the left one is not strictly necessary, as it is still accessible (in an even more flexible way that also allows slicing etc) as long als `t[]` still works. However, maybe we rather want to deprecate instead `t[]`. 

The reason that I would consider providing an interface to at least get the data, is that unfortunately there are many external libraries that only work on `AbstractArray` or `AbstractVector` instances, e.g. DifferentialEquations.jl, several nonlinear optimisation libraries, … . At least with the new data structure, it will become easier to use these in combination with `TensorMap` instances, as it will be a matter of doing

`some_external_function(f, t.data)`

where `f` is a function that encodes an operation on tensors, but now needs to be written as

```julia
function f(data)
 # reconstruct tensor
 t = TensorMap(data, space)
 # act
 tnew = # act on t ...
 # return data in case output is also a tensor (e.g. ODE right hand side)
 return tnew.data
end
```

We could even go one step further and make `AbstractTensorMap{T} <: AbstractVector{T}` with `length(t) = dim(t)`. In that case, `getindex` and `setindex!` with a single index need to be supported and acts directly on `t.data`.

So I think this requires some careful deliberation weighing the pros and cons of different approaches.

lkdvos:
I think for the question of `t[][inds...]` vs `t[inds...]`, my preference goes to the latter. The main reasons being that it is just a little more intuitive, it is a correct overload of `Base.getindex`, and is more intuitive for people that are used to Array objects. Additionally, `t[]` necessarily has to work with views, which can easily become confusing, and while `diagonalt[inds] = val` could in principle be made to work, this becomes a lot more difficult with `diagonalt[][inds] = val`.
I have no real problem with having to copy over some of the AbstractArray implementations to make these convenience methods work.

I am not a big fan of `AbstractTensorMap <: AbstractVector` in the way you describe, because really that only applies to `TensorMap`. For `AdjointTensorMap`, this now has a very weird interplay with what `getindex` should do, as the data is now non-trivially permuted. For `DiagonalTensorMap` you in principle no longer have `dim(diagt) = length(diagt.data)`, and for `BlockTensorMap` this becomes a big hassle to make the indexing work as wanted because now it would be an AbstractArray, but not in the way that is intuitive. I think it would create more problems than it would solve.

As a sidenote, we already have some functionality for the type of functions you describe through `FiniteDifferences.to_vec`:
```julia
t0 = input_tensor
v0, from_vec = to_vec(to)
function my_wrapper_fun(v::Vector)
    t = from_vec(v)
    [...]
    return to_vec(result)[1]
end
```

I also just realised that this allows a little more fine-grained control over the fact that our data does not have the inner product you would expect whenever non-abelian symmetries are involved, i.e. `inner(x.data, y.data) != inner(x, y)` in general. This is a really subtle problem that makes me even more hesitant to make the data part of the interface :upside_down_face:

Jutho:
There are all good points; the `AbstractTensorMap <: AbstractVector` suggestions was just something I wanted to throw out there, not something I would myself be in big favor of. But I do think that it is important for further progress that we can easily interface with libraries that only accept array-like data, and I do think we should have a public interface for that. I am not too concerned about `AdjointTensorMap` and `BlockTensorMap`; in the cases where you want to use e.g. ODE solvers or optimisation methods, the tensors that you will act on will be normal `TensorMap`s I assume. I don't think that situation is dramatically different from some of the `AbstractArray` wrapper or otherwise custom types.

Regarding the inner product; in principle we could consider absorbing a factor of `sqrt(dim(c))` in the data, and then taking it out where needed. Not sure how much overhead that would create and whether it would be compensated by its benefits.

src/auxiliary/linalg.jl
```diff
@@ -66,21 +66,10 @@ using LinearAlgebra: BlasFloat, BlasReal, BlasComplex, checksquare
...
-    Threads.@threads for j in 1:size(A, 2)
-        @simd for i in 1:size(A, 1)
-            @inbounds A[i, j] = i == j
-        end
-    end
+function one!(A::StridedMatrix)
+    A[:] .= 0
+    A[diagind(A)] .= 1
     return A
 end
```
lkdvos: Throughout LinearAlgebra, this seems to be the preferred option (which automatically works for CUDA):
```suggestion
function one!(A::StridedMatrix)
    length(A) > 0 || return A
    copyto!(A, LinearAlgebra.I)
    return A
end
```

src/fusiontrees/iterator.jl
```diff
@@ -59,105 +68,178 @@ function Base.iterate(it::FusionTreeIterator{I,0},
...
-    return f, (lines, vertices, states)
+    out, lines, vertices, states = next
+    vertexlabels = _labelvertices(out, it.coupled, lines, vertices)
+    f = FusionTree(out, coupled, it.isdual, lines, vertexlabels)
+    return f, (out, lines, vertices, states)
 end
 
-function labelvertices(uncoupled::NTuple{2,I}, coupled::I, lines::Tuple{},
-                       vertices::Tuple{Int}) where {I<:Sector}
+@inline function _labelvertices(uncoupled::NTuple{2,I}, coupled::I, lines::Tuple{},
```
lkdvos: Do you know if the vertexlabeltype is ever really used? It adds quite a bit of complexity, and I don't really see the benefit. Maybe we could consider just removing that feature alltogether, as these really are necessarily integers in `1:NSymbol`?

src/spaces/homspace.jl
```diff
@@ -134,3 +134,107 @@ function compose(W::HomSpace{S}, V::HomSpace{S}) where {S}
...
+    # if FusionStyle(I) === UniqueFusion()
+    #     return TaskLocalCache{SectorDict{I,Any}}()
+    # else
+    #     return GlobalCache()
+    # end
+end
+
+tensorstructure(W::HomSpace) = tensorstructure(W, CacheStyle(sectortype(W)))
+
+function tensorstructure(W::HomSpace, ::NoCache)
```
lkdvos: In principle, this name clashes with `TensorOperations.tensorstructure`, which returns the structure associated with the indices. I do think that the name is appropriate here though, so maybe it is okay to keep this.

src/spaces/homspace.jl
```diff
@@ -134,3 +134,107 @@ function compose(W::HomSpace{S}, V::HomSpace{S}) where {S}
...
+    outer_offset = 0
+    for c in blocksectors(W)
+        inner_offset₂ = 0
+        inner_offset₁ = 0
+        for f₂ in fusiontrees(dom, c)
+            s₂ = f₂.uncoupled
+            d₂ = dim(dom, s₂)
+            r₂ = (inner_offset₂ + 1):(inner_offset₂ + d₂)
+            inner_offset₂ = last(r₂)
+            # TODO:  # now we run the code below for every f₂; should we do this separately
```
lkdvos: This is definitely an optimization I did in TensorTrack, it might be worth it to separate that out (although if it is cached it should also not matter too much)

src/tensors/abstracttensor.jl
```diff
@@ -38,7 +38,7 @@ spacetype(::Type{<:AbstractTensorMap{<:Any,S}}) where {S} = S
 
 Return the type of sector `I` of a tensor.
 """
-sectortype(::Type{TT}) where {TT<:AbstractTensorMap} = sectortype(spacetype(TT))
+sectortype(::Type{<:AbstractTensorMap{<:Any,S}}) where {S} = sectortype(S)
```
lkdvos: Is there a specific reason for this change? I tens to avoid accessing type parameters as much as possible, as this is very annoying to change if ever the order changes, hence the previous definition in terms of a function instead of accessing the type parameters. It does not really matter too much, just curious if there was a specific benefit

src/tensors/abstracttensor.jl
```diff
@@ -205,29 +216,90 @@ See also [`blocks`](@ref), [`blocksectors`](@ref), [`blockdim`](@ref) and [`hasb
...
+# auxiliary function
+@inline function trivial_fusiontree(t::AbstractTensorMap)
+    sectortype(t) === Trivial ||
+        throw(SectorMismatch("Only valid for tensors with trivial symmetry"))
+    spaces1 = codomain(t).spaces
+    spaces2 = domain(t).spaces
+    f₁ = FusionTree{Trivial}(map(x -> Trivial(), spaces1), Trivial(), map(isdual, spaces1))
+    f₂ = FusionTree{Trivial}(map(x -> Trivial(), spaces2), Trivial(), map(isdual, spaces2))
+    return (f₁, f₂)
+end
```
lkdvos, line 284: just out of curiosity, do you know if `only(fusiontrees(t))` gives the same result? Is there a noticable speedup?

src/tensors/adjoint.jl
```diff
@@ -6,100 +6,38 @@
 Specific subtype of [`AbstractTensorMap`](@ref) that is a lazy wrapper for representing the
 adjoint of an instance of [`AbstractTensorMap`](@ref).
 """
-struct AdjointTensorMap{T,S,N₁,N₂,TT<:AbstractTensorMap{T,S,N₂,N₁}} <:
+struct AdjointTensorMap{T,S,N₁,N₂,TT<:AbstractTensorMap{T,S,N₁,N₂}} <:
```
lkdvos: Is this correct? Might be a typo?

src/tensors/adjoint.jl
```diff
@@ -6,100 +6,38 @@
...
-const AdjointTrivialTensorMap{T,S,N₁,N₂,TT<:TrivialTensorMap{T,S,N₂,N₁}} =
-    AdjointTensorMap{T,S,N₁,N₂,TT}
-#! format: on
-
 # Constructor: construct from taking adjoint of a tensor
+Base.adjoint(t::AdjointTensorMap) = t.parent
 Base.adjoint(t::AbstractTensorMap) = AdjointTensorMap(t)
-Base.adjoint(t::AdjointTensorMap) = parent(t)
-
-Base.parent(t::AdjointTensorMap) = t.parent
```
lkdvos: I usually find it a little clearer to read code in terms of `t` and `parent(t)` instead of `t` and `t'`, as the latter is a bit confusing since `adjoint(t)` is actually the non-adjoint version. No hard opinions here though

src/tensors/linalg.jl
```diff
@@ -83,10 +90,17 @@ error will be thrown.
...
     codomain(V) ≅ domain(V) ||
         throw(SpaceMismatch("codomain and domain are not isomorphic: $V"))
-    t = tensormaptype(S, N₁, N₂, A)(undef, codomain(V), domain(V))
+    if A <: Number
+        t = TensorMap{A}(undef, V)
+    elseif A <: DenseVector
+        T = scalartype(A)
+        t = TensorMap{T,S,N₁,N₂,A}(undef, V)
+    else
+        throw(ArgumentError("`isomorphism` only supports Number or DenseVector subtypes as first argument"))
```
lkdvos: I think this is precluded by the function signature

src/tensors/linalg.jl
```diff
@@ -121,17 +135,24 @@ end
...
     domain(V) ≾ codomain(V) ||
         throw(SpaceMismatch("$V does not allow for an isometric inclusion"))
-    t = tensormaptype(S, N₁, N₂, A)(undef, codomain(V), domain(V))
+    if A <: Number
+        t = TensorMap{A}(undef, V)
+    elseif A <: DenseVector
+        T = scalartype(A)
+        t = TensorMap{T,S,N₁,N₂,A}(undef, V)
+    else
+        throw(ArgumentError("`isometry` only supports Number or DenseVector subtypes as first argument"))
```
lkdvos: Same comment here

src/tensors/linalg.jl
```diff
@@ -172,15 +193,11 @@ LinearAlgebra.isdiag(t::AbstractTensorMap) = all(LinearAlgebra.isdiag, values(bl
 # Copy, adjoint! and fill:
 function Base.copy!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap)
     space(tdst) == space(tsrc) || throw(SpaceMismatch("$(space(tdst)) ≠ $(space(tsrc))"))
-    for c in blocksectors(tdst)
-        copy!(StridedView(block(tdst, c)), StridedView(block(tsrc, c)))
-    end
+    copy!(tdst.data, tsrc.data)
```
lkdvos, line 194: This is not really an `AbstractTensorMap` implementation, but rather a `TensorMap` implementation, so I would tend to keep the previous definition and just specialize a version for TensorMap

src/tensors/linalg.jl
```diff
@@ -172,15 +193,11 @@ LinearAlgebra.isdiag(t::AbstractTensorMap) = all(LinearAlgebra.isdiag, values(bl
...
     return tdst
 end
 function Base.fill!(t::AbstractTensorMap, value::Number)
-    for (c, b) in blocks(t)
-        fill!(b, value)
-    end
+    fill!(t.data, value)
```
lkdvos, line 204: Here the same comment applies

src/tensors/vectorinterface.jl
lkdvos, line 1: Here again, I don't think it is fair to have these implementations for `AbstractTensorMap`, these should be specializations for `TensorMap`

src/tensors/tensor.jl
```diff
@@ -3,152 +3,114 @@
...
-        M = Matrix{TorA}
-        T = TorA
-    else
-        throw(ArgumentError("the final argument of `tensormaptype` should either be the scalar or the storage type, i.e. a subtype of `Number` or of `DenseMatrix`"))
-    end
-    if I === Trivial
-        return TensorMap{T,S,N₁,N₂,I,M,Nothing,Nothing}
+# TODO: should we deprecate this?
+# It seems quite heavily used in MPSKit.jl, although they are all of the form
+# `tensormaptype(S, N₁, N₂, A)` and so could thus be replaced with `TensorMap{S, N₁, N₂, A}`
```
lkdvos: We could, but it should then be `TensorMap{scalartype(A),S,N1,N2,A}`. I have no strong opinion about this, but I quite like that I don't have to think about `storagetype` vs `scalartype`, so it does serve a purpose.

src/tensors/tensor.jl
```diff
@@ -157,89 +119,28 @@ Construct a `TensorMap` by explicitly specifying its block data.
...
-    blocksectoriterator = blocksectors(codom ← dom)
-    for c in blocksectoriterator
+    
+    T = eltype(valtype(data))
+    t = TensorMap{T}(undef, V)
+    for (c, b) in blocks(t)
         haskey(data, c) || throw(SectorMismatch("no data for block sector $c"))
+        datac = data[c]
+        size(datac) == size(b) || throw(DimensionMismatch("wrong size of block for sector $c"))
+        copy!(b, datac)
```
lkdvos: If we instead have `setindex!(t::AbstractTensorMap, c::Sector)` we could make the size check and haskey check automatic

src/tensors/tensor.jl
```diff
@@ -626,20 +440,31 @@ Return a view into the data slice of `t` corresponding to the splitting - fusion
...
+        haskey(structure.fusiontreeindices, (f₁, f₂)) || throw(SectorMismatch())
     end
     @inbounds begin
+        i = structure.fusiontreeindices[(f₁, f₂)]
+        c = f₁.coupled
+        (d₁, d₂), r = structure.blockstructure[c]
+        r₁, r₂ = structure.fusiontreeranges[i]
         d = (dims(codomain(t), f₁.uncoupled)..., dims(domain(t), f₂.uncoupled)...)
-        return sreshape(StridedView(t.data[c])[t.rowr[c][f₁], t.colr[c][f₂]], d)
+        return sreshape(sreshape(sview(t.data, r), (d₁, d₂))[r₁, r₂], d)
```
lkdvos: This could be slightly simplified by re-using the `block(t, c)` functionality. I am also slightly surprised by this, as I had expected the intermediate step of first unpacking the "matrixblock" and only then unpacking the "fusiontreeblock" could now be avoided by correctly adjusting the fusiontreeranges.

src/spaces/homspace.jl
lkdvos: As a second thought, maybe we could emphasize that this is tied to the specific TensorMap implementation by changing it to `tensormapstructure`?

src/tensors/tensor.jl
```diff
@@ -365,50 +266,71 @@ Construct a `TensorMap` from a plain multidimensional array.
...
+
+In case 2 and 3, the `TensorMap` constructor will reconstruct the tensor data such that the
+resulting tensor `t` satisfies `data == convert(Array, t)`, up to an error specified by `tol`.
+For the case where `sectortype(S) == Trivial` and `data isa DenseArray`, the `data` array is
+simply reshaped into a vector and used as in case 1 so that the memory will still be shared.
+In other cases, new memory will be allocated.
+
+Note that in the case of `N₁ + N₂ = 1`, case 3 also amounts to `data` being a vector, whereas
+when `N₁ + N₂ == 2`, case 2 and case 3 both require `data` to be a matrix. Such ambiguous cases
+are resolved by checking the size of `data` in an attempt to capture all support all possible
```
```suggestion
are resolved by checking the size of `data` in an attempt to support all possible
```

src/tensors/tensor.jl
```diff
@@ -771,20 +589,21 @@ function Base.convert(::Type{TensorMap}, t::AbstractTensorMap)
...
     end
 end
 
 function Base.promote_rule(::Type{<:TT₁},
                            ::Type{<:TT₂}) where {S,N₁,N₂,
                                                  TT₁<:TensorMap{<:Any,S,N₁,N₂},
                                                  TT₂<:TensorMap{<:Any,S,N₁,N₂}}
-    return tensormaptype(S, N₁, N₂, promote_type(storagetype(TT₁), storagetype(TT₂)))
+    T = promote_type(scalartype(TT₁), scalartype(TT₂))
+    A = promote_type(storagetype(TT₁), storagetype(TT₂))
```
lkdvos: I am not sure if this can ever pose problems, but we could ensure compatibility between T and A by:
```suggestion
    A = promote_type(storagetype(TT₁), storagetype(TT₂))
    T = scalartype(A)
```

src/fusiontrees/iterator.jl
Jutho: That's probably a good idea. I don't know why I thought that people would ever come up with custom names for this.

src/auxiliary/linalg.jl
Jutho: I thought my approach also worked for CUDA. I am happy with the suggestion as well, if it works for all cases. I think I had problems with using `I` in the CUDA context before.

src/tensors/abstracttensor.jl
Jutho: No probably not, I'll revert this.

src/tensors/adjoint.jl
Jutho: Yes, definitely is. Result from me first changing the whole adjoint implementation (kind of redoing the work that you already did), and then only afterwards rebasing on master that included your work.

src/tensors/abstracttensor.jl
Jutho: I was indeed doubting about this. Calling `fusiontrees(t)` requires a cache lookup; whereas this is probably almost completely compiled away.

src/tensors/tensor.jl
Jutho: ok, fair enough, I keep it in and can probably use it at other places.

src/tensors/tensor.jl
Jutho: ok, that's part of the indexing discussion then. So you would prefer to also do block access via `getindex` and `setindex!`? I was a bit hesitant in channeling all these different functionalities through the same interface.

src/tensors/tensor.jl
Jutho: Ok, I see what you mean. I went this way because the `r₁, r₂` are easier to compute and can be computed directly from the `ProductSpace` codomain and domain separately. In the resulting `StridedView`, that indeed translates to an offset in the global data vector, a first dimension `length(r₁)`, a first stride given by `size(block(t, c))`, and a second dimension `length(r₂)`. So your suggestion is to cache that data directly? It is less natural from the way that info is computed, but yes, it can be done of course.

Jutho: Also I did not reuse `block(t,c)` since that uses `Base.view` and `reshape`, so then there is an extra conversion step into `StridedView`. Probably the same cost but yeah...

src/tensors/tensor.jl
Jutho: Neither of the two are probably a good idea:
```julia-repl
julia> promote_type(Vector{Float64}, Vector{ComplexF32})
Vector (alias for Array{T, 1} where T)
```

src/tensors/vectorinterface.jl
Jutho: Yes indeed. Damn. I do not want to implement this at the block level since it is an advantage to simply act on the data as one chunck. But I would also prefer to write a general implementation if possible. Any suggestions?

Jutho:
Another question that comes up as I continue with this PR is that all the linear algebra factorisations and things like applying functions to 'square' TensorMaps, current implementation works as:

* apply operation to individual blocks
* collect the output blocks (possibly of different output arguments, e.g. in the case of factorisations)
* construct output tensormaps, thereby directly using the blocks as the actual data

With the new approach, you cannot construct the tensormap from the blocks in a data sharing way, so you allocate new data and copy the blocks into this. This causes (a potentially great deal of) extra allocations.

Maybe this can be resolved by writing mutating versions of these operations in `MatrixAlgebra` that accept the output blocks as arguments? I don't know if there are better suggestions?

src/fusiontrees/iterator.jl
Jutho: Looking at removing this, it seems like this would never have worked with anything other than integers. The values in `f.vertices` are used to index into `Fsymbol` and `Rsymbol`, so they need to be integers and not labels :-)

Jutho: @lkdvos , can you look at removing any reference to vertex labels in TensorKitSectors.jl?

src/tensors/adjoint.jl
Jutho: I saw some Github issue or discourse post of someone complaining that the semantics of parent were not well defined or something (I did not read the thread). That's why I tried to avoid it, but I will revert this.

src/tensors/linalg.jl
Jutho: Yeah, I first thought of getting rid of the type restriction all together, which is currently done at some places, so I guess we have to be consistent. I think I would actually like the error message rather than a generic MethodError, so I think I will get rid of the type restriction.

src/tensors/vectorinterface.jl
Jutho: Ok, fix is on the way, but will consist of generic AbstractArray implementation based on `blocks` and specialised `TensorMap` implementations.

lkdvos: I think it is fine to implement both a `TensorMap` and `AbstractTensorMap` version, the code duplication seems to be somewhat small anyways.

src/tensors/tensor.jl
lkdvos: https://discourse.julialang.org/t/type-promototion-of-vectors/30227/6
https://github.com/JuliaLang/julia/blob/e52a46c5c192bfe16853ae0b63ac33b280fba063/base/range.jl#L1282

It seems like this is not really something that the Base promotion system is designed to resolve. From
what I can find, the current behaviour is to only promote if at least one of the eltypes would not change, which is probably not what we want.

I would maybe suggest inserting our own `promote_storagetype`, which defaults to `Vector{promote_scalartype}` and we can then further specialise however we want? It seems like this is something that would be resolved with the new `Memory` types too, as we can then just hardcode that and only have to deal with the different `addrspace` implementations for GPU/CPU...

lkdvos:
For the MatrixAlgebra, indeed, I cannot really think of a different solution. However, considering that LinearAlgebra does not expose this for these functions, it might be that the overhead of the extra copy is really just negligible for those operations? In any case, I think the code looks a bit nicer too if we don't have to first create all the block data and only then construct the output tensor, which also separates out the allocation and implementation cleanly.

```julia
for c in blocksectors(A)
    eig!((block(V, c), block(D, c)), block(A, c)) 
end
```

src/tensors/tensor.jl
lkdvos: I think I would indeed prefer this, mostly for the `setindex!` part, I like that we can avoid handing the user a view to then use in-place, and the ability to re-use some of the boundschecking functionality is also convenient. Additionally, I feel like it really is quite intuitive and a proper use of `getindex`/`setindex`, so I cannot really think of a reason not to do this, as we also do it for the fusiontrees

src/tensors/tensor.jl
Jutho: I have fixed this locally and will push it in a next commit, as committing this here is more annoying to then not have conflicts with other changes I already have locally.

src/tensors/tensor.jl
Jutho: I see no logic in the response of StefanKarpinski to Matt's question. Maybe we can just use `VectorInterface.promote_add(A1, A2)`?

lkdvos: That indeed seems to work, I think that should be fine. It does rely on type inference, but I guess we can see if it causes any issues and still change it if we have to?

Jutho:
Ok, I finally managed to get a first completely working version. Most of the factorisations are implemented as discussed above by just copying blocks into the output position. This could be done with only minimal changes to the code. Only the SVD (and truncation) implementation was changed (improved in my opinion) quite drastically. There are some things left to do, namely:

- [ ] Update CI to include Julia 1.10 and 1.11
- [ ] Fix docs (in particular, remove references to `TrivialTensorMap`)
- [ ] Investigate why my alternative implementation of `FiniteDifferences.to_vec` does not work (currently commented out)
- [ ] Benchmark the impact of this PR, and in particular the different cache choices for `fusionblockstructure(::HomSpace)`

In future PRs, the sequel to this PR needs to be:
* Streamline matrix factorisation operations so that no additional allocation is necessary, and also change the interface to support backend and allocator keywords
* Make sure that the TensorOperations backends and allocator info are passed on and correctly handled all the way down to the lowest level (e.g. `add_...`)
* Add `scheduler` keyword options, both for operations that act on the matrix block level, and operations that act on the fusion tree block levels.

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+part ``(b_1⊗b_2)⊗b_3 → c`` and the splitting part ``c→(((a_1⊗a_2)⊗a_3)⊗a_4)``. Given that
+the fusion tree can be considered to be the adjoint of a corresponding splitting tree
+``c→(b_1⊗b_2)⊗b_3``, we now first consider splitting trees in isolation. A splitting tree
+which goes from one coupled sectors ``c`` to ``N`` uncoupled sectors ``a_1``, ``a_2``, …,
+``a_N`` needs ``N-2`` additional internal sector labels ``e_1``, …, ``e_{N-2}``, and, if
+`FusionStyle(I) isa GenericFusion`, ``N-1`` additional multiplicity labels ``μ_1``,
+…, ``μ_{N-1}``. We henceforth refer to them as vertex labels, as they are associated with
+the vertices of the splitting tree. In the case of `FusionStyle(I) isa UniqueFusion`, the
+internal sectors ``e_1``, …, ``e_{N-2}`` are completely fixed, for
+`FusionStyle(I) isa MultipleFusion` they can also take different values. In our abstract
```
lkdvos: I dont think `SimpleFusion <: MultipleFusion` so maybe better:
```suggestion
the generic case they can also take different values. In our abstract
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+``\bar{a}`` and depicted with a downgoing arrow), i.e. between the dual space of ``R_a`` on
+which the conjugated irrep acts, or the irrep ``\bar{a}`` to which the complex conjugate of
+irrep ``a`` is isomorphic. This distinction is however important, when certain uncoupled
+sectors in the fusion tree actually originate from a dual space. We use the isomorphisms
+``Z_a:R_a^* → R_{\bar{a}}`` and its adjoint ``Z_a^†:R_{\bar{a}}→R_a^*``, as introduced in
+the section on [topological data of a fusion category](@ref ss_topologicalfusion), to build
+fusion and splitting trees that take the distinction between irreps and their conjugates
+into account. Hence, in the previous example, if e.g. the first and third space in the
+codomain and the second space in the domain of the tensor were dual spaces, the actual pair
+of splitting and fusion tree would look as
```
lkdvos:
```suggestion
of splitting and fusion tree would look like
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+`FusionTree` (which actually represents a splitting tree, but fusion tree is a more common
+term), defined as
+```julia
+struct FusionTree{I<:Sector,N,M,L,T}
+    uncoupled::NTuple{N,I}
+    coupled::I
+    isdual::NTuple{N,Bool}
+    innerlines::NTuple{M,I} # fixed to M = N-2
+    vertices::NTuple{L,T} # fixed to L = N-1
+end
```
lkdvos:
```suggestion
struct FusionTree{I<:Sector,N,M,L}
    uncoupled::NTuple{N,I}
    coupled::I
    isdual::NTuple{N,Bool}
    innerlines::NTuple{M,I} # fixed to M = N-2
    vertices::NTuple{L,Int} # fixed to L = N-1
end
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+Here, the fields are probably self-explanotary. The `isdual` field indicates whether an
+isomorphism is present (if the corresponding value is `true`) or not. Note that the field
+`uncoupled` contains the sectors coming out of the splitting trees, before the possible
+``Z`` isomorphism, i.e. the splitting tree in the above example would have
+`sectors = (a₁, a₂, a₃, a₄)`. The `FusionTree` type has a number of basic properties and
+capabilities, such as checking for equality with `==` and support for
+`hash(f::FusionTree, h::UInt)`, as splitting and fusion trees are used as keys in look-up
+tables (i.e. `AbstractDictionary` instances) to look up certain parts of the data of a
+tensor. The type of `L` of the vertex labels can be `Nothing` when they are not needed
+(i.e. if `FusionStyle(I) isa MultiplicityFreeFusion`).
```
lkdvos:
```suggestion
tensor.
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+length(iter)
+@elapsed sum(n->1, iter)
+@elapsed length(iter)
+s2 = s ⊠ s
+collect(fusiontrees((s2,s2,s2,s2)))
+```
+Note that `FusionTree` instances are shown (printed) in a way that is valid code to
+reproduce them, a property which also holds for both instances of `Sector` and instances of
+`VectorSpace`. All of those should be displayed in a way that can be copy pasted as valid
+code. Furthermore, we use contact to determine how to print e.g. a sector. In isolation,
```
lkdvos, line 127: I might be missing something, but what does "we use contact" mean?

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+operation in the basis of canonically ordered splitting and fusion trees. In this section,
+we will ignore the ``Z`` isomorphisms, as they are just trivially reshuffled under the
+different operations that we describe. These manipulations are used as low-level methods by
+the `TensorMap` methods discussed on the next page. As such, they are not exported by
+TensorKit.jl, nor do they overload similarly named methods from Julia Base (see `split` and
+`merge` below).
+
+The first operation we discuss is an elementary braid of two neighbouring sectors
+(indices), i.e. a so-called Artin braid or Artin generator of the braid group. Because
+these two sectors do not appear on the same fusion vertex, some recoupling is necessary.
```
lkdvos:
```suggestion
these two sectors do not necessarily appear on the same fusion vertex, some recoupling is necessary.
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+between the (co)evaluation (exact pairing) and the fusion tensors, discussed in
+[topological data of a fusion category](@ref ss_topologicalfusion). The main ingredient
+that we need is summarized in
+
+![line bending](img/tree-linebending.svg)
+
+We will only need the B-symbol and not the A-symbol. Applying the left evaluation on the
+second sector of a splitting tensor thus yields a linear combination of fusion tensors
+(when `FusionStyle(I) == GenericFusion()`, or just a scalar times the corresponding
+fusion tensor otherwise), with corresponding ``Z`` ismorphism. Taking the adjoint of this
```
lkdvos:
```suggestion
fusion tensor otherwise), with corresponding ``Z`` isomorphism. Taking the adjoint of this
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+relation yields the required relation to transform a fusion tensor into a splitting tensor
+with an added ``Z^†`` isomorphism.
+
+However, we have to be careful if we bend a line on which a ``Z`` isomorphism (or its
+adjoint) is already present. Indeed, it is exactly for this operation that we explicitly
+need to take the presence of these isomorphisms into account. Indeed, we obtain the relation
```
lkdvos:
```suggestion
need to take the presence of these isomorphisms into account, obtaining the relation
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+
+As before, there is a simplified interface for the case where
+`BraidingStyle(I) isa SymmetricBraiding` and the levels are not needed. This is simply
+given by
+
+[`permute(f1::FusionTree{I,N₁}, f2::FusionTree{I,N₂}, p1::NTuple{N₁′,Int}, p2::NTuple{N₂′,Int})`](@ref)
+
+The `braid` and `permute` routines for double fusion trees will be the main access point for
+corresponding manipulations on tensors. As a consequence, results from this routine are
+memoized, i.e. they are stored in some package wide 'least-recently used' cache (from
```
lkdvos:
```suggestion
memoized, i.e. they are stored in some package-wide 'least-recently used' cache (from
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,435 @@
...
+tree manipulations (which we nonetheless try to avoid) will not seriously affect
+performance of tensor manipulations.
+
+### Inspecting fusion trees as tensors
+For those cases where the fusion and splitting tensors have an explicit representation as
+a tensor, i.e. a morphism in the category `Vect` (this essentially coincides with the case
+of group representations), this explicit representation can be created, which can be useful
+for checking purposes. Hereto, it is necessary that the *splitting tensor*
+``X^{ab}_{c,μ}``, i.e. the Clebsch-Gordan coefficients of the group, are encoded via the
+routine `fusiontensor(a,b,c [,μ = nothing])`, where the last argument is only necessary in
```
lkdvos:
```suggestion
routine `fusiontensor(a,b,c [,μ = 1])`, where the last argument is only necessary in
```

docs/src/man/sectors.md
```diff
@@ -104,11 +104,14 @@ The dimensions of the spaces ``R_a`` on which representation ``a`` acts are deno
...
+manipulations, which we summarize here:
 
 ![summary](img/tree-summary.svg)
 
-As always, we refer to the subsection on [topological data of a unitary fusion category](@ref ss_topologicalfusion) for further details.
+As always, we refer to the subsection on
+[topological data of a unitary fusion category](@ref ss_topologicalfusion) for further
+details.
 
 Finally, for the implementation, it will be useful to distinguish between an number of
```
lkdvos:
```suggestion
Finally, for the implementation, it will be useful to distinguish between a number of
```

docs/src/man/sectors.md
```diff
@@ -693,17 +699,13 @@ these types can be created in a type stable manner.
 As mentioned, the convenience mehtod `Vect[I]` will return the concrete type
```
lkdvos:
```suggestion
As mentioned, the convenience method `Vect[I]` will return the concrete type
```

docs/src/man/spaces.md
```diff
@@ -169,6 +169,12 @@ generally, the simple objects of a fusion category.
     complex data, a one-time warning will be printed, but most operations should continue
     to work nonetheless.
 
+One more important instance of `ElementarySpace` is the `GradedSpace`, which is used to
+represent a graded complex vector space with Euclidean product, where the grading is
```
lkdvos:
```suggestion
represent a graded complex vector space with Euclidean inner product, where the grading is
```

There is no such thing as a Euclidean product, only Euclidean inner product! :grin:

src/tensors/abstracttensor.jl
```diff
@@ -196,23 +186,36 @@ function adjointtensorindices(t::AbstractTensorMap, p::Index2Tuple)
...
 """
-blocks(t::AbstractTensorMap) = SectorDict(c => block(t, c) for c in blocksectors(t)) # TODO: make iterator
+    fusionblockstructure(t::AbstractTensorMap) -> TensorStructure
 
-@doc """
-    block(t::AbstractTensorMap, c::Sector) -> DenseMatrix
+Return the necessary structure information to decompose a tensor in blocks labeled by
+coupled sectors and in subblocks labeled by a splitting-fusion tree couple.
+"""
+fusionblockstructure(t::AbstractTensorMap) = fusionblockstructure(space(t))
```
lkdvos, line 210: Am I correct in assuming that this really is tied to `TensorMap`, and not necessarily to `AbstractTensorMap`? Would this still be correct for `AdjointTensorMap`?

Jutho: It depends on what you mean? The functions still works as it just acts on the underlying `HomSpace` instance, but of course the information it retrieves is only really useful `TensorMap` instances.

docs/src/man/fusiontrees.md
Jutho: Not sure how that word got in there, and what I had in mind instead. This is old docs that I did not (yet) change within this PR.

Jutho: Got it: this is supposed to be "context". 

We use context to determine how to print ...

(corresponding to the `context` keyword in Julia's print methods).

Jutho:
The basic benchmarks from the benchmark folder seem to convey that (contraction) performance should remain approximately the same. It will be worthwhile to combine this with Bumper.jl, as in that case there should be almost no remaining allocations (bumper allocates the data part, and the structure part is now cached), and to try some of the more challenging cases with several symmetries combined, which was one of the reasons behind the motivation for caching the structure info.




[mporesults.pdf](https://github.com/user-attachments/files/17498896/mporesults.pdf)
[peporesults.pdf](https://github.com/user-attachments/files/17498897/peporesults.pdf)

Jutho:
I think this is ready to be merged. I'll leave it open until the end of the weekend to await other comments. The updated docs can be viewed here:
https://jutho.github.io/TensorKit.jl/previews/PR140/

maartenvd:
I haven't investigated it further, but I see a significant (10x) performance (1ms to 8ms) decrease when switching from v0.12.7 to this branch. I also expected having to do even less matmuls than I still need to do on this branch.
```

using TensorKit, BenchmarkTools

include("submult.jl")

let
    L1 = Vect[(Irrep[U₁] ⊠ Irrep[SU₂] ⊠ FermionParity)]((-4, 0, 0)=>3, (-6, 0, 0)=>24, (-8, 0, 0)=>34, (-10, 0, 0)=>12, (-12, 0, 0)=>1, (-4, 1, 0)=>2, (-6, 1, 0)=>29, (-8, 1, 0)=>42, (-10, 1, 0)=>13, (-6, 2, 0)=>2, (-8, 2, 0)=>4, (-5, 1/2, 1)=>18, (-7, 1/2, 1)=>53, (-9, 1/2, 1)=>34, (-11, 1/2, 1)=>3, (-5, 3/2, 1)=>4, (-7, 3/2, 1)=>18, (-9, 3/2, 1)=>8)
    L2 = Vect[(Irrep[U₁] ⊠ Irrep[SU₂] ⊠ FermionParity)]((0, 0, 0)=>1, (2, 0, 0)=>1, (1, 1/2, 1)=>1)
    R1 = Vect[(Irrep[U₁] ⊠ Irrep[SU₂] ⊠ FermionParity)]((-4, 0, 0)=>3, (-6, 0, 0)=>24, (-8, 0, 0)=>34, (-10, 0, 0)=>12, (-12, 0, 0)=>1, (-4, 1, 0)=>2, (-6, 1, 0)=>29, (-8, 1, 0)=>42, (-10, 1, 0)=>13, (-6, 2, 0)=>2, (-8, 2, 0)=>4, (-5, 1/2, 1)=>18, (-7, 1/2, 1)=>53, (-9, 1/2, 1)=>34, (-11, 1/2, 1)=>3, (-5, 3/2, 1)=>4, (-7, 3/2, 1)=>18, (-9, 3/2, 1)=>8)
    R2 = Vect[(Irrep[U₁] ⊠ Irrep[SU₂] ⊠ FermionParity)]((0, 0, 0)=>1, (2, 0, 0)=>1, (1, 1/2, 1)=>1)
    R3 = Vect[(Irrep[U₁] ⊠ Irrep[SU₂] ⊠ FermionParity)]((-1, 1/2, 1)=>1)

    R4 = Vect[(Irrep[U₁] ⊠ Irrep[SU₂] ⊠ FermionParity)]((-4, 0, 0)=>3, (-6, 0, 0)=>25, (-8, 0, 0)=>34, (-10, 0, 0)=>12, (-12, 0, 0)=>1, (-4, 1, 0)=>2, (-6, 1, 0)=>31, (-8, 1, 0)=>44, (-10, 1, 0)=>13, (-6, 2, 0)=>2, (-8, 2, 0)=>5, (-5, 1/2, 1)=>18, (-7, 1/2, 1)=>50, (-9, 1/2, 1)=>30, (-11, 1/2, 1)=>2, (-5, 3/2, 1)=>4, (-7, 3/2, 1)=>20, (-9, 3/2, 1)=>6) 
    R5 = Vect[(Irrep[U₁] ⊠ Irrep[SU₂] ⊠ FermionParity)]((0, 0, 0)=>1, (2, 0, 0)=>1, (1, 1/2, 1)=>1)'


    T1 = TensorMap{Float64}(undef,(L1 ⊗ L2) ← (R1 ⊗ R2 ⊗ R3))*0 # I'd like to init with rand, but that is not yet implemented?
    T2 = TensorMap{Float64}(undef,(R1 ⊗ R2) ← (R4 ⊗ R5))*0;
    T3 = TensorMap{Float64}(undef,(L1 ⊗ L2) ← (R4 ⊗ R5 ⊗ R3))*0;

    #T1 = TensorMap(rand,Float64,(L1 ⊗ L2) ← (R1 ⊗ R2 ⊗ R3))
    #T2 = TensorMap(rand,Float64,(R1 ⊗ R2) ← (R4 ⊗ R5));
    #T3 = TensorMap(rand,Float64,(L1 ⊗ L2) ← (R4 ⊗ R5 ⊗ R3));

    lsm = LeftSubMult(space(T1),space(T2))

    @benchmark begin
        rmul!($T3,false)
        $lsm($T3,$T1,$T2)
    end

end

```
The code that I have is convoluted, but the thing I'm timing is extremely straightforward, and can be found at the bottom of [submult.txt](https://github.com/user-attachments/files/17529835/submult.txt) (attached as txt - github refuses .jl)

lkdvos:
I might be missing something, but I quickly looked at your file and it seems at the bottom, you are using sectors to index into the `data` field of the tensor, which should now be a dense vector, so I don't think that can work?

maartenvd:
I completely forgot that the underlying data field also changed! The speedup would come from the re-ordered fusiontrees, but I'll need to change the code a little bit

Jutho:
I'll also take a look at this asap.Serious performance loss should indeed not happen and would have to be resolved before merging this.

Regarding the random initialization; the new syntax is:
```julia
rand(Float64,(L1 ⊗ L2) ← (R1 ⊗ R2 ⊗ R3))
```
That was actually in another PR from @lkdvos , but I did implement deprecation warnings for that change in this PR, so in principle the old approach should still work? I see what is wrong; the old syntax was `TensorMap(rand, Float64, ...)`. This is still supported but in deprecation mode. I don't think `TensorMap{Float64}(rand, ...)` was ever supported?

maartenvd:
I was testing an old version of jh/tensorstructure (with some extra local
changes). I will reimplement on the latest master today, it really should
be strictly faster (less matmuls and less dictionary indexing)

lkdvos:
With latest master you mean latest version of this branch right?

maartenvd:
yep - will also check if the blockstructure caching gives a sufficient
enough speedup to do away with some of the uglier things in
mpskitexperimental/qchem, but that will take more time. In general I'm
always a bit afraid of the possible performance hit of blocking caches

src/spaces/homspace.jl
```diff
@@ -134,3 +140,119 @@ function compose(W::HomSpace{S}, V::HomSpace{S}) where {S}
...
+
+        offset₁ = 0
+        for f₁ in fusiontrees(codom, c)
+            push!(splittingtrees, f₁)
+            d₁ = dim(codom, f₁.uncoupled)
+            push!(splittingstructure, (offset₁, d₁))
+            offset₁ += d₁
+        end
+        blockdim₁ = offset₁
+        strides = (1, blockdim₁)
```
maartenvd, line 201: Doesn't this mean that [i,j] is next to [i+1,j] in memory? Why is tensor tensor multiplication then still fast?

src/spaces/homspace.jl
```diff
@@ -134,3 +140,119 @@ function compose(W::HomSpace{S}, V::HomSpace{S}) where {S}
...
+        blockstructure[c] = (blocksize, blockrange)
+    end
+
+    fusiontreeindices = sizehint!(FusionTreeDict{Tuple{F₁,F₂},Int}(),
+                                  length(fusiontreelist))
+    for (i, f₁₂) in enumerate(fusiontreelist)
+        fusiontreeindices[f₁₂] = i
+    end
+    totaldim = blockoffset
+    structure = FusionBlockStructure{I,F₁,F₂}(totaldim, blockstructure,
```
maartenvd: This structure makes it quite difficult to figure out what stridedview corresponds with [:,f2] (for a given f2.coupled). So what used to be the old rowr/colr takes a bit of compute to get out again. Not really a problem on it's own, but I'm surprised that that operation is never needed (for example if I want to permute two spaces in the codomain, do we then iterate over all possible (f1,f2) pairs?)

src/spaces/homspace.jl
Jutho: Not sure I understand the question? The structure is not so much different than the old one, except that there is a different order among the fusion trees that couple to a certain blocksector `c` (as requested by you), and that the old matrix blocks are just stacked after each other in one long vector.

src/spaces/homspace.jl
Jutho: The `permute` function does indeed not specialise for permutations that do not cross the domain or the codomain. If I remember correctly, the idea, as dicussed with @lkdvos, was that all `(f1,f2)` pairs are now easily associated with a linear index, such that the current caches for permutations (and transpositions) can take the form of a sparse matrix, which would then also make it easier to compute their inverse etc.

I do see what you mean; that doesn't really come up in any of the implementations so far (it's always acting on individual (f1,f2) pairs), but it is probably needed for what you are trying to do. I am happy to further revise the structure.

Jutho:
> yep - will also check if the blockstructure caching gives a sufficient enough speedup to do away with some of the uglier things in mpskitexperimental/qchem, but that will take more time. In general I'm always a bit afraid of the possible performance hit of blocking caches
> […](#)

I am also suspicious of the blocking cache, so some benchmark in an actual test case that involves several tasks/threads would be great. That's why there is also a task local cache implementation, though there is not yet functionality for readily selecting that, other than replacing the existing method by doing
```
TensorKit.function CacheStyle(I::Type{<:Sector})
    return TaskLocalCache{Dict{Any,Any}}()
end
```

Yet another alternative would be a global cache based on https://github.com/JuliaConcurrent/MultiThreadedCaches.jl

src/spaces/homspace.jl
maartenvd: I got confused by strides. I thought that the strides for a matrix (d1,d2) were (d2,1). Apparantly they're (1,d1), which is exactly what you're doing here...

src/spaces/homspace.jl
lkdvos: I guess the claim is that you never really need the `rowr` and `colr` anymore anyways, since it does not really make sense to first try and obtain a view of the block before obtaining a view of the subblock. Therefore, there is no real difference between permutations within a block (keeping coupled charges fixed) and permutations that do not have this, since these all index into one contiguous vector anyways. Before, you would have to do something like:
1. Get block for `c` (dictionary lookup)
2. Index into that block with `rowr[c][f1]` and `colr[c][f2]`

While now, it should really just be:
1. Index into the data using the structure

The hope being that we reduce dictionary lookups etc.

Additionally, the structure is sorted/stored in a vector, so we should be able to avoid additional dict lookups within `add_transform!` by replacing the `fusiontreemap` with a simple sparse matrix, using integers instead of hashables (this is not yet implemented though). Of course, we still want the option to have "random access" patterns, so there is also still a `fusiontreeindices` dictionary that maps fusiontrees to the correct index. In this way, I would hope to be able to at least optimize `add_transform!` to do a single cache-lookup for the structure, and then have just vector indexing for extracting the correct data. I might try my hand at that implementation if I find some time tonight.

maartenvd:
My updated submult is even more of a mess but 
- it now indeed only does 27 mul! calls, instead of 281
- it is only a little bit slower (1ms -> 1.1ms)

The slowdown might be due to the bad alignment or something more subtle. It is not significant enough to care, but quite surprising nevertheless.

[submult.txt](https://github.com/user-attachments/files/17531267/submult.txt)

src/spaces/homspace.jl
maartenvd: You can often gain a lot of speed if you can vectorize a group of similar operations. Consider that you are manipulating a part of the domain where all blocks [f1,f2] get mapped to [f1,f2'] (so not a function of f1), then you might be able to simultaneously vectorize a group of operations [:,f2] -> [:,f2'], which should be significantly faster.

To give a concrete - but perhaps not relevant - example, let's say you have a tensormap and you want to project out a single fusiontree in the codomain. With the current datastructure, you'd have to go over every (f1,f2) pair and zero out if f1 matches. In the old tensorkit, you could have written something like data[f1.coupled][rowr[f1],:].=0, which should be significantly faster.

All that aside - having indexmanipulations implemented as sparse matrices is going to be amazing!

src/spaces/homspace.jl
```diff
@@ -134,3 +140,119 @@ function compose(W::HomSpace{S}, V::HomSpace{S}) where {S}
...
+    N₁ = length(codom)
+    N₂ = length(dom)
+    I = sectortype(W)
+    F₁ = fusiontreetype(I, N₁)
+    F₂ = fusiontreetype(I, N₂)
+
+    # output structure
+    blockstructure = SectorDict{I,Tuple{Tuple{Int,Int},UnitRange{Int}}}() # size, range
+    fusiontreelist = Vector{Tuple{F₁,F₂}}()
+    fusiontreestructure = Vector{Tuple{Tuple{Int,Int},Tuple{Int,Int},Int}}() # size, strides, offset
```
lkdvos: I might be missing something here, but I find it a bit strange to store the `fusiontreestructure` sizes and strides as if it is a matrix. This means that within the `getindex(t, f1, f2)`, we first have to explicitly construct the `StridedView` as a matrix, and then reshape it into a `N1+N2` dimensional array. Can we not just immediately cache the size and strides of that resulting array?

Jutho: Probably, I know I tried as you suggested this before, but I found it conceptually much harder to compute and store that information in one go. I don't recall if there was a fundamental obstacle; I guess not. It is just that that information is much more entangled, as the strides for the "column" part depend on the "row" information. Datastructure wise, it then requires tuples of length `N1+N2`, whereas now only `F₁` and `F₂` are type parameters. I am wondering if there exists any case where the overhead of doing that final reshape of the `StridedView` matrix would really be significant, and would e.g. take more time than the global cache lookup itself.

lkdvos: I bumped into this yesterday as I was trying to rewrite the `add_transform!` with a SparseMatrix. The thing is, if I need to have an additional reshape in obtaining the subblocks, it becomes a bit more involved as I then also need to simultaneously loop over the fusiontrees to have the uncoupled charges. I agree that the cached structure is absolutely conceptually hard to follow, but I would argue that "the strides for a matrix form of a block that still has to be reshaped" isn't that much better.

My main argument to changing this is however the following preliminary benchmark results I got with the SparseMatrix implementation:
```julia
V = SU2Space(i => 4 for i in 0:1//2:3)
t1 = rand(V^3);
p = ((3, 2, 1), ())
t2 = TensorKit.permute(t1, p);
```
```julia
julia> @benchmark add_permute!($t2, $t1, $p, $true, $false, $(NewBackend(StridedBLAS())))
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  34.656 μs … 66.047 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     35.047 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   35.910 μs ±  2.070 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅█▆▂  ▅▇▄▂   ▁▁▁   ▁▁     ▃▂     ▁▂                         ▂
  █████▆███████████▅▇███▇▇▇▇██▇▆▅▆▆███▆▅▄▄▅▇▇▆▄▃▄▃▃▆█▅▅▄▃▁▃▃▆ █
  34.7 μs      Histogram: log(frequency) by time      44.8 μs <

 Memory estimate: 272 bytes, allocs estimate: 6.

julia> @benchmark add_permute!($t2, $t1, $p, $true, $false, $(StridedBLAS()))
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  204.842 μs …  19.483 ms  ┊ GC (min … max): 0.00% … 98.48%
 Time  (median):     216.778 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   234.075 μs ± 274.092 μs  ┊ GC (mean ± σ):  2.98% ±  2.99%

    █▇▅▃▃▃▄▆▆▃▁▁                                                ▁
  █████████████████▇▆▆▆▆▆▅▅▆▅▅▄▄▄▅▅▄▅▄▂▂▂▃▃▄▄▄▄▃▄▄▄▅▄▂▃▅▆▅▆▅▆▅▆ █
  205 μs        Histogram: log(frequency) by time        393 μs <

 Memory estimate: 81.59 KiB, allocs estimate: 1385.
```

I'll try to get it cleaned up a bit into a new branch, maybe you could run the entire benchmark suite again after?

Jutho: Wow, is that because of better multithreading? Or even in the single threaded case? Where is that speedup coming from?

So it seems that for some optimisations, having immediately the final StridedView field data for a given `(f1,f2)` is convenient, whereas for other optimisations (@maartenvd) the ideal data is the orignal `rowr` data for a given `f1` and separately `colr` for a given `f2` would be convenient? Whereas the current structure is a mix that jointly stores `rowr` and `colr` data for given `(f1,f2)`, which is optimal for neither case? :smile:

I will look into changing this tonight, and possibly having all of this available? We anyway need to first construct the old data before we can compute the `strides` data, so we can just as well cache everything.

lkdvos: It's just single threaded, but with a whole lot less dictionary lookups

maartenvd: That is amazing; I had such a large overhead for add_permute that I precalculated the exact offsets in the data, and circumvented most dict lookups! This version might be faster than what I spaghetti'd together.

lkdvos: #167 is what I am working on. It could do with a large benchmark, but at the very least I feel it is quite elegant, and from my preliminary testing it is a lot faster. (Keep in mind that I disabled multithreading for now, which can also skew the results for smaller tensors)

My best guess is that really, having the offsets and everything precomputed in the caches, in vectors instead of dictionaries is just much less overhead, and additionally I think I managed to reduce the global cache lookups to 1 per transformation, instead of having to fetch that cache every single time a block needs to be extracted.

maartenvd: If you want an actual tensor that I encountered during dmrg, you could look at the first contraction I did here https://github.com/maartenvd/MPSKitExperimental.jl/blob/main/examples/submult.ipynb The fastloop code will be broken, but I would be interested in knowing just how much faster you can get that particular contraction now

PR #141, opened 2024-08-03
New indexing and iteration for ProductSector

Jutho:
ProductSector can be indexed and will iterate using an order that is based on running through slices of constant Manhattan distance. This enables indexing and meaningful iteration if infinite sectors are included in the product sector.

### DIFF
--- a/TensorKitSectors/src/auxiliary.jl
+++ b/TensorKitSectors/src/auxiliary.jl
@@ -13,3 +13,77 @@ function _kron(A, B)
     end
     return C
 end
+
+# Manhattan based distance enumeration: t is supposed to be one-based index
+# TODO: is there any way to make this faster?
+@inline function num_manhattan_points(d::Int, sz::Dims{1})
+    return Int(sz[1] > d)
+end
+@inline function num_manhattan_points(d::Int, sz::Dims{N}) where {N}
+    d == 0 && return 1
+    num = 0
+    for i in 1:min(sz[1], d + 1)
+        num += num_manhattan_points(d - i + 1, Base.tail(sz))
+    end
+    return num
+end
+
+@inline localoffset(d, I::Tuple{Int}, sz::Tuple{Int}) = 0
+@inline function localoffset(d, I, sz)
+    offset = 0
+    for i in 1:(I[1] - 1)
+        offset += num_manhattan_points(d - i + 1, Base.tail(sz))
+    end
+    offset += localoffset(d - I[1] + 1, Base.tail(I), Base.tail(sz))
+    return offset
+end
+
+to_manhattan_index(::Tuple{}, ::Tuple{}) = 1
+to_manhattan_index(I::Tuple{Int}, ::Tuple{Int}) = I[1]
+function to_manhattan_index(I::Dims{N}, sz::Dims{N}) where {N}
+    d = sum(I) - N # Manhattan distance to origin for one-based indices
+    d == 0 && return 1
+    index = 1
+    # count all the points with smaller Manhatten distance
+    for k in 0:(d - 1)
+        index += num_manhattan_points(k, sz)
+    end
+    return index + localoffset(d, I, sz)
+end
+
+
+
+@inline invertlocaloffset(d, offset, sz::Tuple{Int}) = (d+1,)
+@inline function invertlocaloffset(d, offset, sz)
+    i₁ = 1
+    while i₁ < sz[1]
+        jump = num_manhattan_points(d - i₁ + 1, Base.tail(sz))
+        if offset < jump
+            break
+        end
+        offset -= jump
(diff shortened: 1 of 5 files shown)


TensorKitSectors/src/product.jl
```diff
@@ -166,8 +166,8 @@ dim(p::ProductSector) = *(dim.(p.sectors)...)
 
 Base.isequal(p1::ProductSector, p2::ProductSector) = isequal(p1.sectors, p2.sectors)
 Base.hash(p::ProductSector, h::UInt) = hash(p.sectors, h)
-function Base.isless(p1::ProductSector{T}, p2::ProductSector{T}) where {T}
-    return isless(reverse(p1.sectors), reverse(p2.sectors))
+function Base.isless(p1::P, p2::P) where {P<:ProductSector}
+    return isless(findindex(values(P), p1), findindex(values(P), p2))
```
lkdvos, line 170: I think here it is possible to try to avoid computing the manhattan index if we really want:
- first compute the manhattan distance and compare
- if equal, only the `localoffset` should be compared

I would even guess that this second point can be determined without even computing the offset, by comparing the cartesianindices?

TensorKitSectors/src/auxiliary.jl
```diff
@@ -13,3 +13,78 @@ function _kron(A, B)
     end
     return C
 end
+
+# Manhattan based distance enumeration: I is supposed to be one-based index
+# TODO: is there any way to make this faster?
```
lkdvos, line 18: There's definitely some number theory stuff that could come in handy here, if I am not mistaken, the relevant terms are triangular numbers, tetrahedral numbers, and their generalizations.
There are general formulas for this:
```math
 \sum_{n_{k-1} = 1}^{n_k} \sum_{n_{k-2} = 1}^{n_{k-1}} \cdots \sum_{n_1 = 1}^{n_2} n_1 = \begin{pmatrix} n_k + k - 1 \\ k \end{pmatrix}
```

This is however for infinite grids, but I feel like it should be possible to split the grid into regions that are either hyperrectangular, or "hypertriangular", both of which have "volume formulas"?

https://en.wikipedia.org/wiki/Tetrahedral_number

TensorKitSectors/src/product.jl
```diff
@@ -29,25 +30,24 @@ function Base.size(::SectorValues{ProductSector{T}}) where {T<:SectorTuple}
...
-        throw(ArgumentError("cannot index into infinite product sector"))
-    return LinearIndices(size(P))[CartesianIndex(findindex.(values.(_sectors(T)),
-                                                            c.sectors))]
+                   c::ProductSector{T}) where {T<:SectorTuple}
+    return to_manhattan_index(findindex.(values.(_sectors(T)), c.sectors), _size(P))
+end
+
+function Base.iterate(P::SectorValues{ProductSector{T}}, i=1) where {T<:SectorTuple}
+    Base.IteratorSize(P) != Base.IsInfinite() && i > length(P) && return nothing
+    return getindex(P, i), i + 1
```
lkdvos, line 50: I guess here is the biggest possible speed trap, it feels like it could be possible to iterate through these without having to map to linear indices, similar to how cartesianindices can be iterated without the integer divisions.

lkdvos (approved):
I just added some of my thoughts as a reply to the "can we make this faster" comment, but I am not sure if this really matters, and presumably that really is over-engineering things.
At the very least, I would say to merge as-is, and if necessary optimize later

TensorKitSectors/src/auxiliary.jl
Jutho: The binomial formula is what I started with, but that indeed only works if there are no upper bounds, i.e. if all the sectors involved are infinite.

TensorKitSectors/src/product.jl
Jutho: Given that iteration over all possible sector values is not really used in the code, except for making tests, I don't care too much about the speed of this :-)

Jutho:
Ok, I think I will merge as is. I have ran the tests locally, but currently the TensorKitSectors tests are not running in CI. While analysing CI, also noticed something weird. The MacOS tests are all slightly quicker then on Ubuntu (for the tensor tests, they do not have all the sector choices in common, but those that do were always a bit faster on Mac), up to the point of AD. The AD tests seem anyway to take very long compared to the rest of the tests on both platforms, but on Mac they become really excessive. The AD tests for `Rep[U1]` take 14 minutes on Ubuntu and 87 minutes (!) on Mac. Not sure what is going on there; maybe excessive memory leading to caching or something?

lkdvos:
It could also be the RNG, I think the AD tests are really fully random (including spaces/arrows etc) now, making it quite hard to control the weight of the tests...

PR #146, opened 2024-08-23
Add `sectorscalartype`

lkdvos:
This is a minor change that centralizes the determination of the `eltype` of the topological data.
This simplifies some of the fusion tree manipulations, as it makes it easier to determine the output type.

Additionally, this also provides a hook for types that don't necessarily want to have `Fsymbol(one(I), one(I),....)`, either because it is not efficient, or because `one(I)` is hard to determine (multi-fusion stuff).

Finally, it could also provide a starting point for having sectors with different eltypes, as one could imagine wanting SU2 symbols with BigFloat precision etc.

I tried to not change too much, but left a couple comments in the code about things I am unsure about. We should probably also discuss the fusion tree manipulations constructions, as the `@isdefined` construction might not be necessary now. I don't know if this has performance implications (either on the compiler or at runtime), but it might just be slightly more readible as well.

### DIFF
--- a/TensorKitSectors/src/irreps.jl
+++ b/TensorKitSectors/src/irreps.jl
@@ -197,27 +197,26 @@ findindex(::SectorValues{SU2Irrep}, s::SU2Irrep) = twice(s.j) + 1
 dim(s::SU2Irrep) = twice(s.j) + 1
 
 FusionStyle(::Type{SU2Irrep}) = SimpleFusion()
+sectorscalartype(::Type{SU2Irrep}) = Float64
 Base.isreal(::Type{SU2Irrep}) = true
 
 Nsymbol(sa::SU2Irrep, sb::SU2Irrep, sc::SU2Irrep) = WignerSymbols.δ(sa.j, sb.j, sc.j)
 
-Fscalartype(::Type{SU2Irrep}) = Float64
 function Fsymbol(s1::SU2Irrep, s2::SU2Irrep, s3::SU2Irrep,
                  s4::SU2Irrep, s5::SU2Irrep, s6::SU2Irrep)
     if all(==(_su2one), (s1, s2, s3, s4, s5, s6))
         return 1.0
     else
         return sqrtdim(s5) * sqrtdim(s6) *
-               WignerSymbols.racahW(Fscalartype(SU2Irrep), s1.j, s2.j,
+               WignerSymbols.racahW(sectorscalartype(SU2Irrep), s1.j, s2.j,
                                     s4.j, s3.j, s5.j, s6.j)
     end
 end
 
-Rscalartype(::Type{SU2Irrep}) = Fscalartype(SU2Irrep)
 function Rsymbol(sa::SU2Irrep, sb::SU2Irrep, sc::SU2Irrep)
-    Nsymbol(sa, sb, sc) || return zero(Rscalartype(SU2Irrep))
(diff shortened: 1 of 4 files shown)


TensorKitSectors/src/irreps.jl
```diff
@@ -348,52 +353,55 @@ function Nsymbol(a::CU1Irrep, b::CU1Irrep, c::CU1Irrep)
...
     end
     # from here on, a, b, c are neither 0+ or 0-
-    s = sqrt(2) / 2
+    s = T(sqrt(2) / 2)
     if a == b == c
         if d == a
             if e.j == 0
                 if f.j == 0
-                    return f.s == 1 ? -0.5 : 0.5
+                    return f.s == 1 ? T(-0.5) : T(0.5)
```
Jutho: `T(1//2)` ? not that it matters

src/fusiontrees/manipulations.jl
```diff
@@ -925,7 +921,7 @@ function braid(f::FusionTree{I,N},
                p::NTuple{N,Int}) where {I<:Sector,N}
     TupleTools.isperm(p) || throw(ArgumentError("not a valid permutation: $p"))
     if FusionStyle(I) isa UniqueFusion && BraidingStyle(I) isa SymmetricBraiding
-        coeff = Rsymbol(one(I), one(I), one(I))
+        coeff = Rsymbol(one(I), one(I), one(I)) # one(Reltype(I)) ?
```
Jutho: Yes I think this can become `one(Reltype(I))`

src/fusiontrees/manipulations.jl
```diff
@@ -1020,11 +1013,9 @@ function braid(f₁::FusionTree{I}, f₂::FusionTree{I},
         end
     else
         if usebraidcache_nonabelian[]
-            u = one(I)
             T = BraidingStyle(I) isa NoBraiding ?
-                typeof(Fsymbol(u, u, u, u, u, u)[1, 1, 1, 1]) :
-                typeof(sqrtdim(u) * Fsymbol(u, u, u, u, u, u)[1, 1, 1, 1] *
-                       Rsymbol(u, u, u)[1, 1])
+                Feltype(I) :
+                typeof(one(Feltype(I)) * one(Reltype(I)) * sqrtdim(one(I))) # do we need sqrtdim here ?
```
Jutho: We probably don't need `sqrtdim`. For nonabelian sectors, the `Feltype` should already be a floating point type probably.

lkdvos:
As discussed, I now replaced everything with just a unified `sectorscalartype`. I started going through removing the `@isdefined` calls, but actually it is not that elegant, as the `SingletonDict`s don't really support pre-initializing the dictionary. I think it really is better to just leave them for now.

If tests turn green, this is good to go for me.

Jutho:
I am wondering to what extent it would be an abomination to replace `sectorscalartype(somesectortype)` with simply `scalartype(somesectortype)`? It is not really a correct application of method overloading, as it is definitely changing the way/meaning of `scalartype` from its original intention. But `sectorscalartype(somesectortype)` sounds like overkill.

lkdvos:
I have to admit that this also crossed my mind, however I was a bit worried about something like a vector of sectors which suddenly has scalartype Float64, and then the default definition of zerovector might return a vector of floats (although I did not test if this is actually the case).
In the end, it doesn't matter too much, I don't think many external users would use this, so maybe we shouldn't even export it

Jutho:
Yes you are right that this will cause unforeseen problems further on. By lack of a better suggestion I will merge as is.

PR #148, opened 2024-08-26
AdjointTensorMap can wrap generic tensor types

lkdvos:
This changes the `AdjointTensorMap` type such that it can wrap any `AbstractTensorMap` instead of forcing it to be a `TensorMap` instance.


Jutho:
Looks good; this is in preparation of other tensor types?

lkdvos:
Indeed, it should work for `BlockTensorMap`, and it's probably a more versatile approach that could also include possible `DiagonalTensorMap` in the future

PR #151, opened 2024-09-09
Excise TensorKitSectors

lkdvos:
This uses the newly created TensorKitSectors package to make packages that want to implement new sectors have a lighter dependency, and to lighten the workload of testing in the main package. Additionally, other (symmetric) tensor software might benefit from the topological data contained in that package without having to depend on the full TensorKit stack.


Jutho:
Look's good; thanks!

PR #159, opened 2024-10-17
Remove fishy loading of extension in extension

lkdvos:
this avoids loading TensorOperationsChainRulesCoreExt from within TensorKitChainRulesCoreExt, which seems to lead to issues from julia 1.11.1 onwards

Will backport this to the 0.12 branch after it gets merged.


Jutho:
So this is not going to work ... 😄 Should I cancel the CI runs?

Issue #158, opened 2024-10-17
LoadError: type Nothing has no field _conj


Yue-Zhengyuan:
On Julia 1.11.1, I get the following error when building TensorKit: 

```
(@v1.11) pkg> precompile
Precompiling project...
  ✗ TensorKit → TensorKitChainRulesCoreExt
  0 dependencies successfully precompiled in 1 seconds. 466 already precompiled.

ERROR: The following 1 direct dependency failed to precompile:

TensorKitChainRulesCoreExt 

Failed to precompile TensorKitChainRulesCoreExt [74f4665c-2d76-51f1-bd11-a86ebbb150d0] to "/Users/zhengyuanyue/.julia/compiled/v1.11/TensorKitChainRulesCoreExt/jl_uJjNv9".
ERROR: LoadError: type Nothing has no field _conj
Stacktrace:
 [1] getproperty(x::Nothing, f::Symbol)
   @ Base ./Base.jl:49
 [2] top-level scope
   @ ~/.julia/packages/TensorKit/59D34/ext/TensorKitChainRulesCoreExt/TensorKitChainRulesCoreExt.jl:19
 [3] include
   @ ./Base.jl:557 [inlined]
 [4] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2790
 [5] top-level scope
   @ stdin:5
in expression starting at /Users/zhengyuanyue/.julia/packages/TensorKit/59D34/ext/TensorKitChainRulesCoreExt/TensorKitChainRulesCoreExt.jl:1
in expression starting at stdin:
```

Package versions:
```
[07d1fe3e] TensorKit v0.12.6
⌅ [6aa20fa7] TensorOperations v4.1.1
```

lkdvos:
Aha, I see that my "extension hacking" has finally come to bite me in the backside...

[...]

PR #160, opened 2024-10-17
Remove fishy imports from other extension

lkdvos:
This removes the fishy imports of `_conj` and `trivtuple` from the `TensorOperationsChainRulesCoreExt` within the
`TensorKitChainRulesCoreExt`, which is no longer supported in julia 1.11.1


lkdvos:
The formatcheck is a result of JuliaFormatter v2, I still need to investigate what is changing there

lkdvos:
This should resolve #158 in the 0.12.x versions.

PR #161, opened 2024-10-17
`scalar` and `AdjointTensorMap` AD rrules

### LINKED ISSUE #157: Autodiff doesn't work for * for AdjointTensorMap
Here is the example:
```julia
using TensorKit, Zygote
g(x)=real(scalar(x'*x))
V=Z3Space(0=>2,1=>2,2=>2);
A=randn(ComplexF64,V*V*V',one(V));
gradient(g,A)
```

Output:
```
ERROR: MethodError: no method matching *(::ChainRulesCore.Tangent{…}, ::TensorKit.AdjointTensorMap{…})
The function `*` exists, but no method is defined for this combination of argument types.
```

lkdvos:
This should resolve #157 .

!!! Needs to be ported to the new version still


Jutho:
We should probably add some tests for all of this, but I will merge for now.

PR #167, opened 2024-10-27
Treetransformer optimizations

lkdvos:
This PR refactors some of the structure code to obtain a (hopefully) faster kernel for the `add_transform!` implementation.
The main changes are that the cached structures now immediately return the correct strides and sizes, and the kernels make use of that feature, and avoid all dictionary lookups.
I also made sure to loop over the destination subblocks first, such that this should be more convenient to allow for multithreaded approaches.

We might want to consider merging that other PR first, and just merging this straight to master...

## TODO
- [x] abelian specialization
- [ ] adjoint specialization
- [ ] multithreaded experiments
- [ ] benchmarks

### DIFF
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -81,7 +81,8 @@ end
         d = (dims(V2 ⊗ V1, f₁.uncoupled)..., dims(V1 ⊗ V2, f₂.uncoupled)...)
         n1 = d[1] * d[2]
         n2 = d[3] * d[4]
-        data = fill!(storagetype(b)(undef, (n1, n2)), zero(scalartype(b)))
+        data = storagetype(b)(undef, (n1, n2))
+        fill!(data, zero(eltype(b)))
         a1, a2 = f₂.uncoupled
         if f₁.uncoupled == (a2, a1)
             braiddict = artin_braid(f₂, 1; inv=b.adjoint)
@@ -138,20 +139,22 @@ function block(b::BraidingTensor, s::Sector)
 
     length(data) == 0 && return data # s ∉ blocksectors(b)
 
-    data = fill!(data, zero(scalartype(b)))
+    data = fill!(data, zero(eltype(b)))
 
     V1, V2 = domain(b)
     if sectortype(b) === Trivial
         d1, d2 = dim(V1), dim(V2)
         subblock = sreshape(StridedView(data), (d1, d2, d2, d1))
         @inbounds for i in axes(subblock, 1), j in axes(subblock, 2)
-            subblock[i, j, j, i] = one(scalartype(b))
+            subblock[i, j, j, i] = one(eltype(b))
         end
         return data
     end
 
     structure = fusionblockstructure(b)
-    for ((f1, f2), (sz, str, _)) in
+    base_offset = first(structure.blockstructure[s][2]) - 1
+
+    for ((f1, f2), (sz, str, off)) in
         zip(structure.fusiontreelist, structure.fusiontreestructure)
         if (f1.uncoupled != reverse(f2.uncoupled)) || !(f1.coupled == f2.coupled == s)
             continue
@@ -161,8 +164,8 @@ function block(b::BraidingTensor, s::Sector)
         haskey(braiddict, f1) || continue
         r = braiddict[f1]
 
-        # discard offset because single block
-        subblock = StridedView(data, sz, str)
+        # change offset to account for single block
+        subblock = StridedView(data, sz, str, off - base_offset)
         @inbounds for i in axes(subblock, 1), j in axes(subblock, 2)
             subblock[i, j, j, i] = r
         end
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -117,7 +117,7 @@ function Base.copy!(t::TensorMap, b::BraidingTensor)
             braiddict = artin_braid(f₂, 1; inv=b.adjoint)
             r = convert(scalartype(t), get(braiddict, f₁, zero(valtype(braiddict))))
         end
-        for i in 1:size(data, 1), j in 1:size(data, 2)
+        @inbounds for i in axes(data, 1), j in axes(data, 2)
             data[i, j, j, i] = r
         end
     end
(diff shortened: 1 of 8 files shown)


src/tensors/braidingtensor.jl
```diff
@@ -126,39 +126,48 @@ end
...
-    if sectortype(b) == Trivial
+
+    # TODO: probably always square?
+    m = blockdim(codomain(b), s)
+    n = blockdim(domain(b), s)
+    data = Matrix{eltype(b)}(undef, (m, n))
+
+    length(data) == 0 && return data # s ∉ blocksectors(b)
+
+    data = fill!(data, zero(scalartype(b)))
```
Jutho: It is a bit odd looking that here you use `scalartype`, whereas to create `data` you used `eltype`.

src/tensors/braidingtensor.jl
```diff
@@ -126,39 +126,48 @@ end
...
+        if (f1.uncoupled != reverse(f2.uncoupled)) || !(f1.coupled == f2.coupled == s)
+            continue
+        end
+
+        braiddict = artin_braid(f2, 1; inv=b.adjoint)
+        haskey(braiddict, f1) || continue
+        r = braiddict[f1]
+
+        # discard offset because single block
+        subblock = StridedView(data, sz, str)
```
Jutho: I am confused by this. There can be different fusion trees that appear in the same block right? And the local offset within the block is important for that.

src/tensors/treetransformers.jl
```diff
@@ -0,0 +1,157 @@
...
+
+    structure_dst = fusionblockstructure(Vdst)
+    structure_src = fusionblockstructure(Vsrc)
+
+    rows = Int[]
+    cols = Int[]
+    vals = sectorscalartype(sectortype(Vdst))[]
+
+    for (f1, f2) in structure_src.fusiontreelist
+        row = structure_src.fusiontreeindices[(f1, f2)]
```
Jutho: Is this not simply `(row, (f1, f2)) in enumerate(structure_src.fusiontreelist)`?

src/tensors/treetransformers.jl
```diff
@@ -0,0 +1,157 @@
...
+@noinline function _get_treepermuter(A, key)
+    d::A = get!(treepermutercache, key) do
+        return _treepermuter(key)
+    end
+    return d
+end
+function _treepermuter((Vdst, Vsrc, p))
+    fusiontreetransform(f1, f2) = permute(f1, f2, p...)
+    return TreeTransformer(fusiontreetransform, Vsrc, Vdst)
+end
```
Jutho, line 107: Maybe all of this could become:

```julia
for (transform, transformer) in
    ((:permute, :permuter), (:braid, :braider), (:transpose, :transposer))

    treetransformcache = Symbol("tree", transformer, "cache")
    usetreetransformcache = Symbol("usetree", transformer, "cache")
    treetransformer = Symbol("tree", transformer)
    _get_treetransformer = Symbol("_get_", treetransformer)
    _treetransformer = Symbol("_", treetransformer)

    @eval begin
        const $treetransformcache = LRU{Any,Any}(; maxsize=10^5)
        const $usetreetransformcache = Ref{Bool}(true)

        function $treetransformer(::AbstractTensorMap, ::AbstractTensorMap, p::Index2Tuple)
            return fusiontreetransform(f1, f2) = $transform(f1, f2, p...)
        end
        function $treetransformer(tdst::TensorMap, tsrc::TensorMap, p::Index2Tuple)
            if $usetreetransformcache[]
                key = (space(tdst), space(tsrc), p)
                A = treetransformertype(space(tdst), space(tsrc))
                return $_get_treetransformer(A, key)
            else
                return $_treetransformer((space(tdst), space(tsrc), p))
            end
        end
        @noinline function $_get_treetransformer(A, key)
            d::A = get!($treetransformcache, key) do
                return $_treetransformer(key)
            end
            return d
        end
        function $_treetransformer((Vdst, Vsrc, p))
            fusiontreetransform(f1, f2) = $transform(f1, f2, p...)
            return TreeTransformer(fusiontreetransform, Vsrc, Vdst)
        end
    end
end
```

src/tensors/indexmanipulations.jl
```diff
@@ -279,8 +279,8 @@ See also [`permute`](@ref), [`permute!`](@ref), [`add_braid!`](@ref), [`add_tran
                                           α::Number,
                                           β::Number,
                                           backend::AbstractBackend...) where {T,S,N₁,N₂}
```
Jutho: Not sure why all the type parameters are still here even after previous changes, but it seem they can be removed. Similarly for the braid and transpose methods below.

Jutho:
For other optimisations, like the ones that @maartenvd wants to implement, it would still be useful to have the old `rowr` and `colr` fields available, so I would add those to the `FusionBlockStructure`:

```julia
struct FusionBlockStructure{I,N,F₁,F₂}
    totaldim::Int
    blockstructure::SectorDict{I,Tuple{Tuple{Int,Int},UnitRange{Int}}}
    rowr::SectorDict{I,FusionTreeDict{F₁,UnitRange{Int}}}
    colr::SectorDict{I,FusionTreeDict{F₂,UnitRange{Int}}}
    fusiontreelist::Vector{Tuple{F₁,F₂}}
    fusiontreestructure::Vector{Tuple{NTuple{N,Int},NTuple{N,Int},Int}}
    fusiontreeindices::FusionTreeDict{Tuple{F₁,F₂},Int}
end
```
and then also fill them in `fusionblockstructure(W::HomSpace, ::NoCache)`.

lkdvos:
For the `rowr` and `colr`, I think it's actually not super hard to compute them from the cached data: it amounts to taking the `StridedView` information and subtracting the offset of the coupled block. (See my implementation of `block(::BraidinTensor)` (now fixed)). I'm not sure it makes sense to bundle all of this information and cache it together, as I would expect that for these kinds of optimizations, you would want to cache a custom amount of precomputed data anyways.

What I would like to have instead is `blocklist`, `blockstructure` and `blockindices`, which would allow a similar optimization in terms of `Vector` instead of `Dict` access for things that act on the blocks.

Additional notes:
- I think it would be great if there were some more tests for `BraidingTensor`
- I'm not entirely sure how to fit the `AdjointTensorMap` specializations in. It's not so much that it is hard to implement, but I would need to change the keys to the caches to also include a flag for this. In that sense, it would actually be easier to have `add_transpose!(tdst, tsrc, p, conj, alfa, beta)`, and map the `AdjointTensorMap` implementations to that, instead of doing it the other way around.
- For the case where both `src` and `dst` are `AdjointTensorMap`, I think I would have to map it to the regular case at the level of `add_permute!` and `add_braid!` etc, as the transformer is already created before `add_transform!` is called. I'm okay with this, but I am happy with alternative suggestions.

Jutho:
Benchmark results: 

[mporesults.pdf](https://github.com/user-attachments/files/17542413/mporesults.pdf)
[peporesults.pdf](https://github.com/user-attachments/files/17542415/peporesults.pdf)

Jutho:
While I see why you want a `conj` flag, it is not really the conjugate tensor, from a conceptual puristic perspective, it is not really that you can reformulate the adjoint case to the case of a conjugate tensor, as that would have conjugated fusion trees (which is not what happens) and dual spaces, i.e. `add_transform!(a, b, p)` is not not really as `add_transform(a, b', conj=true, transformed_p)`. Anyway, I am happy to try to implement this after merging this, if at least you think this is otherwise ready for merging.

Regarding the `colr` and `rowr`, I would like @maartenvd's opinion, but I think there are cases where exactly that data is useful. We anyway compute it (implicitly), so the only overhead would be from also caching it. I am not sure if we should worry about that.

lkdvos:
From the benchmark results, I'd say this is ready to merge into the jh/tensorstructure branch already, as it seems like it's a speedup in general. I would expect the results to be even more exaggerated in the non-abelian cases, so that's good news!

maartenvd:
> Regarding the `colr` and `rowr`, I would like @maartenvd's opinion, but I think there are cases where exactly that data is useful. We anyway compute it (implicitly), so the only overhead would be from also caching it. I am not sure if we should worry about that.

It could be useful if you have fusiontree manipulations acting on either the domain or codomain. But in practical planar mps code, you don't often have that (I can try to quantify it for the mpskit tests?). I am personally using it for the multiplication in 
Untitled

but there you actually need slightly different information than colr/rowr. Rather, I need to know what slice of contiguous data corresponds with the fusion trees starting with charges [q1,q2,q3] in the domain. I'm not sure if you want to cache that information for all possible q1,q2,q3,q4,..., If you want to calculate it on the other hand, then I think there is a more efficient way than trying to reconstruct it from rowr/colr.

Jutho:
Ok. then I will merge this PR as is.

In time, I do think we want a constant `Fuser<:AbstractTensorMap` that might help with what @maartenvd is doing, i.e. that would have an optimised implementation when it is contracted with the first set of spaces in the domain or codomain.

Jutho:
I think I will just merge right now. There is something strange with the CI on macOS. The test all run faster than the ones on Linux, at least for fusion trees and spaces. Also for tensors for those sectors that they have in common (but different sectors are tested on different platforms). But then when we reach the AD tests, those on MacOS take extremely long, in the order 30 minutes per sector type, whereas those on Linux (for the AD testsuite the sectors are the same) finish in about 3 - 4 minutes per sector type. (Maybe this also happens locally; I should check).

And the nightly tests only start after all the stable julia versions on all the platforms have finished.

PR #173, opened 2024-11-10
small patch to complex, real and imag

Jutho:


src/tensors/tensor.jl
```diff
@@ -389,7 +389,7 @@ function Base.complex(t::AbstractTensorMap)
     if scalartype(t) <: Complex
         return t
     else
-        return copy!(similar(t, complex(scalartype(t))), t)
+        return TensorMap(complex(t.data), space(t))
```
lkdvos: I'm not a big fan of having `TensorMap` hard-coded here. I think `similar`, by definition, should return a mutable tensor to copy the data into, which by default returns a `TensorMap`, similar to how `Array` is the default for `similar(::AbstractArray)`, but I need to overload this for `BlockTensorMap`, which is easier if everything passes through similar.

src/tensors/tensor.jl
```diff
@@ -581,27 +581,19 @@ end
...
-    if isreal(sectortype(t))
-        return TensorMap(real(t.data), codomain(t), domain(t))
-    else
-        msg = "`real` has not been implemented for `$(typeof(t))`."
-        throw(ArgumentError(msg))
-    end
+    # TODO: should we reformulate the old checks in terms of `sectorscalartype`
+    # For anyonic categories, complex numbers typically come into play only in the
+    # braiding and not in the fusion. Hence, it can make sense to work with real
+    # tensors if no braiding is required.
```
lkdvos: I wouldn't mind having `real`/`imag` work even if the fusion data is not real. I understand that it might lead to bugs here and there, but for example for singular values, it is quite natural to allow real numbers (since there are no fusion tensors anyways).

src/tensors/tensor.jl
Jutho: It feels like it's the second time I make this mistake. I forget that this method works on `AbstractTensorMap`, whereas it is in tensormap.jl. Maybe I should move it. However, I do think having the simple `TensorMap` specialisation around is also useful.

src/tensors/tensor.jl
```diff
@@ -581,27 +581,19 @@ end
 # Real and imaginary parts
 #---------------------------
 function Base.real(t::AbstractTensorMap)
-    # `isreal` for a `Sector` returns true iff the F and R symbols are real. This guarantees
-    # that the real/imaginary part of a tensor `t` can be obtained by just taking
-    # real/imaginary part of the degeneracy data.
-    if isreal(sectortype(t))
-        return TensorMap(real(t.data), codomain(t), domain(t))
```
Jutho, line 588: So this one slipped through in the previous change, as it should actually also only work for `TensorMap` arguments. That one is not entirely straightforward to define for `AbstractTensorMap`.

Jutho:
AD CI on mac is really going nuts; not sure if this now changed more due to this PR, or whether it is a random glitch. I run the test locally, and if successful, I think this is ok to merge?

PR #174, opened 2024-11-10
WIP: implementation of DiagonalTensorMap

Jutho:
I think we do want to have a `DiagonalTensorMap` type for the singular values and eigenvalues in the corresponding factorisations. Changing those tensor types arguably constitutes a breaking change, so we do not want to do this after releasing 1.0. 

In this first PR, I will just implement the type (finished) and a minimal implementation with some tests, but not yet start to use it. 

Switching `svd` and `eig`, `eigh` to use `DiagonalTensorMap` will constitute a second PR.


src/tensors/diagtensor.jl
```diff
@@ -0,0 +1,113 @@
...
+function DiagonalTensorMap(data::DenseVector{T}, V::IndexSpace) where {T}
+    return DiagonalTensorMap{T}(data, V)
+end
+
+# TODO: more constructors needed?
+
+# Special case adjoint:
+#-----------------------
+Base.adjoint(t::DiagonalTensorMap{<:Real}) = t
+Base.adjoint(t::DiagonalTensorMap{<:Complex}) = DiagonalTensorMap(conj(t.data), t.domain)
```
lkdvos: Is this a temporary solution, or would you prefer to not have to deal with `AdjointTensorMap{DiagonalTensorMap}`? In principle, in the current approach, this would work, but I agree that handling the dispatch is annoying

lkdvos (approved):
Looks very reasonable to me!
I would think that definitely here, we can probably get an efficient block iterator quite easily, which might alleviate the overhead of computing the block location, but for now, it might not even show up in profilers anyways.

src/tensors/diagtensor.jl
Jutho: I tried to mimick `LinearAlgebra.Diagonal`. It seems that the adjoint of a complex `Diagonal` object becomes:
```julia
Diagonal{ComplexF64, Base.ReshapedArray{ComplexF64, 1, Adjoint{ComplexF64, Vector{ComplexF64}}, Tuple{}}}
```
So it is not simply using the `Adjoint` wrapper, but makes sure the outer type is `Diagonal` again. I also think it is useful to make the adjoint be a `DiagonalTensorMap` instance again, so this was the easiest solution.

Jutho:
> Looks very reasonable to me!
> I would think that definitely here, we can probably get an efficient block iterator quite easily, which might alleviate the overhead of computing the block location, but for now, it might not even show up in profilers anyways.

Yes I agree the current `blocks` is not optimal, as it recomputes the offset for every block separately. One could write an iterator, or actually make that one return a `SectorDict`, as in many cases people might actually be interested in having all of them available at the same time, e.g. to make a plot of the entanglement spectrum.

src/tensors/diagtensor.jl
```diff
@@ -0,0 +1,279 @@
...
+        end
+        d′ = typeof(d)(undef, dual(d.domain))
+        for (c, b) in blocks(d)
+            f = only(fusiontrees(codomain(d), c))
+            ((f′, _), coeff) = only(permute(f, f, p₁, p₂))
+            c′ = f′.coupled
+            scale!(block(d′, c′), b, coeff)
+        end
+        return d′
+    else
```
lkdvos: Is this fair? I would think that `permute(d, ((1, 2), ()))` is still possible, it should just return a `TensorMap` instead of a `DiagonalTensorMap`

src/tensors/diagtensor.jl
```diff
@@ -0,0 +1,279 @@
...
+    return DiagonalTensorMap(one.(d.data), d.domain)
+end
+function Base.zero(d::DiagonalTensorMap)
+    return DiagonalTensorMap(zero.(d.data), d.domain)
+end
+
+function compose(d1::DiagonalTensorMap, d2::DiagonalTensorMap)
+    d1.domain == d2.domain || throw(SpaceMismatch())
+    return DiagonalTensorMap(d1.data .* d2.data, d1.domain)
+end
```
lkdvos: Should this implementation be moved to `mul!(::DiagonalTensorMap, ::DiagonalTensorMap, ::DiagonalTensorMap)`, and use `compose_dest` to figure out that the result is `Diagonal`?

src/spaces/vectorspaces.jl
```diff
@@ -111,6 +111,13 @@ field(V::ElementarySpace) = field(typeof(V))
 Return the degeneracy dimension corresponding to the sector `s` of the vector space `V`.
 """ dim(::ElementarySpace, ::Sector)
 
+@doc """
+    reduceddim(V::ElementarySpace) -> Int
+
+Return the sum of all degeneracy dimensions of the vector space `V`.
+"""
+reduceddim(V::ElementarySpace) = sum(c -> dim(V, c), sectors(V); init=0)
```
lkdvos:
```suggestion
reduceddim(V::ElementarySpace) = sum(Base.Fix1(dim, V), sectors(V); init=0)
```

src/TensorKit.jl
```diff
@@ -183,6 +184,7 @@ include("spaces/vectorspaces.jl")
 include("tensors/abstracttensor.jl")
 # include("tensors/tensortreeiterator.jl")
 include("tensors/tensor.jl")
+include("tensors/diagtensor.jl")
```
lkdvos: As a suggestion, I feel like `diagonal.jl` is more in line with `adjoint.jl` for the filenames.

src/spaces/vectorspaces.jl
Jutho: Is this supposed to be `Base.Fix1`?

src/tensors/diagtensor.jl
Jutho: It is still possible, but this method is specifically for an `Index2Tuple{1,1}`. The other options go via the generic method and yield a `TensorMap` indeed.

src/tensors/diagtensor.jl
Jutho: We can also specialise `mul!(::DiagonalTensorMap, ::DiagonalTensorMap, ::DiagonalTensorMap)`, although already now this will lead to `mul!(::Diagonal,::Diagonal,::Diagonal)` on the blocks, so it is pretty good. But indeed, it can be faster since the spaces need to be the same and so it can just act on the vector as a whole without iterating over the blocks.

Jutho:
Ok, if all tests work, I think this is getting more or less to a final stage for now.

src/spaces/vectorspaces.jl
lkdvos: yes, my bad.

src/tensors/diagtensor.jl
lkdvos: Aha, I missed that detail, thanks!

src/tensors/diagtensor.jl
lkdvos: Mostly, I meant that hoisting the implementation to `mul!` is slightly more generic, since that would then presumably also be used in the `tensorcontract!` functions, which use `mul!` instead of `compose`?

lkdvos:
Before merging, as a feature request that will be useful for PEPSKit, powers of `DiagonalTensorMap` with real exponents would be nice to define here (along with the rrules). (https://github.com/QuantumKitHub/PEPSKit.jl/blob/90375ca9b34b8c3fc2c7e3550cbdfb033163ccff/src/utility/util.jl#L26)
For negative exponents, it would also be convenient to have a function that has some kind of cutoff, but I am not sure if such a function already exists for matrices even? `pseudopower`?

Jutho:
> Before merging, as a feature request that will be useful for PEPSKit, powers of `DiagonalTensorMap` with real exponents would be nice to define here (along with the rrules). (https://github.com/QuantumKitHub/PEPSKit.jl/blob/90375ca9b34b8c3fc2c7e3550cbdfb033163ccff/src/utility/util.jl#L26)
> For negative exponents, it would also be convenient to have a function that has some kind of cutoff, but I am not sure if such a function already exists for matrices even? `pseudopower`?

Checking the matrix implementation, it is another one of those that is type unstable. For a real matrix, if `p` is integer, or `A` is positive definite, then the result is real, otherwise it becomes complex. What is furthermore funny, is this:

```julia
julia> A^0.57
2×2 Symmetric{ComplexF64, Matrix{ComplexF64}}:
  0.121409+0.155758im  -0.251072+0.107136im
 -0.251072+0.107136im   0.313729+0.0736924im

julia> Diagonal(eigvals(A))^0.57
ERROR: DomainError with -0.07888175758153752:
Exponentiation yielding a complex result requires a complex argument.
Replace x^y with (x+0im)^y, Complex(x)^y, or similar.
Stacktrace:
...
```

Because of all of this, I would very much like to postpone this to a separate PR.

lkdvos:
I'm definitely okay with having this as a separate PR. 

The results with the `Diagonal` you show are however exactly the behavior I would expect, and I would be completely okay with that throwing a `DomainError`. The cases where PEPSKit uses this, the diagonal tensor is the result of an SVD, so it shouldn't be a problem there.

src/tensors/diagonal.jl
```diff
@@ -198,10 +198,21 @@ function Base.zero(d::DiagonalTensorMap)
...
+function compose_dest(d1::DiagonalTensorMap, d2::DiagonalTensorMap)
+    A = promote_type(storagetype(d1), storagetype(d2))
+    S = spacetype(d1)
+    T = scalartype(A)
+    return DiagonalTensorMap{T,S,A}(undef, d1.domain)
+end
+
+function LinearAlgebra.mul!(dC::DiagonalTensorMap,
+                            dA::DiagonalTensorMap,
+                            dB::DiagonalTensorMap)
```
lkdvos: Probably also with the `alfa` and `beta` arguments included?

Jutho:
Ok, thanks for the extra additions @lkdvos . I think this is now ready to be merged.

PR #180, opened 2024-11-20
Fix converting empty tensors to array

### LINKED ISSUE #178: convert to Array fails for empty TensorMap
Hi! Tensorkit allows to create a dummy TensorMap without any allowed fusion tree. However converting it to `Array` errors. I think it should return `zeros(size(t))`.
```julia
t = TensorMap( U1Space(1=>1) ← U1Space(1=>1)')
convert(Array, t)
```

```julia
ERROR: UndefVarError: `A` not defined
Stacktrace:
 [1] convert(::Type{…}, t::TensorMap{…})
   @ TensorKit ~/.julia/packages/TensorKit/dtQg2/src/tensors/abstracttensor.jl:291
 [2] top-level scope
   @ REPL[3]:1
```

lkdvos:
This PR fixes #178

### DIFF
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -497,21 +497,12 @@ function Base.convert(::Type{Array}, t::AbstractTensorMap)
     else
         cod = codomain(t)
         dom = domain(t)
-        local A
+        T = promote_type(scalartype(t), sectortype(t))
+        A = zeros(T, dims(cod)..., dims(dom)...)
         for (f₁, f₂) in fusiontrees(t)
             F = convert(Array, (f₁, f₂))
-            if !(@isdefined A)
-                if eltype(F) <: Complex
-                    T = complex(float(scalartype(t)))
-                elseif eltype(F) <: Integer
-                    T = scalartype(t)
-                else
-                    T = float(scalartype(t))
-                end
-                A = fill(zero(T), (dims(cod)..., dims(dom)...))
-            end
             Aslice = StridedView(A)[axes(cod, f₁.uncoupled)..., axes(dom, f₂.uncoupled)...]
-            axpy!(1, StridedView(_kron(convert(Array, t[f₁, f₂]), F)), Aslice)
+            add!(Aslice, StridedView(_kron(convert(Array, t[f₁, f₂]), F)))
         end
         return A
     end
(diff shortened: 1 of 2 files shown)


test/tensors.jl
```diff
@@ -126,6 +126,11 @@ for V in spacelist
                         @test t === @constinferred TensorMap(t.data, W)
                     end
                 end
+                for T in (Int, Float32, ComplexF64)
+                    t = randn(T, W1 ⊗ W2, typeof(W1)())
```
Jutho: `W1`, `W2`, ... are already `HomSpace` objects; I guess you wanted `V1` and `V2`. 

Should we have a `zero(V)` to replace `typeof(V)()`? Since we don't really use `+` for the direct sum of spaces, I am not sure if recycling `zero` is really appropriate. Then again:
Screenshot 2024-11-20 at 09 30 55

Jutho (approved):

test/tensors.jl
lkdvos: Yeah, I noticed that I messed up, I tried fixing this just before I left the office yesterday. I'll fix it later today, and add the zero, since it's also just more practical

Jutho:
It would be great if you could also add `Base.zero(V::ElementarySpace)` on line 92 of "docs/src/lib/spaces.md". (If current CI has already finished, this can be with [skip ci]; the docs will be built when this is merged.)

src/tensors/abstracttensor.jl
```diff
@@ -497,7 +497,7 @@ function Base.convert(::Type{Array}, t::AbstractTensorMap)
     else
         cod = codomain(t)
         dom = domain(t)
-        T = isreal(sectorscalartype(I)) ? scalartype(t) : complex(scalartype(t))
+        T = sectorscalartype(I) <: Complex ? complex(scalartype(t)) : scalartype(t)
```
Jutho: Given that `real` and `complex` do work on `Number` subtypes to e.g. convert Float64 to ComplexF64, I understand the confusion of `isreal` not working. Maybe that is worthy of a Julia issue.

lkdvos: Yeah, it's not super consistent. I just realised I still messed up with the integer types, life is hard

PR #189, opened 2024-12-13
add 🐬  functionality

Jutho:
Flipping tensor indices, for @Adwait-Naravane and @VictorVanthilt . 

Be careful with fermions, as shown in the tests, simply flipping both of a pair of contracted indices does not yield the same contraction when fermions are involved.

### DIFF
--- a/src/tensors/indexmanipulations.jl
+++ b/src/tensors/indexmanipulations.jl
@@ -8,13 +8,18 @@ Return a new tensor that is isomorphic to `t` but where the arrows on the indice
 """
 function flip(t::AbstractTensorMap, I)
     P = flip(space(t), I)
-    t2 = similar(t, P)
-    for (c, b) in blocks(t)
-        copy!(t2[c], b)
+    t′ = similar(t, P)
+    for (f₁, f₂) in fusiontrees(t)
+        f₁′, f₂′ = f₁, f₂
+        factor = one(scalartype(t))
+        for i in I
+            (f₁′, f₂′), s = only(flip(f₁′, f₂′, i))
+            factor *= s
+        end
+        scale!(t′[f₁′, f₂′], t[f₁, f₂], factor)
     end
-    return t
+    return t′
 end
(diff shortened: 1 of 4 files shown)


Jutho:
Ok this is not correct yet. I will have to investigate further.

src/tensors/indexmanipulations.jl
```diff
@@ -1,5 +1,21 @@
...
+"""
+function flip(t::AbstractTensorMap, I)
+    P = flip(space(t), I)
+    t2 = similar(t, P)
+    for (c, b) in blocks(t)
+        copy!(t2[c], b)
+    end
+    return t
+end
+flip(t::TensorMap, I) = TensorMap(t.data, flip(space(t), I))
```
lkdvos: I think here it's probably best to also include a `copy` keyword, similar to `permute`, which can be used to control the sharing of data. At least, it's good to have this in the docstring somewhere

lkdvos (commented):
Definitely like the ability to have this!

Just as a sidenote, I'm fine if this is not possible or we don't have the time for this yet:
Do you know if there is any way to make this consistently work with fermions? For example, what happens if we define `flip` to be:

- `i` in codomain and `!isdual` -> flip
- `i` in codomain and `isdual` -> flip + twist
- `i` in domain and `!isdual` -> flip + twist
- `i` in domain and `isdual` -> flip
I think In some sense, what I am thinking of is that `contract` would basically become equivalent to `flip`ping all contracted indices to be ``<bra|ket>`` instead of the other way around, and then contracting. I'm just not entirely sure if this is consistent, as in that it commutes with arbitrary permutations + contractions

Jutho:
Ok, this turned out slightly more complicated as anticipated, because of Frobenius Schur factors. However, I think the implementation is now fixed, and includes the suggestion of @lkdvos to also include twists. I think this guarantees that if both spaces in a pair of `@tensor` contracted indices are flipped, then the contraction is unaltered, irrespective of whether either of the indices live in the domain or codomain. This might fail in an `@planar` context however.

PR #190, opened 2024-12-17
use diagonaltensormap

### LINKED ISSUE #181: convert to floating point in matrix factorisations
```julia-repl
julia> t = TensorMap([1 0 ; 0 -1], ℂ^2, ℂ^2)
TensorMap(ℂ^2 ← ℂ^2):
 1   0
 0  -1


julia> tsvd(t)
ERROR: MethodError: no method matching svd!(::Base.ReshapedArray{Int64, 2, SubArray{Int64, 1, Vector{Int64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, ::SDD)
The function `svd!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  svd!(::StridedMatrix{T}, ::Union{SDD, SVD}) where T<:Union{Float32, Float64, ComplexF64, ComplexF32}
   @ TensorKit ~/.julia/dev/TensorKit/src/auxiliary/linalg.jl:274
```

Jutho:
This starts using the `DiagonalTensorMap` type in svd and eigenvalue decompositions. It also fixes #181 .

### DIFF
--- a/src/tensors/factorizations.jl
+++ b/src/tensors/factorizations.jl
@@ -1,5 +1,18 @@
 # Tensor factorization
 #----------------------
+function factorisation_scalartype(t::AbstractTensorMap)
+    T = scalartype(t)
+    return promote_type(Float32, typeof(zero(T) / sqrt(abs2(one(T)))))
+end
+factorisation_scalartype(f, t) = factorisation_scalartype(t)
+
+function permutedcopy_oftype(t::AbstractTensorMap, T::Type{<:Number}, p::Index2Tuple)
+    return permute!(similar(t, T, permute(space(t), p)), t, p)
+end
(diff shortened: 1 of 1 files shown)


lkdvos (approved):
I dug a bit into the LinearAlgebra code, for the scalartype they use: https://github.com/JuliaLang/LinearAlgebra.jl/blob/1137b4c7fa8297cef17c4ae0982d7d89d4ab7dd8/src/eigen.jl#L319
I don't know if it makes a difference what the actual implementation is, but I quite like if there is a single function that could be overloaded if needed to control this behavior.

Otherwise looks good to me!

Jutho:
@lkdvos , is the latest commit more in line but what you had in mind?

lkdvos:
Yes, exactly! I also like that you factored out the "copy of type" function, looks very clean to me!

PR #195, opened 2025-01-06
Fix similar

### LINKED ISSUE #194: `similar` does not work if storagetype is a PtrArray
It seems that TensorKit.jl v0.14.1 is currently not well compatible with TensorOperations.jl. I try to use the new feature `allocator = ManualAllocator()` to decrease the memory pressure, which however results in an error in certain cases. The error is finally due to a `similar` call in `@tensor` macro, when a permutation is needed. Below is a minimal code example to reproduce the error:
```julia
using TensorKit
using TensorKit.TensorOperations
using TensorKit.TensorOperations: tensoralloc_contract, ManualAllocator

A = id(Rep[SU₂](0=>1))
B = id(Rep[SU₂](0=>1))

t = tensoralloc_contract(Float64, A, ((1, ), (2,)), false, B, ((1,), (2,)), false, ((1,), (2,)), Val{true}(), ManualAllocator())

similar(t) #  MethodError: no method matching PtrArrays.PtrArray{Float64, 1}(::UndefInitializer, ::Int64)
```  
Temporarily, I add a constructor
```julia
function TensorMap{T, S, N₁, N₂, A}(::UndefInitializer,
	space::TensorMapSpace{S, N₁, N₂}) where {T, S <: IndexSpace,
	N₁, N₂,
	A <: PtrArray{T, 1}}
	d = fusionblockstructure(space).totaldim
	# data = A(undef, d) # this line will lead to an error 
     data = Array{T, 1}(undef, d)
	if !isbitstype(T)
		zerovector!(data)
	end
	return TensorMap{T, S, N₁, N₂, Array{T, 1}}(data, space)
end
```
to make my code continue running. Since there is not a `tensorfree!` for this tensor in `@tensor`, I construct a common tensor to avoid potential memory leaks. Is this necessary?

Any suggestions on the temporary solution would be greatly appreciated. I also hope the future version of TensorKit.jl can provide an official one to fix this issue.

Jutho:
This should fix #194 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TensorKit"
 uuid = "07d1fe3e-3e46-537d-9eac-e9e13d0d4cec"
 authors = ["Jutho Haegeman"]
-version = "0.14.1"
+version = "0.14.2"
 
 [deps]
 LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -67,6 +67,8 @@ field(::Type{TT}) where {TT<:AbstractTensorMap} = field(spacetype(TT))
 Return the type of vector that stores the data of a tensor.
 """ storagetype
 
+similarstoragetype(TT::Type{<:AbstractTensorMap}) = similarstoragetype(TT, scalartype(TT))
+
 function similarstoragetype(TT::Type{<:AbstractTensorMap}, ::Type{T}) where {T}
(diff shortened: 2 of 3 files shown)


Jutho:
Not sure why this causes type instabilities in `rrule`s specifically, but I will investigate.

lkdvos:
Failures seem to be unrelated: https://github.com/Jutho/TensorKit.jl/actions/runs/12673053638/job/35318509216

Jutho:
So we merge this and open a new issue or PR for the type instabilities in the new Zygote version?

lkdvos:
Ok for me I think

PR #202, opened 2025-01-16
`rrule` for `copy_oftype`, `permutedcopy_oftype`

### LINKED ISSUE #201: AD of SVD question
Hi! 

I am not sure if this is a bug or I am doing something wrong, but it seems that there are some issues in the chain rules for tsvd.

## example 1

```
using TensorKit
using Zygote
function tst(A)
	U, S, V, = tsvd(A)
	return tr(S)
end
A0 = TensorMap(randn, Z2Space(4,4)←Z2Space(4,4))
gradient(tst, A0)
```
In this example Zygote complains about mutating arrays somewhere in the code. The error message:
```
ERROR: Mutating arrays is not supported -- called copyto!(Vector{Float64}, ...)
This error occurs when you ask Zygote to differentiate operations that change
the elements of arrays in place (e.g. setting values with x .= ...)
```

## example 2

```
using TensorKit
using Zygote
function tst(A)
	U, S, V, = tsvd(A)
	return tr(U*V)
end
A0 = TensorMap(randn, Z2Space(4,4)←Z2Space(4,4))
gradient(tst, A0)
``` 
This test function looks quite similar, but the error is different. Here is the error message:

```
ERROR: MethodError: no method matching block(::ChainRulesCore.Thunk{Zygote.var"#163#164"{Nothing, ChainRulesCore.Thunk{…}}}, ::Z2Irrep)
The function `block` exists, but no method is defined for this combination of argument types.
 ```

### PR DESCRIPTION
**Pull request by lkdvos**
Fixes #201

### DIFF
--- a/ext/TensorKitChainRulesCoreExt/constructors.jl
+++ b/ext/TensorKitChainRulesCoreExt/constructors.jl
@@ -29,7 +29,7 @@ function ChainRulesCore.rrule(::typeof(TensorKit.permutedcopy_oftype), t::Abstra
     project = ProjectTo(t)
     function permutedcopy_oftype_pullback(Δt)
         invp = TensorKit._canonicalize(TupleTools.invperm(linearize(p)), t)
-        return project(TensorKit.permutedcopy_oftype(unthunk(Δt), scalartype(t), invp))
+        return project(TensorKit.permute(unthunk(Δt), invp))
     end
     return TensorKit.permutedcopy_oftype(t, T, p), permutedcopy_oftype_pullback
 end
--- a/ext/TensorKitChainRulesCoreExt/constructors.jl
+++ b/ext/TensorKitChainRulesCoreExt/constructors.jl
@@ -29,7 +29,8 @@ function ChainRulesCore.rrule(::typeof(TensorKit.permutedcopy_oftype), t::Abstra
     project = ProjectTo(t)
     function permutedcopy_oftype_pullback(Δt)
         invp = TensorKit._canonicalize(TupleTools.invperm(linearize(p)), t)
-        return NoTangent(), project(TensorKit.permute(unthunk(Δt), invp)), NoTangent(), NoTangent()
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by lkdvos** [comment]
I think the remaining error is most easily fixed by just doing permuting and then projecting, instead of trying to do both at the same time. It’s not the most efficient, but until it actually shows up as a problem somewhere I wouldn’t worry too much about it.

File: ext/TensorKitChainRulesCoreExt/constructors.jl [hunk 1]
```diff
@@ -17,6 +17,23 @@ function ChainRulesCore.rrule(::typeof(Base.copy), t::AbstractTensorMap)
...
+    copy_oftype_pullback(Δt) = NoTangent(), project(unthunk(Δt)), NoTangent()
+    return TensorKit.copy_oftype(t, T), copy_oftype_pullback
+end
+
+function ChainRulesCore.rrule(::typeof(TensorKit.permutedcopy_oftype), t::AbstractTensorMap,
+                              T::Type{<:Number}, p::Index2Tuple)
+    project = ProjectTo(t)
+    function permutedcopy_oftype_pullback(Δt)
+        invp = TensorKit._canonicalize(TupleTools.invperm(linearize(p)), t)
+        return project(TensorKit.permutedcopy_oftype(unthunk(Δt), scalartype(t), invp))
```
Jutho: This doesn't work if `T` (which should coincide with `scalartype(Δt)`) is complex and `scalartype(t)` is real. This shouldn't really happen in the case where this function is being used (namely in the factorisations, where the purpose is to go from `Integer` or `Rational` types to floating point types, but not to change between real and complex). This triggers the question: should we just modify the tests, or make this work in general? There is no direct functionality to combine taking the real part and permuting.

Jutho: Oh I hadn't see your comment.

lkdvos (commented):
Should be good to go for me

Jutho:
The joy of online editing: after finally getting it right, the formatting was messed up :smile:

PR #203, opened 2025-01-17
Multi-threading attempt III

lkdvos:
This is a continuation of #100 and #117 in an attempt to properly address multithreading over blocks in the various parts of the code.

To achieve this, I added:
- `backend, allocator` support to the TensorOperations functions
- `backend, allocator` support to the indexmanipulations
- a `TensorKitBackend` that holds a scheduler and a backend to pass on
- `BlockIterator` to avoid having to find the cached structures in a multithreaded loop, reducing overall cache lookups and hopefully avoiding lock contention

Before continuing and pushing this through to the other functions, some questions:
- all introduced names are up for discussion
- should `mul!` take a scheduler or a backend?
- do we want to remove some functions that are now duplicates with slightly altered functionality? `permute!`, `add_permute!` and `tensoradd!` all do more or less the same thing
- Is it fair to maybe just import the `TensorOperations` functions, and write everything in terms of that?
- Various comments throughout the code itself


src/tensors/tensor.jl
```diff
@@ -559,18 +566,22 @@ function Base.show(io::IO, t::TensorMap)
...
+        #         println(io, "* Data for sector ", f₁.uncoupled, " ← ", f₂.uncoupled, ":")
+        #         Base.print_array(io, t[f₁, f₂])
+        #         println(io)
+        #     end
+        # else
+        #     for (f₁, f₂) in fusiontrees(t)
+        #         println(io, "* Data for fusiontree ", f₁, " ← ", f₂, ":")
+        #         Base.print_array(io, t[f₁, f₂])
+        #         println(io)
+        #     end
```
Jutho: Is this a temporary change or would you like to change printing for `TensorMap`s?

src/tensors/tensoroperations.jl
```diff
@@ -144,6 +144,29 @@ TO.tensorcost(t::AbstractTensorMap, i::Int) = dim(space(t, i))
 # IMPLEMENTATONS
 #----------------
 
+# Backend implementation
+# ----------------------
+# TODO: figure out a name
+# TODO: what should be the default scheduler?
+# TODO: should we allow a separate scheduler for "blocks" and "subblocks"
```
Jutho: Yes please. Or we should at least have different defaults. But given that a method like `tensorcontract!` wants to do both a permutation step (subblocks) and a multiplication (blocks), it would need to know about both schedulers.

src/tensors/tensor.jl
lkdvos: I personally prefer printing the big blocks, simply because it is a lot more compact. Inspecting the contents of two tensors is quite annoying with the fusiontrees in the way. Here I just temporarily changed it because I was debugging and forgot to change it back, I would definitely do that in a separate PR anyways

Jutho:
I've only browsed through this quickly so far. The design is definitely different than what I had in mind. What I had in mind was that "all" TensorKit functions would accept three final arguments (probably in this order):
* `scheduler`
* `backend`
* `allocator`

where `backend` and `allocator` are just passed through to the relevant TensorOperations or MatrixAlgebraKit method (and could thus take its values from there), and `scheduler` is specific to the TensorKit method and is used in deciding how to tun over coupled charge blocks or over fusiontree subblocks. 

Ultimately that is of course the same, except for the fact that now you first have to wrap the underlying TensorOperations `Backend` or MatrixAlgebraKit `Algorithm` (as it is called there) in a `TensorKitBackend`. That is easy to do for the defaults, but maybe a bit more cumbersome for the experiments. Also, suppose you want to specify a scheduler, but don't care about the backend. Is that easy to do? Do we need to harmonize the default backend structure between TensorOperations and MatrixAlgebraKit ?

Jutho:
Also, as response to the question, yes also `mul!` is supposed to become part of MatrixAlgebraKit and potentially support different backends/algorithms (I guess Octavian is not going to happen, but there might still be cases where interesting different matrix multiplication backends exist).

lkdvos:
I think having some kind of wrapper is a bit inevitable, since for example in `tensorcontract!` I would need a `mul!` backend, a backend for the tensoroperations itself, a scheduler, an allocator, etc. If there is only a single `backend`, then `mul!` would alos have to support the tensoroperations backends, so that might not be the most convenient. In any case, I liked having them together as a single argument just to not have too many, but this is of course equivalent to a tuple...

The logic of using a `DefaultBackend`, which is equivalent to selecting a backend at runtime seems to work rather well, ie in the current implementation I pass `DefaultBackend` as the "array backend" to the tensoroperations functions, which then select the correct backend a bit further down the stack. I would be okay with having a similar implementation for the `scheduler` and `allocator` as well if that helps.

One thing that I am kind of becoming more and more in favour of is the idea of simply putting the `allocator` and `scheduler` (and maybe even the `backend` for functions other than TensorOperations) in a scoped value, instead of explicitly passing them around. Realistically, it's not actually that convenient to change the arguments of `mul!` or `tsvd!` calls, since these are typically hidden in some lower levels of the code, and passing them all the way up is a bit of a pain. For MPSKit, we would have to rewrite almost all code to pass these around, and I think this is precisely what `ScopedValues` should solve.

In some sense, what I would see as a balance between these things is:
- The direct functions such as `mul!`, `tsvd!` etc have a final argument `backend` that can be used to control the implementation, which includes the necessary configurations: a `scheduler` where applicable, an `allocator`, and possibly some additional options for the implementation.
- These final arguments have a default value that can be controlled via scoped values, such that it becomes easy to alter from anywhere in the callstack. This is achieved via the `select_backend` pattern we have in TensorOperations
- In order to allow for "dynamically selecting a default value", `DefaultBackend` can be used to control subalgorithms, which will then select a backend using `select_backend` again.

To keep consistency, I'm also okay with having `backend, allocator` as final arguments, but introducing another `scheduler` seems like it should just be part of the `backend`, since it is really implementation-dependent what that should be and whether or not it is present. It would make more sense to me to have something like the following to indicate a parallelized implementation of `algorithm` over blocks through a `scheduler`.
```julia
struct BlockAlgorithm <: AbstractAlgorithm
    scheduler
    algorithm
end
```

Jutho:
I am definitely in favor of something that can be controlled via scoped values.

src/tensors/backends.jl
```diff
@@ -0,0 +1,107 @@
...
+function select_scheduler(scheduler=OhMyThreads.Implementation.NotGiven(); kwargs...)
+    return if scheduler == OhMyThreads.Implementation.NotGiven() && isempty(kwargs)
+        Threads.nthreads() == 1 ? SerialScheduler() : DynamicScheduler()
+    else
+        OhMyThreads.Implementation._scheduler_from_userinput(scheduler; kwargs...)
+    end
+end
+
+"""
+    set_blockscheduler!([scheduler]; kwargs...) -> previuos
```
Jutho:
```suggestion
    set_blockscheduler!([scheduler]; kwargs...) -> previous
```

src/tensors/indexmanipulations.jl
```diff
@@ -434,18 +434,54 @@ See also [`transpose`](@ref), [`transpose!`](@ref), [`add_permute!`](@ref), [`ad
...
+        TC = typeof(C)
+        TA = typeof(A)
+        throw(ArgumentError("No suitable backend found for `add_transform!` and tensor types $TC and $TA"))
+    else # error for unknown backend
+        TC = typeof(C)
+        TA = typeof(A)
+        throw(ArgumentError("Unknown backend $backend for `add_transform!` and tensor types $TC and $TA"))
+    end
+end
+function TO.select_backend(::typeof(add_transform!), C::AbstractTensorMap,
```
Jutho: Should this one be moved into backends.jl like the ones for `TO.tensoradd!` and friends.

src/tensors/linalg.jl
```diff
@@ -351,7 +386,21 @@ function LinearAlgebra.mul!(tC::AbstractTensorMap,
     return tC
 end
 
-# TODO: consider spawning threads for different blocks, support backends
+function threaded_mul!(tC::AbstractTensorMap, tA::AbstractTensorMap, tB::AbstractTensorMap,
+                       α::Number, β::Number, scheduler::Scheduler)
+    # obtain cached data before multithreading
+    bCs, bAs, bBs = blocks(tC), blocks(tA), blocks(tB)
+
+    tforeach(blocksectors(tC); scheduler) do c
```
Jutho, line 396: Is it not possible to do
```julia
    tforeach(bCs; scheduler) do (c, bC)
        if haskey(bAs, c) # then also bBs should have it
            mul!(bC, bAs[c], bBs[c], α, β)
        elseif !isone(β)
            scale!(bC, β)
        end
    end
```

Jutho (commented):
A first set of comments; I still have to go through src/tensors/tensoroperations.jl, but this looks very promising!

ZongYongyue:
There is a strange bug in function ``threaded_mul!``— indexing ``bAs`` and ``bBs`` throws an error. They only work correctly when converted into dictionaries, but this shouldn’t be the case. ``blocks(tA)`` and ``blocks(tB)`` should definitely be indexable.

ZongYongyue:
Some feedback:
1.Using ```TensorKit.set_blockscheduler!(:dynamic)``` throws an error:
  ```julia
ERROR: LoadError: MethodError: no method matching setindex!(::ScopedValue{Scheduler}, ::DynamicScheduler{OhMyThreads.Schedulers.FixedCount, ChunkSplitters.Consecutive})
The function `setindex!` exists, but no method is defined for this combination of argument types.
```
2.When ``TensorKit.TensorKitBackend().blockscheduler`` and ``TensorKit.TensorKitBackend().subblockscheduler`` are set to ``SerialScheduler``, ``threaded_mul!`` is called, and an error occurs when indexing bAs and bBs at ```mul!(bC, bAs[c], bBs[c], α, β)```
```julia
ERROR: LoadError: MethodError: no method matching getindex(::Base.Generator{Vector{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}}, TensorKit.var"#157#158"{BraidingTensor{Float64, GradedSpace{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Int64}}}}}, ::ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}})
```
3.When ``TensorKit.TensorKitBackend().blockscheduler`` and ``TensorKit.TensorKitBackend()``.subblockscheduler are set to ``DynamicScheduler``, an error occurs at ``tforeach(bCs; scheduler) do (c, bC)``:
```julia
ERROR: LoadError: ArgumentError: Arguments of type TensorKit.BlockIterator{TensorMap{Float64, GradedSpace{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Int64}}, 2, 1, Vector{Float64}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Tuple{Tuple{Int64, Int64}, UnitRange{Int64}}}} are not compatible with chunks, either implement a custom chunks method for your type, or implement the custom type interface (see https://juliafolds2.github.io/ChunkSplitters.jl/dev/)
```

ZongYongyue:
In the second case, I changed bAs and bBs to dicts, and the program ran successfully. I then simulated a Hubbard model of size 2×5 with D=512 and found that the speed was faster than both the current master version and the old version.

ZongYongyue:
``add!`` throws an error now:
```julia
Warning: The function `add!` is not implemented for (values of) type `Tuple{Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, Float64, VectorInterface.One, VectorInterface.One}`;
│ this fallback will disappear in future versions of VectorInterface.jl
└ @ VectorInterface ~/.julia/packages/VectorInterface/J6qCR/src/fallbacks.jl:143
ERROR: LoadError: ArgumentError: No fallback for applying `add!` to (values of) type `Tuple{Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, Float64, VectorInterface.One, VectorInterface.One}` could be determined
Stacktrace:
  [1] add!(y::Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, x::Float64, α::VectorInterface.One, β::VectorInterface.One)
    @ VectorInterface ~/.julia/packages/VectorInterface/J6qCR/src/fallbacks.jl:150
  [2] add!(ty::TensorMap{Float64, GradedSpace{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Int64}}, 2, 2, Vector{Float64}}, tx::BraidingTensor{Float64, GradedSpace{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Int64}}}, α::VectorInterface.One, β::VectorInterface.One)
    @ TensorKit ~/Library/Mobile Documents/com~apple~CloudDocs/Clone/Jutho/TensorKit.jl-ld-multithreading2/src/tensors/vectorinterface.jl:77
  [3] add!
    @ ~/.julia/packages/VectorInterface/J6qCR/src/interface.jl:124 [inlined]
  [4] add(ty::TensorMap{Float64, GradedSpace{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Int64}}, 2, 2, Vector{Float64}}, tx::BraidingTensor{Float64, GradedSpace{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Int64}}}, α::VectorInterface.One, β::VectorInterface.One)
    @ TensorKit ~/Library/Mobile Documents/com~apple~CloudDocs/Clone/Jutho/TensorKit.jl-ld-multithreading2/src/tensors/vectorinterface.jl:71
  [5] add
    @ ~/.julia/packages/VectorInterface/J6qCR/src/interface.jl:107 [inlined]
  [6] +(t1::TensorMap{Float64, GradedSpace{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Int64}}, 2, 2, Vector{Float64}}, t2::BraidingTensor{Float64, GradedSpace{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, U1Irrep, U1Irrep}}, Int64}}})
    @ TensorKit ~/Library/Mobile Documents/com~apple~CloudDocs/Clone/Jutho/TensorKit.jl-ld-multithreading2/src/tensors/linalg.jl:7
```

lkdvos:
Thanks for the feedback!
For 1. I just forgot how ScopedValues work, and you can't actually change the default like that. I simply removed those methods
For 2. this was a problem with the blockiterator of `BraidingTensor`, which I had forgotten about. This should now be resolved. (I also added in some specializations, which might even avoid ever getting to that part of the code to begin with) (EDIT: hopefully fixed now?)

For 3. this is an interesting thing we might want to work around or with: the collections that `OhMyThreads` can handle need to have an implementation for `ChunkSplitters.jl`. The easiest way to accomplish this is to simply use `blocksectors` instead, since that avoids many of the other issues.
The alternative is to actually try and support `ChunkSplitters` on the blockiterators directly, but that requires us to either implement `firstindex`, `lastindex` `length` and `view` on them, or directly write the implementations for `chunks` and `index_chunks`. I think for now, simply using `blocksectors` is the easiest way forwards, because otherwise we might have to interpret `BlockIterator <: AbstractVector`, and I'm not sure that's something we should do.

ZongYongyue:
Thank you very much for your prompt fix -- it has resolved all the bugs I have found.

ZongYongyue:
Possible bug? In multithreading, using SU(2) symmetry causes @planar to throw an error, but using only U(1) symmetry does not:

```julia
TensorKit.with_blockscheduler(DynamicScheduler()) do
    TensorKit.with_subblockscheduler(DynamicScheduler()) do
        E = e_plus(Float64, SU2Irrep, U1Irrep; side=:L, filling=filling)'
        F = isomorphism(storagetype(E), flip(space(E, 2)), space(E, 2))
        @planar e⁻[-1; -2 -3] := E[-1 1; -2] * F[-3; 1]
    end
end

ERROR: LoadError: ArgumentError: Arguments of type Base.Iterators.ProductIterator{Tuple{Base.Iterators.ProductIterator{Tuple{TensorKitSectors.SectorSet{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, TensorKit.var"#93#94"{GradedSpace{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, Int64}}}, Vector{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}}}, TensorKitSectors.SectorSet{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, TensorKit.var"#93#94"{GradedSpace{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, Int64}}}, Vector{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}}}}}, Base.Iterators.ProductIterator{Tuple{TensorKitSectors.SectorSet{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, TensorKit.var"#93#94"{GradedSpace{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, TensorKit.SortedVectorDict{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}, Int64}}}, Vector{ProductSector{Tuple{FermionParity, SU2Irrep, U1Irrep}}}}}}}} are not compatible with chunks, either implement a custom chunks method for your type, or implement the custom type interface (see https://juliafolds2.github.io/ChunkSplitters.jl/dev/)
Stacktrace:

TensorKit.with_blockscheduler(DynamicScheduler()) do
    TensorKit.with_subblockscheduler(DynamicScheduler()) do
        E = e_plus(Float64, U1Irrep, U1Irrep; side=:L, filling=filling)'
        F = isomorphism(storagetype(E), flip(space(E, 2)), space(E, 2))
        @planar e⁻[-1; -2 -3] := E[-1 1; -2] * F[-3; 1]
        println(e⁻)
    end
end

TensorMap(Vect[(FermionParity ⊠ Irrep[U₁] ⊠ Irrep[U₁])]((0, 0, 1)=>1, (0, 0, -1)=>1, (1, 1, 0)=>1, (1, -1, 0)=>1) ← (Vect[(FermionParity ⊠ Irrep[U₁] ⊠ Irrep[U₁])]((0, 0, 1)=>1, (0, 0, -1)=>1, (1, 1, 0)=>1, (1, -1, 0)=>1) ⊗ Vect[(FermionParity ⊠ Irrep[U₁] ⊠ Irrep[U₁])]((1, -1, -1)=>1))):
* Data for sector ((FermionParity(0) ⊠ Irrep[U₁](0) ⊠ Irrep[U₁](-1)),) ← ((FermionParity(1) ⊠ Irrep[U₁](1) ⊠ Irrep[U₁](0)), (FermionParity(1) ⊠ Irrep[U₁](-1) ⊠ Irrep[U₁](-1))):
[:, :, 1] =
 1.0
* Data for sector ((FermionParity(1) ⊠ Irrep[U₁](-1) ⊠ Irrep[U₁](0)),) ← ((FermionParity(0) ⊠ Irrep[U₁](0) ⊠ Irrep[U₁](1)), (FermionParity(1) ⊠ Irrep[U₁](-1) ⊠ Irrep[U₁](-1))):
[:, :, 1] =
 -1.0
```

lkdvos:
Could you also attach what `e_plus` does so I can reproduce the error?

lkdvos:
Should now be resolved. Interesting to note here is that this occurs when permuting an `AdjointTensorMap`, which is still not taking the "fast implementation" codepath. A profiler might point out if this is worth specializing as well.

Confusio:
Any updates on the PR? I was wondering if a working version might come out soon.

Issue #204, opened 2025-01-17
a small issue of leftorth!


guochu:
Dear authors:
        In the function leftorth! (line 314 in factorizations.jl), if the input has empty blocks, but with a nontrivial domain, it will raise an error of "QRdata not defined". This is because you have used "isempty(blocks(t))" in line 328 but "isempty(blocksectors(domain(t)))" in line 355, which may not always be the same.
        Could you fix this small issue? You may simply change "isempty(blocksectors(domain(t)))"  into "isempty(blocks(t))" in line 355.

best!
guochu

PR #205, opened 2025-01-17
Fix leftorth! and rightorth! with empty blocks

Jutho:
Issue #204


Jutho:
Should we tag a patch release after this is finished and merged?

lkdvos:
I would say there is no reason not to.

PR #206, opened 2025-01-17
Add dedicated `BlockIterator`s

lkdvos:
This PR adds some dedicated iterators for going through the blocks of a single tensor.
This replaces the previous default implementation in terms of `Iterators.map`, in order to have more type stability (well-defined `eltype` and `length`), and the option to index into the iterators.
This should otherwise not change any external functionality (and in particular, it re-enables the syntax `blocks(t)[c]`, which worked before v0.13) so I think it could be considered non-breaking.

### DIFF
--- a/src/spaces/homspace.jl
+++ b/src/spaces/homspace.jl
@@ -349,3 +349,18 @@ function fusionblockstructure(W::HomSpace, ::GlobalLRUCache)
     end
     return structure
 end
+
+# Diagonal ranges
+#----------------
+# TODO: is this something we want to cache?
+function diagonalblockstructure(W::HomSpace)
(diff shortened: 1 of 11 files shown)


src/spaces/homspace.jl
```diff
@@ -349,3 +349,18 @@ function fusionblockstructure(W::HomSpace, ::GlobalLRUCache)
     end
     return structure
 end
+
+# Diagonal ranges
+#----------------
+# TODO: is this something we want to cache?
+function diagonalblockstructure(W::HomSpace)
```
Jutho, line 355: If we consider this part of the official interface, we might want to add a check for `W isa HomSpace{S,1,1}` and `W.domain[1] == W.codomain[1]`?

PR #208, opened 2025-01-19
DiagonalTensorMap constructor rrule 

ebelnikola:
This PR adds `rrule` for the `DiagonalTensorMap` constructor and defines `ProjectTo` for `DiagonalTensorMap`.

### DIFF
--- a/ext/TensorKitChainRulesCoreExt/constructors.jl
+++ b/ext/TensorKitChainRulesCoreExt/constructors.jl
@@ -12,8 +12,10 @@ function ChainRulesCore.rrule(::Type{TensorMap}, d::DenseArray, args...; kwargs.
     return TensorMap(d, args...; kwargs...), TensorMap_pullback
 end
 
-# these are not the conversion from array, but actually take in data parameters
-# -- as a result, requires quantum dimensions
+# these are not the conversion to/from array, but actually take in data parameters
+# -- as a result, requires quantum dimensions to keep inner product the same:
+#   ⟨Δdata, ∂data⟩ = ⟨Δtensor, ∂tensor⟩ = ∑_c d_c ⟨Δtensor_c, ∂tensor_c⟩
+#   ⟹ Δdata = d_c Δtensor_c
 function ChainRulesCore.rrule(::Type{TensorMap{T}}, data::DenseVector,
                               V::TensorMapSpace) where {T}
     t = TensorMap{T}(data, V)
@@ -21,7 +23,7 @@ function ChainRulesCore.rrule(::Type{TensorMap{T}}, data::DenseVector,
     function TensorMap_pullback(Δt_)
         Δt = copy(unthunk(Δt_))
         for (c, b) in blocks(Δt)
-            scale!(b, TensorKit.invsqrtdim(c))
+            scale!(b, dim(c))
         end
         ∂data = P(Δt.data)
         return NoTangent(), ∂data, NoTangent()
@@ -37,7 +39,7 @@ function ChainRulesCore.rrule(::Type{<:DiagonalTensorMap}, data::DenseVector, ar
         # unclear if we're allowed to modify/take ownership of the input
         Δt = copy(unthunk(Δt_))
         for (c, b) in blocks(Δt)
-            scale!(b, TensorKit.invsqrtdim(c))
+            scale!(b, dim(c))
         end
         ∂data = P(Δt.data)
         return NoTangent(), ∂data, NoTangent()
@@ -51,7 +53,7 @@ function ChainRulesCore.rrule(::typeof(Base.getproperty), t::TensorMap, prop::Sy
             # unclear if we're allowed to modify/take ownership of the input
             t′ = typeof(t)(copy(unthunk(Δdata)), t.space)
             for (c, b) in blocks(t′)
-                scale!(b, TensorKit.sqrtdim(c))
+                scale!(b, inv(dim(c)))
             end
             return NoTangent(), t′, NoTangent()
         end
@@ -70,7 +72,7 @@ function ChainRulesCore.rrule(::typeof(Base.getproperty), t::DiagonalTensorMap,
             # unclear if we're allowed to modify/take ownership of the input
             t′ = typeof(t)(copy(unthunk(Δdata)), t.domain)
             for (c, b) in blocks(t′)
-                scale!(b, TensorKit.sqrtdim(c))
+                scale!(b, inv(dim(c)))
             end
             return NoTangent(), t′, NoTangent()
         end
--- a/ext/TensorKitChainRulesCoreExt/constructors.jl
+++ b/ext/TensorKitChainRulesCoreExt/constructors.jl
@@ -12,6 +12,16 @@ function ChainRulesCore.rrule(::Type{<:TensorMap}, d::DenseArray, args...; kwarg
     return TensorMap(d, args...; kwargs...), TensorMap_pullback
 end
 
+function ChainRulesCore.rrule(::Type{<:DiagonalTensorMap}, d::DenseVector, args...; kwargs...)
+    D=TensorMap(d, args...; kwargs...)
+    project_D=ProjectTo(D)
+    function DiagonalTensorMap_pullback(Δt)
+        ∂d = project_D(unthunk(Δt)).data
+        return NoTangent(), ∂d, ntuple(_ -> NoTangent(), length(args))...
+    end
+    return D, DiagonalTensorMap_pullback
+end
+
 function ChainRulesCore.rrule(::typeof(Base.copy), t::AbstractTensorMap)
     copy_pullback(Δt) = NoTangent(), Δt
     return copy(t), copy_pullback
--- a/ext/TensorKitChainRulesCoreExt/constructors.jl
+++ b/ext/TensorKitChainRulesCoreExt/constructors.jl
@@ -21,7 +21,7 @@ function ChainRulesCore.rrule(::Type{TensorMap{T}}, data::DenseVector,
     function TensorMap_pullback(Δt_)
         Δt = copy(unthunk(Δt_))
         for (c, b) in blocks(Δt)
-            scale!(b, TensorKit.sqrtdim(c))
+            scale!(b, TensorKit.invsqrtdim(c))
         end
         ∂data = P(Δt.data)
         return NoTangent(), ∂data, NoTangent()
@@ -37,7 +37,7 @@ function ChainRulesCore.rrule(::Type{<:DiagonalTensorMap}, data::DenseVector, ar
         # unclear if we're allowed to modify/take ownership of the input
         Δt = copy(unthunk(Δt_))
         for (c, b) in blocks(Δt)
-            scale!(b, TensorKit.sqrtdim(c))
+            scale!(b, TensorKit.invsqrtdim(c))
         end
         ∂data = P(Δt.data)
         return NoTangent(), ∂data, NoTangent()
@@ -51,7 +51,7 @@ function ChainRulesCore.rrule(::typeof(Base.getproperty), t::TensorMap, prop::Sy
             # unclear if we're allowed to modify/take ownership of the input
             t′ = typeof(t)(copy(unthunk(Δdata)), t.space)
             for (c, b) in blocks(t′)
-                scale!(b, TensorKit.invsqrtdim(c))
+                scale!(b, TensorKit.sqrtdim(c))
             end
             return NoTangent(), t′, NoTangent()
         end
@@ -70,7 +70,7 @@ function ChainRulesCore.rrule(::typeof(Base.getproperty), t::DiagonalTensorMap,
             # unclear if we're allowed to modify/take ownership of the input
             t′ = typeof(t)(copy(unthunk(Δdata)), t.domain)
             for (c, b) in blocks(t′)
-                scale!(b, TensorKit.invsqrtdim(c))
+                scale!(b, TensorKit.sqrtdim(c))
             end
             return NoTangent(), t′, NoTangent()
         end
--- a/ext/TensorKitChainRulesCoreExt/constructors.jl
+++ b/ext/TensorKitChainRulesCoreExt/constructors.jl
@@ -4,7 +4,7 @@
 @non_differentiable TensorKit.isometry(args...)
 @non_differentiable TensorKit.unitary(args...)
 
-function ChainRulesCore.rrule(::Type{<:TensorMap}, d::DenseArray, args...; kwargs...)
+function ChainRulesCore.rrule(::Type{TensorMap}, d::DenseArray, args...; kwargs...)
     function TensorMap_pullback(Δt)
         ∂d = convert(Array, unthunk(Δt))
         return NoTangent(), ∂d, ntuple(_ -> NoTangent(), length(args))...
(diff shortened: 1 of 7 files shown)


ext/TensorKitChainRulesCoreExt/constructors.jl
```diff
@@ -12,6 +12,17 @@ function ChainRulesCore.rrule(::Type{<:TensorMap}, d::DenseArray, args...; kwarg
...
 end
 
+function ChainRulesCore.rrule(::Type{<:DiagonalTensorMap}, d::DenseVector, args...;
+                              kwargs...)
+    D = DiagonalTensorMap(d, args...; kwargs...)
+    project_D = ProjectTo(D)
+    function DiagonalTensorMap_pullback(Δt)
+        ∂d = project_D(unthunk(Δt)).data
+        return NoTangent(), ∂d, ntuple(_ -> NoTangent(), length(args))...
+    end
```
lkdvos: This implementation slightly surprises me, I would have expected the projection to be based off `d`. In the end, these things probably boil down to the same thing?

```suggestion
    project_d = ProjectTo(d)
    function DiagonalTensorMap_pullback(Δt)
        ∂d = project_d(unthunk(Δt).data)
        return NoTangent(), ∂d, ntuple(_ -> NoTangent(), length(args))...
    end
    D = DiagonalTensorMap(d, args...; kwargs...)
```

ext/TensorKitChainRulesCoreExt/utility.jl
```diff
@@ -29,3 +29,15 @@ function (::ProjectTo{T1})(x::T2) where {S,N1,N2,T1<:AbstractTensorMap{<:Any,S,N
     end
     return y
 end
+
+function (::ProjectTo{T1})(x::T2) where {S,NumType,StorType,
+                                         T1<:DiagonalTensorMap{NumType,S,StorType},
+                                         T2<:AbstractTensorMap{<:Any,S,1,1}}
+    T1 === T2 && return x
+    y = DiagonalTensorMap{NumType,S,StorType}(undef, space(x, 1))
```
lkdvos:
```suggestion
function (::ProjectTo{T1})(x::T2) where {T1<:DiagonalTensorMap,T2<:AbstractTensorMap}
    T1 === T2 && return x
    y = DiagonalTensorMap{scalartype(T1),spacetype(T1),storagetype(T1)}(undef, space(x, 1))
```
In general, I prefer to not use too many parametric types in the signature, and rather use the functions to obtain them. I feel like that's a bit more flexible, and also somewhat easier to read.

test/ad.jl
```diff
@@ -144,6 +147,20 @@ Vlist = ((ℂ^2, (ℂ^3)', ℂ^3, ℂ^2, (ℂ^2)'),
...
     end
 
+    @timedtestset "Basic utility (DiagonalTensor)" begin
+        for NumType in [Float64, ComplexF64]
+            for v in V
+                T1 = DiagonalTensorMap(randn(NumType, dim(v)), v)
+                T2 = TensorMap(T1)
+
+                P1 = ProjectTo(T1)
+                @test P1(T2) == T1
```
lkdvos: It might also be a good idea to include a test that has mixed scalartypes

lkdvos (commented):
Thank you for taking the time to add this, this is really appreciated!
I left a few comments in the code, but overall looks great!

ext/TensorKitChainRulesCoreExt/constructors.jl
ebelnikola: It seems to me this is not the same. The issue here, is that sometimes \Delta t may be of some non-diagonal type. I expect that in this situation your version will return incorrect tangent (with a lot of zeros from off-diagonal parts of \Delta t).  Am I missing something?

lkdvos: Ah, good point, I think I missed this.
Somehow, I was expecting the input to the pullback to always be a `DiagonalTensorMap`, since this really is a constructor, and if not, there's probably a projector missing in some other rrule...
I also found [some comments in the rrules for Diagonal](https://github.com/JuliaDiff/ChainRules.jl/blob/02e21bacd8cb0aec3e80678ec2d0e1e1a964e9e3/src/rulesets/LinearAlgebra/structured.jl#L52) where a similar discussion is taking place.
Long story short though, it seems like their solution is more similar to yours, so that's definitely okay for me.

lkdvos:
I'll have a look tomorrow to help you out, the interplay with the symmetries is somewhat expected.
Did you have a specific use-case for this by the way?
The easiest way to circumvent this is to not define the `rrule` for this constructor, but instead define it for a conversion between a non-symmetric `Diagonal` and a symmetric `DiagonalTensorMap`, just like the `TensorMap` implementation. (Note that that one is also not defined in terms of the internal data representation, but instead with the general array). For abelian symmetries there should not be any difference, but for non-abelian ones there is, and then it becomes slightly confusing to implement this correctly.

ebelnikola:
Yes, I have a use-case. I needed to differentiate through a tensor RG algorithm, and this was the last missing detail. For me, everything works perfectly already with what is provided in this PR. However, I also wanted to contribute something useful.

I believe the problem lies somewhere in the tests, not in the implementation. I checked the pullback result by hand, and it is correct. I noticed that `test_rrule(DiagonalTensorMap, T.data, T.domain)` passes if one removes the `sqrtdim(c)` factor and its inverse from the `to_vec` function in TensorKitFiniteDifferencesExt.jl. However, this doesn't make a lot of sense, as I understand you wanted to have isometry there.

As for the `Diagonal` implementation: The problem here is that there's no constructor for `DiagonalTensorMap` that takes `Diagonal`. Also, Zygote asked me for an `rrule` for this method.

lkdvos:
I didn't find time today, but it's still on my TODO.
The problem is indeed the tests, in the sense that while computing derivatives, there is a bit of a choice of metric: you can either consider a euclidean metric on the parameters, or use the metric from the actual tensors. Usually the latter makes more sense, since it makes it such that working with symmetric tensors is equivalent to working without the symmetry, but here the rrule implements the former, because the constructor is working explicitly with the stored data. (Which I think is in fact correct)
However, the FiniteDifferences support is set up in such a way to work with the latter, hence the discrepancy between the two.

This is more or less why I asked in what context you are encountering this: if you really have a diagonal array that you wish to convert into a tensormap, to then work with tensormaps, we should probably define the appropriate converters for that, which include the correct scalings. If not, I would be interested in the stacktrace, or a MWE, such that I might be able to spot where this is coming from.

Maybe we should also consider changing/expanding on the definitions of `diag` and `diagm` [here](https://github.com/Jutho/TensorKit.jl/blob/660bdf782515decbb4fc55da98d76b34688e28e7/src/tensors/linalg.jl#L168), since that could be what you need (and should already be AD-compatible)?

ext/TensorKitChainRulesCoreExt/utility.jl
```diff
@@ -29,3 +29,13 @@ function (::ProjectTo{T1})(x::T2) where {S,N1,N2,T1<:AbstractTensorMap{<:Any,S,N
     end
     return y
 end
+
+function (::ProjectTo{T1})(x::T2) where {T1<:DiagonalTensorMap,T2<:AbstractTensorMap}
+    T1 === T2 && return x
+    y = DiagonalTensorMap{scalartype(T1),spacetype(T1),storagetype(T1)}(undef, space(x, 1))
```
Jutho: Can we make this into
```julia
function (::ProjectTo{DiagonalTensorMap{T,S,A}})(x::AbstractTensorMap) where {T,S,A}
    x isa DiagonalTensorMap{T,S,A} && return x
    V = space(x, 1)
    space(x) == (V ← V) || throw(SpaceMismatch())
    y = DiagonalTensorMap{T,S,A}(undef, V)
    for (c, b) in blocks(y)
         p = ProjectTo(b)
         b .= p(block(x, c))
     end
     return y
 end
```
I think that reads a bit more pleasantly and has a check on the space of `x` being correct.

test/ad.jl
```diff
@@ -15,6 +15,10 @@ end
 function ChainRulesTestUtils.rand_tangent(rng::AbstractRNG, x::AbstractTensorMap)
     return randn!(similar(x))
 end
+function ChainRulesTestUtils.rand_tangent(rng::AbstractRNG, x::DiagonalTensorMap)
+    S = x.domain
+    return DiagonalTensorMap(randn(eltype(x), sum(values(S.dims))), S)
```
Jutho: I would prefer to not use `S` for a space but rather use `V`. Also, there is a function `reduceddim(V)` for `sum(values(V.dims))` (as the latter uses private implementation details).

ext/TensorKitChainRulesCoreExt/constructors.jl
Jutho: I am also undecided between the two options. I agree that in most cases, e.g. tensor contractions, there should already have been a `ProjectTo(D)` being called on the adjoint variable `Δt` before it enters this `rrule`, and therefore not explicitly having `project_d(Δt)` in this rule might serve as a way to find missing projectors elsewhere.

On the other hand, from a user perspective, I can also see the advantage of just having this in `project_D` in here for safety. 

While in principle that is independent of still having to call `project_d = ProjectTo(d)` on the output `∂d`, it is probably true that `project_d` will not be doing anything (will act as the identity) if we already have projected `Δt` to a diagonal tensor with the right scalar type and storage type.

ebelnikola:
> This is more or less why I asked in what context you are encountering this: if you really have a diagonal array that you wish to convert into a tensormap, to then work with tensormaps, we should probably define the appropriate converters for that, which include the correct scalings. If not, I would be interested in the stacktrace, or a MWE, such that I might be able to spot where this is coming from.

Hmmm, in fact, now I cannot reproduce the problem I had. 

Here is an example of the code where I encountered a problem initially:
```
using TensorKit
using Zygote
data=rand(8)
T=DiagonalTensorMap(data, Z2Space(4,4))
f(S)=tr(sqrt(S))
gradient(f,T)
```
This returns:
```
ERROR: Need an adjoint for constructor DiagonalTensorMap{Float64, GradedSpace{Z2Irrep, Tuple{Int64, Int64}}, Vector{Float64}}. Gradient is of type TensorMap{Float64, GradedSpace{Z2Irrep, Tuple{Int64, Int64}}, 1, 1, Vector{Float64}}
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] (::Zygote.Jnew{DiagonalTensorMap{…}, Nothing, false})(Δ::TensorMap{Float64, GradedSpace{…}, 1, 1, Vector{…}})
    @ Zygote ~/.julia/packages/Zygote/D9opX/src/lib/lib.jl:334
  [3] (::Zygote.var"#2229#back#333"{Zygote.Jnew{…}})(Δ::TensorMap{Float64, GradedSpace{…}, 1, 1, Vector{…}})
    @ Zygote ~/.julia/packages/ZygoteRules/M4xmc/src/adjoint.jl:72
  [4] DiagonalTensorMap
    @ ~/.julia/dev/TensorKit/src/tensors/diagonal.jl:20 [inlined]
  [5] (::Zygote.Pullback{Tuple{…}, Any})(Δ::TensorMap{Float64, GradedSpace{…}, 1, 1, Vector{…}})
    @ Zygote ~/.julia/packages/Zygote/D9opX/src/compiler/interface2.jl:0
  [6] DiagonalTensorMap
    @ ~/.julia/dev/TensorKit/src/tensors/diagonal.jl:48 [inlined]
  [7] (::Zygote.Pullback{Tuple{…}, Any})(Δ::TensorMap{Float64, GradedSpace{…}, 1, 1, Vector{…}})
    @ Zygote ~/.julia/packages/Zygote/D9opX/src/compiler/interface2.jl:0
  [8] DiagonalTensorMap
    @ ~/.julia/dev/TensorKit/src/tensors/diagonal.jl:52 [inlined]
  [9] sqrt
    @ ~/.julia/dev/TensorKit/src/tensors/diagonal.jl:311 [inlined]
 [10] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::TensorMap{Float64, GradedSpace{…}, 1, 1, Vector{…}})
    @ Zygote ~/.julia/packages/Zygote/D9opX/src/compiler/interface2.jl:0
 [11] f
    @ ~/Codes/TensorRG/tmp2_local.jl:21 [inlined]
 [12] (::Zygote.var"#88#89"{Zygote.Pullback{Tuple{…}, Tuple{…}}})(Δ::Float64)
    @ Zygote ~/.julia/packages/Zygote/D9opX/src/compiler/interface.jl:97
 [13] gradient(f::Function, args::DiagonalTensorMap{Float64, GradedSpace{Z2Irrep, Tuple{Int64, Int64}}, Vector{Float64}})
    @ Zygote ~/.julia/packages/Zygote/D9opX/src/compiler/interface.jl:154
 [14] top-level scope
    @ ~/Codes/TensorRG/tmp2_local.jl:23
```
To fix this, I defined an `rrule` for `sqrt`. A few days ago, even after this, Zygote asked for the constructor's adjoint. However, when I am repeating this now, everything works fine without it. So yes, it appears now that there was no particular global problem solved by this PR except that `ProjectTo` had unexpected behavior before.

lkdvos:
Ah, I see now. The original problem comes from our specializations of matrix functions like `sqrt` for `DiagonalTensorMap`, which unpack the data, apply the function, and then repack it. The repacking step indeed requires a `rrule`, since that is the constructor of `DiagonalTensorMap`, and it should not include the dimensions, as implemented right here.

The biggest annoyance is using ChainRulesTestUtils.jl, since going from tensors to vectors in that context is understood as "desymmetrizing", ie weighing by quantum dimensions. This would be resolved if we would test `sqrt` instead, since there the unpacking-repacking are combined and would cancel each other, again making everything consistent.

As a separate note, the unpacking step might actually also require such a `rrule`, since accessing fields yields strange tangent types, which we are avoiding everywhere since our `rrules` will not be able to handle them.

Finally, we could also define the rrules for the matrix functions directly, but again ChainRulesTestUtils is being annoying: `sqrt` does weird things:
```julia
julia> sqrt(rand(3,3))
3×3 Matrix{ComplexF64}:
 0.251394+0.530824im  0.644466-0.46924im   0.245255+0.000886708im
  0.28562+0.198812im  0.729492-0.175747im  0.278139+0.000332103im
 0.652492-1.06654im   0.235591+0.942803im  0.368666-0.00178158im

julia> sqrt(rand(3,3))
3×3 Matrix{Float64}:
  0.976625  0.341331  -0.00378496
 -0.117352  0.820173   0.654396
  0.520547  0.424884   0.495172
julia> sqrt(Diagonal(randn(3)))
3×3 Diagonal{Float64, Vector{Float64}}:
 0.547541   ⋅         ⋅
  ⋅        0.484821   ⋅
  ⋅         ⋅        0.111613

julia> sqrt(Diagonal(randn(3)))
ERROR: DomainError with -0.31158918553265935:
```
And since the finite differences tests will generate jacobians for the number of real parameters, this yields jacobians with weird sizes.

I'm honestly not sure what the best approach would be

lkdvos:
In principle, this should now no longer require the `rrule` for the `DiagonalTensorMap` constructor. I'm okay with leaving that in, since I am pretty sure it is correct, but I don't think battling ChainRulesTestUtils.jl to make that behave is worth it. 

The comment about accessing the fields of tensors still remains, in the sense that we don't really support AD for that, but at least now these functions that access them have custom rules defined.

For `sqrt(::DiagonalTensorMap)`, I think it would also be possible to use a forwards rule instead, but I dont really know what the performance implications are.

lkdvos:
It seems like the failing nightly tests are due to [this commit](https://github.com/JuliaLang/LinearAlgebra.jl/pull/1156), where now additionally everything has become even more problematic:

v1:
```julia
julia> d = Diagonal([1., -1.])
2×2 Diagonal{Float64, Vector{Float64}}:
 1.0    ⋅
  ⋅   -1.0

julia> sqrt(d)
ERROR: DomainError with -1.0:
sqrt was called with a negative real argument but will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).

julia> sqrt(Array(d))
2×2 Matrix{ComplexF64}:
 1.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+1.0im

julia> sqrt(Array(d) + [0. 0.1; 0. 0.]) # array but not strictly diagonal
2×2 Matrix{ComplexF64}:
 1.0+0.0im  0.05-0.05im
 0.0+0.0im   0.0+1.0im
```

nightly:
```julia
julia> using LinearAlgebra

julia> d = Diagonal([1., -1.])
2×2 Diagonal{Float64, Vector{Float64}}:
 1.0    ⋅
  ⋅   -1.0

julia> sqrt(d)
ERROR: DomainError with -1.0:
sqrt was called with a negative real argument but will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).

julia> sqrt(Array(d))
ERROR: DomainError with -1.0:
sqrt was called with a negative real argument but will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).

julia> sqrt(Array(d) + [0. 0.1; 0. 0.]) # array but not strictly diagonal
2×2 Matrix{ComplexF64}:
 1.0+0.0im  0.05-0.05im
 0.0+0.0im   0.0+1.0im
```

lkdvos:
(https://discourse.julialang.org/t/type-instability-and-or-behavior-instability-of-sqrt-on-nightly/125428)

test/ad.jl
```diff
@@ -184,6 +229,27 @@ Vlist = ((ℂ^2, (ℂ^3)', ℂ^3, ℂ^2, (ℂ^2)'),
...
 
+    @timedtestset "Matrix functions ($T)" for T in (Float64, ComplexF64)
+        for f in (sqrt, exp)
+            check_inferred = false # !(T <: Real) # not type-stable for real functions
+            t1 = randn(T, V[1] ← V[1])
+            t2 = randn(T, V[2] ← V[2])
+            d = DiagonalTensorMap{T}(undef, V[1])
+            randn!(d.data)
+            if T <: Real
+                d.data .= abs.(d.data)
```
Jutho: I assume this is only necessary for `f === sqrt`? Alternatively, you could write
```
logdata = randn(T, rdim(V[1])
d = DiagonalTensorMap(exp.(logdata), V[1])
```
This will create a purely positive `DiagonalTensorMap` in the real case.

test/runtests.jl
```diff
@@ -60,9 +60,13 @@ include("spaces.jl")
 include("tensors.jl")
 include("diagonal.jl")
 include("planar.jl")
-if !(Sys.isapple()) # TODO: remove once we know why this is so slow on macOS
-    include("ad.jl")
+# TODO: remove once we know AD is slow on macOS CI
```
Jutho, line 63: This is a very strange construct. Why not
```julia
if !(Sys.isapple() && get(ENV, "CI", false))
    include("ad.jl")
end
```

lkdvos: It's mostly taken from [here](https://github.com/Jutho/TensorKit.jl/blob/33f10bc8ee826540b9113654c96ebb34800d27df/test/tensors.jl#L58).
I think the biggest problem is that the environment variables are strings, so they really become `"true"` and `"false"`, which makes for an annoying interface.
I don't think the construction above works, because `"true"` would not evaluate as a bool...

test/ad.jl
lkdvos: Just realized that there literally is a `randexp` function to do that in one go... Thanks for the suggestion, I'll change that.

Jutho: I don't think `randexp` works with complex numbers.

test/runtests.jl
Jutho: ok, `parse(Bool, get(ENV, "CI", "false"))`

test/ad.jl
lkdvos: But for complex numbers I don't need them to be positive, so I can just use `randn!`. I can definitely change it if you want, anything goes I guess?

test/runtests.jl
lkdvos: Sorry I missed this, I tried `get(ENV, "CI", "false") == "true"`. Does that work?

lkdvos:
Seems like this finally worked. Ready to go for me!

Jutho:
Looks good to me as well. I will merge tomorrow morning if the tests have successfully completed. Are we good to tag a patch update?

Issue #211, opened 2025-01-31
Providing a consistent IO method for all kinds of tensors


Yue-Zhengyuan:
It would be great if TensorKit provides out-of-box functions to save all kinds of tensors to disk and load it back (just like PyTorch's `save` and `load`). The issue I'm currently running into is the IO of a `DiagonalTensorMap`. For ordinary `TensorMap`s, I can use the following functions for IO (which, at least for ordinary `TensorMap`s, actually works for both 0.12.7 and newer versions of TensorKit):

```julia
using TensorKit
using JLD2

function save_tensor(filename::AbstractString, t::AbstractTensorMap)
    # the resulting Dict is "dense", the same as an ordinary TensorMap
    return JLD2.save_object(filename, convert(Dict, t))
end

function load_tensor(filename::AbstractString)
    # conversion back to DiagonalTensorMap is not implemented
    return convert(TensorMap, JLD2.load_object(filename))
end
```

In addition, there is not a convenient conversion from a diagonal `TensorMap` back to `DiagonalTensorMap`. Then I cannot easily load a saved tensor back as a `DiagonalTensorMap`. I guess there is a similar problem for some other special kind of tensors.

Yue-Zhengyuan:
Currently the converted `Dict` from a tensor is like follows:

```julia
t = randn(ℂ^2 ← ℂ^2)
obj = convert(Dict, t)
```

Output:

```
Dict{Symbol, Any} with 3 entries:
  :codomain => "ProductSpace(ℂ^2)"
  :domain   => "ProductSpace(ℂ^2)"
  :data     => Dict{String, Any}("Trivial()"=>[-0.995765 -0.126797; -1.02334 -1…
```

I suggest to add one more entry `:kind` to store the tensor kind (`TensorMap`, `DiagonalTensorMap`, etc.), and the data is the original data without being converted first to the data of an ordinary tensor. For compatibility, when `:kind` is missing from old files, just default it to `TensorMap`. It remains to implement `TensorKit.convert` from the Dict back to each tensor kind.

Jutho:
[...]

PR #212, opened 2025-02-03
Add DiagonalTensorMap constructors and converters

Yue-Zhengyuan:
Changes:

- In README guide of transferring data from old to new TensorKit, ~the old data is now specified by `blocks` instead of `fusiontrees`~.
- Added the conversion from a TensorMap to a DiagonalTensorMap. It first checks if the input is indeed diagonal.

### DIFF
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -91,6 +91,9 @@ end
 function Base.convert(D::Type{<:DiagonalTensorMap}, d::DiagonalTensorMap)
     return DiagonalTensorMap(convert(storagetype(D), d.data), d.domain)
 end
+function Base.convert(::Type{DiagonalTensorMap}, d::Dict{Symbol,Any})
+    return convert(DiagonalTensorMap, convert(TensorMap, d))
+end
 
 # Complex, real and imaginary parts
 #-----------------------------------
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -52,6 +52,15 @@ function DiagonalTensorMap(data::DenseVector{T}, V::IndexSpace) where {T}
     return DiagonalTensorMap{T}(data, V)
 end
 
+function DiagonalTensorMap(t::AbstractTensorMap)
+    isa(t, DiagonalTensorMap) && return t
+    @assert domain(t) == codomain(t) "Domain and codomain of the input tensor are different."
+    @assert numin(t) == numout(t) == 1 "Domain and codomain of the input tensor are not an IndexSpace."
+    @assert all(Diagonal(b) == b for (k, b) in blocks(t)) "Input tensor is not diagonal."
+    data = vcat((LinearAlgebra.diag(b) for (k, b) in blocks(t))...)
+    return DiagonalTensorMap(data, space(t, 1))
+end
+
 # TODO: more constructors needed?
 
 # Special case adjoint:
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -54,8 +54,10 @@ end
 
 function DiagonalTensorMap(t::AbstractTensorMap{T,S,1,1}) where {T,S}
     isa(t, DiagonalTensorMap) && return t
-    @assert domain(t) == codomain(t) "Domain and codomain of the input tensor are different."
-    @assert all(Diagonal(b) == b for (k, b) in blocks(t)) "Input tensor is not diagonal."
+    domain(t) == codomain(t) ||
+        throw(SpaceMismatch("DiagonalTensorMap requires equal domain and codomain"))
+    all(LinearAlgebra.isdiag ∘ last, blocks(t)) ||
+        throw(ArgumentError("DiagonalTensorMap requires input tensor that is diagonal"))
     data = vcat((LinearAlgebra.diag(b) for (k, b) in blocks(t))...)
     return DiagonalTensorMap(data, space(t, 1))
 end
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -87,13 +87,10 @@ end
 TensorMap(d::DiagonalTensorMap) = copy!(similar(d), d)
 Base.convert(::Type{TensorMap}, d::DiagonalTensorMap) = TensorMap(d)
 
-function Base.convert(::Type{DiagonalTensorMap{T,S,A}},
-                      d::DiagonalTensorMap{T,S,A}) where {T,S,A}
-    return d
-end
 function Base.convert(D::Type{<:DiagonalTensorMap}, d::DiagonalTensorMap)
-    return DiagonalTensorMap(convert(storagetype(D), d.data), d.domain)
+    return (d isa D) ? d : DiagonalTensorMap(convert(storagetype(D), d.data), d.domain)
 end
+Base.convert(::Type{DiagonalTensorMap}, t::DiagonalTensorMap) = t
 function Base.convert(::Type{DiagonalTensorMap}, t::AbstractTensorMap)
     LinearAlgebra.isdiag(t) ||
         throw(ArgumentError("DiagonalTensorMap requires input tensor that is diagonal"))
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -95,7 +95,7 @@ function Base.convert(D::Type{<:DiagonalTensorMap}, d::DiagonalTensorMap)
     return DiagonalTensorMap(convert(storagetype(D), d.data), d.domain)
 end
 function Base.convert(::Type{DiagonalTensorMap}, t::AbstractTensorMap)
-    all(LinearAlgebra.isdiag ∘ last, blocks(t)) ||
+    LinearAlgebra.isdiag(t) ||
         throw(ArgumentError("DiagonalTensorMap requires input tensor that is diagonal"))
     return DiagonalTensorMap(t)
 end
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -63,7 +63,7 @@ function DiagonalTensorMap(t::AbstractTensorMap{T,S,1,1}) where {T,S}
         # TODO: rewrite in terms of `diagview` from MatrixAlgebraKit.jl
         copy!(b.diag, view(bt, LinearAlgebra.diagind(bt)))
     end
-    return t
+    return d
 end
 
 # TODO: more constructors needed?
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -56,10 +56,14 @@ function DiagonalTensorMap(t::AbstractTensorMap{T,S,1,1}) where {T,S}
     isa(t, DiagonalTensorMap) && return t
     domain(t) == codomain(t) ||
         throw(SpaceMismatch("DiagonalTensorMap requires equal domain and codomain"))
-    all(LinearAlgebra.isdiag ∘ last, blocks(t)) ||
-        throw(ArgumentError("DiagonalTensorMap requires input tensor that is diagonal"))
-    data = vcat((LinearAlgebra.diag(b) for (k, b) in blocks(t))...)
-    return DiagonalTensorMap(data, space(t, 1))
+    A = storagetype(t)
+    d = DiagonalTensorMap{T,S,A}(undef, space(t, 1))
+    for (c, b) in blocks(d)
+        bt = block(t, c)
+        # TODO: rewrite in terms of `diagview` from MatrixAlgebraKit.jl
+        copy!(b.diag, view(bt, LinearAlgebra.diagind(bt)))
+    end
+    return t
 end
 
 # TODO: more constructors needed?
@@ -82,7 +86,6 @@ function Base.copy!(t::AbstractTensorMap, d::DiagonalTensorMap)
 end
 TensorMap(d::DiagonalTensorMap) = copy!(similar(d), d)
 Base.convert(::Type{TensorMap}, d::DiagonalTensorMap) = TensorMap(d)
-Base.convert(::Type{DiagonalTensorMap}, t::AbstractTensorMap) = DiagonalTensorMap(t)
 
 function Base.convert(::Type{DiagonalTensorMap{T,S,A}},
                       d::DiagonalTensorMap{T,S,A}) where {T,S,A}
@@ -90,6 +93,11 @@ function Base.convert(::Type{DiagonalTensorMap{T,S,A}},
 end
 function Base.convert(D::Type{<:DiagonalTensorMap}, d::DiagonalTensorMap)
     return DiagonalTensorMap(convert(storagetype(D), d.data), d.domain)
+end
+function Base.convert(::Type{DiagonalTensorMap}, t::AbstractTensorMap)
+    all(LinearAlgebra.isdiag ∘ last, blocks(t)) ||
+        throw(ArgumentError("DiagonalTensorMap requires input tensor that is diagonal"))
+    return DiagonalTensorMap(t)
 end
 function Base.convert(::Type{DiagonalTensorMap}, d::Dict{Symbol,Any})
     return convert(DiagonalTensorMap, convert(TensorMap, d))
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -52,10 +52,9 @@ function DiagonalTensorMap(data::DenseVector{T}, V::IndexSpace) where {T}
     return DiagonalTensorMap{T}(data, V)
 end
 
-function DiagonalTensorMap(t::AbstractTensorMap)
+function DiagonalTensorMap(t::AbstractTensorMap{T,S,1,1}) where {T,S}
     isa(t, DiagonalTensorMap) && return t
     @assert domain(t) == codomain(t) "Domain and codomain of the input tensor are different."
-    @assert numin(t) == numout(t) == 1 "Domain and codomain of the input tensor are not an IndexSpace."
     @assert all(Diagonal(b) == b for (k, b) in blocks(t)) "Input tensor is not diagonal."
     data = vcat((LinearAlgebra.diag(b) for (k, b) in blocks(t))...)
     return DiagonalTensorMap(data, space(t, 1))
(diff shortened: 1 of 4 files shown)


lkdvos:
I don't think that this is fully correct. We really need the data to be associated to the fusiontrees, because their order within the blocks changed between versions. As a result, simply loading in the blocks would yield permuted data, which is definitely not what you want.

src/tensors/diagonal.jl
```diff
@@ -72,6 +80,7 @@ function Base.copy!(t::AbstractTensorMap, d::DiagonalTensorMap)
 end
 TensorMap(d::DiagonalTensorMap) = copy!(similar(d), d)
 Base.convert(::Type{TensorMap}, d::DiagonalTensorMap) = TensorMap(d)
+Base.convert(::Type{DiagonalTensorMap}, t::AbstractTensorMap) = DiagonalTensorMap(t)
```
lkdvos: I'm not entirely convinced by this converter. In principle, conversion needs to be lossless, ie you should be able to convert back and forth without losing data. This implementation actually really projects the data onto the diagonal, so if you were to convert back and forth you lose the off-diagonal information

src/tensors/tensor.jl
```diff
@@ -395,15 +395,9 @@ Base.copy(t::TensorMap) = typeof(t)(copy(t.data), t.space)
...
-    d[:domain] = repr(domain(t))
-    data = Dict{String,Any}()
-    for (c, b) in blocks(t)
-        data[repr(c)] = Array(b)
-    end
-    d[:data] = data
-    return d
+    return Dict(:codomain => repr(codomain(t)),
+                :domain => repr(domain(t)),
+                :data => Dict(repr(c) => Array(b) for (c, b) in blocks(t)))
```
lkdvos: I'm not sure if this actually matters, but this might change the data type of the data dict.

src/tensors/diagonal.jl
Yue-Zhengyuan: For tensors with off-diagonal elements, the assertions in the constructor will prevent it form being converted.

src/tensors/diagonal.jl
```diff
@@ -52,6 +52,14 @@ function DiagonalTensorMap(data::DenseVector{T}, V::IndexSpace) where {T}
     return DiagonalTensorMap{T}(data, V)
 end
 
+function DiagonalTensorMap(t::AbstractTensorMap{T,S,1,1}) where {T,S}
+    isa(t, DiagonalTensorMap) && return t
+    @assert domain(t) == codomain(t) "Domain and codomain of the input tensor are different."
+    @assert all(Diagonal(b) == b for (k, b) in blocks(t)) "Input tensor is not diagonal."
```
Jutho: I'd prefer somethign like: 

```julia
domain(t) == codomain(t) || throw(SpaceMismatch("DiagonalTensorMap requires equal domain and codomain"))
all(LinearAlgebra.isdiag ∘ last, blocks(t)) || throw(ArgumentError("DiagonalTensorMap requires input tensor that is diagonal"))
```

Jutho: Also, it seems like `LinearAlgebra.Diagonal(m::Matrix)` happily discards the off-diagonal data and just takes the diagonal of `m`. Should we follow this behaviour or not? If we do, then I the check for `isdiag` should move to `convert`, otherwise it is indeed implemented here.

src/tensors/tensor.jl
Jutho: Is there any benefit to this change? While being slightly shorter, I do like that the old version makes it very clear which specific `Dict` parameters the output type will have.

src/tensors/diagonal.jl
lkdvos: Yes, this is actually how it is supposed to be: a converter is lossless, and should work both ways, (since this might happen silently when assigning fields/entries of typed containers), but a constructor does not require this property. https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Conversion-vs.-Construction

src/tensors/tensor.jl
```diff
@@ -395,15 +395,9 @@ Base.copy(t::TensorMap) = typeof(t)(copy(t.data), t.space)
...
 end
 function Base.convert(::Type{TensorMap}, d::Dict{Symbol,Any})
```
Jutho: I wouldn't mind having an explicit method
```julia
Base.convert(::Type{DiagonalTensorMap), d::Dict{Symbol, Any}) = convert(DiagonalTensorMap, convert(TensorMap, d))
```

src/tensors/diagonal.jl
lkdvos: Sorry, I missed these assertions. See my comment there.

src/tensors/tensor.jl
Yue-Zhengyuan: I've added this method. But it also has the issue that whether `convert` (which is supposed to be lossless) is the suitable function as Lukas suggested.

src/tensors/tensor.jl
Yue-Zhengyuan: I just wanted to let Julia determine the type of the constructed Dict. I can revert to the old version to manually set the Dict type.

src/tensors/tensor.jl
Jutho: Can I make some edits to your PR? I finished grading exams 😄

Yue-Zhengyuan: Sure, no problem 😄

lkdvos:
So, summarizing my thoughts a bit here:

For the functions, I would say the best approach is:
- The `DiagonalTensorMap(::AbstractTensorMap)` constructor should take in arbitrary tensormaps, and simply "project onto the diagonal", ie discard off-diagonal elements. This is similar to how `LinearAlgebra.Diagonal` functions.
- The `convert(::Type{<:DiagonalTensorMap}, x::AbstractTensorMap)` should first check if `x` is actually diagonal, and then either throw or do the conversion. This way, this is lossless.
- The `convert(::Type{<:DiagonalTensorMap}, x::Dict)` can be defined in terms of `convert(DiagonalTensorMap, convert(TensorMap, x))` (or a more efficient direct path). This way, it will throw whenever you try loading in data that is not diagonal.

For the best practice on saving and loading states, I'm actually a bit torn. It seems to me that JLD2 has actually come quite a long way, and is quite good at supporting arbitrary datatypes. Additionally, after reading through the documentation a bit more, there are several features that could make our lives a lot easier:

- [Custom serialization](https://juliaio.github.io/JLD2.jl/dev/customserialization/) could be used to automatically convert everything to a "stable" datatype, which we could consider supporting on a more long-term basis. In particular, this could be a simple `Dict` wrapper like we had before.
- [Explicit type remapping](https://juliaio.github.io/JLD2.jl/dev/advanced/#Explicit-Type-Remapping) and [Upgrading old structures on load](https://juliaio.github.io/JLD2.jl/dev/advanced/#Upgrading-old-structures-on-load) could make changes in the future seamless for a user.

The combination of these two could mean that we simply define something like:
```julia
struct TensorData_v1
    # stable data format
end
```
and store tensors simply by using JLD2's automatic serialization. Then, since the custom serialization is stable, we can load in tensors from any version by implementing the correct deserializer.
(similarly for `DiagonalTensorData`, etc)

Additionally, if we would ever like to go to a different data format, we could just add
```julia
struct TensorData_v2 end
```
and remap the types on load.

src/tensors/diagonal.jl
```diff
@@ -80,6 +94,14 @@ end
 function Base.convert(D::Type{<:DiagonalTensorMap}, d::DiagonalTensorMap)
     return DiagonalTensorMap(convert(storagetype(D), d.data), d.domain)
 end
+function Base.convert(::Type{DiagonalTensorMap}, t::AbstractTensorMap)
+    all(LinearAlgebra.isdiag ∘ last, blocks(t)) ||
```
lkdvos:
```suggestion
    LinearAlgebra.isdiag(t) ||
```
This function is actually defined, I implemented these a while ago: https://github.com/Jutho/TensorKit.jl/blob/450ff9bf21176859123ecfa39b3e769e571a4a9d/src/tensors/linalg.jl#L168-L176
They might require some revisions, since this predates `DiagonalTensorMap` (would also fix some todos)

Jutho:
I think this is good to go from my end; the failure on nightly seems unrelated. @Yue-Zhengyuan , anything that you still like to see changed?

lkdvos (approved):

Yue-Zhengyuan:
I'm OK to fix in another PR the transferring from tensors from old to new TensorKit.

Jutho:
I seem to have missed this, but what is wrong with the description of how to transfer tensors from TensorKit prior to v0.13 to newer versions as currently in the README? I've restored the README because the updated description you wrote was incorrect, as shown by the example in #211 . Is there anything else that you would like to see in this regard?

Yue-Zhengyuan:
- Currently JLD2 does not have the function `jldload`. I suggested to replace it with `load_object`. 
- The fusion trees structure in new TensorKit is defined differently from the old TensorKit, so the new TensorKit cannot reconstruct it from the old string representation. 

Example: first, run the following in environment with v0.12.7

```julia
using TensorKit
using JLD2

function save_tensor_tmp(filename::AbstractString, t::AbstractTensorMap)
    t_dict = Dict(
        :space => space(t), 
        :data => Dict((f₁, f₂) => t[f₁, f₂] for (f₁, f₂) in fusiontrees(t))
    )
    jldsave(filename; t_dict)
    return nothing
end

V1, V2 = ℂ^2, ℂ^3
a = TensorMap(randn, Float64, V1 ← V2);
println(fusiontrees(a)) # output: ((nothing, nothing),)
save_tensor_tmp("tmp.jld2", a)
```

Next, run the following in environment with v0.14.3

```julia
using TensorKit
using JLD2

function load_tensor_tmp(filename::AbstractString)
    t_dict = JLD2.load_object(filename)
    T = eltype(valtype(t_dict[:data]))
    t = TensorMap{T}(undef, t_dict[:space])
    for ((f₁, f₂), val) in t_dict[:data]
        t[f₁, f₂] .= val
    end
    return t
end

a =  load_tensor_tmp("tmp.jld2");
```

Error message:

```
MethodError: no method matching getindex(::TensorMap{Float64, ComplexSpace, 1, 1, Vector{Float64}}, ::Nothing, ::Nothing)
The function `getindex` exists, but no method is defined for this combination of argument types.
...
```

This is because in v0.14.3, the `fusiontrees` of a trivial tensor is defined as

```julia
a = randn(ℂ^2 ← ℂ^3)
fusiontrees(a)
#= output:
1-element Vector{Tuple{FusionTree{Trivial, 1, 0, 0}, FusionTree{Trivial, 1, 0, 0}}}:
 (FusionTree{Trivial}((Trivial(),), Trivial(), (false,), ()), FusionTree{Trivial}((Trivial(),), Trivial(), (false,), ()))
=#
```

There are also problems with other sector types. For example, for an SU(2) tensor, the error message is

```
┌ Warning: read type FusionTree has a different number of parameters from type FusionTree in workspace; reconstructing
└ @ JLD2 [/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/data/reconstructing_datatypes.jl:543](https://file+.vscode-resource.vscode-cdn.net/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/data/reconstructing_datatypes.jl:543)
┌ Warning: read type NTuple{N, T} where {N, T} has a different number of parameters from type NTuple{N, T} where {N, T} in workspace; reconstructing
└ @ JLD2 [/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/data/reconstructing_datatypes.jl:543](https://file+.vscode-resource.vscode-cdn.net/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/data/reconstructing_datatypes.jl:543)
┌ Warning: custom serialization of Dict{JLD2.ReconstructedStatic{Symbol("NTuple{N, T} where {N, T}{2,JLD2.ReconstructedStatic{Symbol(\"FusionTree{SU2Irrep,1,0,0,Nothing}\"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}"), (Symbol("1"), Symbol("2")), Tuple{JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}, JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}},StridedViews.StridedView{Float64, 2, Matrix{Float64}, typeof(identity)}} encountered, but the type does not exist in the workspace; the data will be read unconverted
└ @ JLD2 [/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/data/reconstructing_datatypes.jl:108](https://file+.vscode-resource.vscode-cdn.net/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/data/reconstructing_datatypes.jl:108)
┌ Warning: type Pair{JLD2.ReconstructedStatic{Symbol("NTuple{N, T} where {N, T}{2,JLD2.ReconstructedStatic{Symbol(\"FusionTree{SU2Irrep,1,0,0,Nothing}\"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}"), (Symbol("1"), Symbol("2")), Tuple{JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}, JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}},StridedViews.StridedView{Float64, 2, Matrix{Float64}, typeof(identity)}} does not exist in workspace; interpreting Array{Pair{JLD2.ReconstructedStatic{Symbol("NTuple{N, T} where {N, T}{2,JLD2.ReconstructedStatic{Symbol(\"FusionTree{SU2Irrep,1,0,0,Nothing}\"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}"), (Symbol("1"), Symbol("2")), Tuple{JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}, JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}},StridedViews.StridedView{Float64, 2, Matrix{Float64}, typeof(identity)}}} as Array{Any}
└ @ JLD2 [/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/datasets.jl:128](https://file+.vscode-resource.vscode-cdn.net/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/datasets.jl:128)
┌ Warning: some parameters could not be resolved for type Pair{JLD2.ReconstructedStatic{Symbol("NTuple{N, T} where {N, T}{2,JLD2.ReconstructedStatic{Symbol(\"FusionTree{SU2Irrep,1,0,0,Nothing}\"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}"), (Symbol("1"), Symbol("2")), Tuple{JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}, JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}},StridedViews.StridedView{Float64, 2, Matrix{Float64}, typeof(identity)}}; reading as Pair{JLD2.ReconstructedStatic{Symbol("NTuple{N, T} where {N, T}{2,JLD2.ReconstructedStatic{Symbol(\"FusionTree{SU2Irrep,1,0,0,Nothing}\"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}"), (Symbol("1"), Symbol("2")), Tuple{JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}, JLD2.ReconstructedStatic{Symbol("FusionTree{SU2Irrep,1,0,0,Nothing}"), (:uncoupled, :coupled, :isdual, :innerlines, :vertices), Tuple{Tuple{SU2Irrep}, SU2Irrep, Tuple{Bool}, Tuple{}, Tuple{}}}}}, StridedViews.StridedView{Float64, 2, Matrix{Float64}, typeof(identity)}}
└ @ JLD2 [/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/data/reconstructing_datatypes.jl:566](https://file+.vscode-resource.vscode-cdn.net/Users/zhengyuanyue/.julia/packages/JLD2/nPYlZ/src/data/reconstructing_datatypes.jl:566)
MethodError: no method matching valtype(::JLD2.SerializedDict)
The function `valtype` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  valtype(!Matched::Type{Union{}}, Any...)
   @ Base abstractarray.jl:195
  valtype(!Matched::Type{<:AbstractDict{<:Any, V}}) where V
   @ Base abstractdict.jl:322
  valtype(!Matched::Type{<:AbstractArray})
   @ Base abstractarray.jl:213
  ...
```

Jutho:
Ok, I surely had missed/forgotten that part of the issue. My apologies and thanks for laying it out so clearly. It would indeed be good to fix this in a separate PR. I will merge this one and think about possible solutions.

PR #214, opened 2025-02-05
Add rrule for matrix exponential

### LINKED ISSUE #213: rrule for exponential of TensorMap
Currently there is no rrule for the exponential function. I have tried
`function ChainRulesCore.rrule(::typeof(exp), A::AbstractTensorMap)
    expA = exp(A)
    exp_pullback(Δexp) = (NoTangent(), expA * Δexp)
    return expA, exp_pullback
end`
but this gives me an error when checking it with `test_rrule(exp, randn(Float64, ComplexSpace(3), ComplexSpace(3)))`.

lkdvos:
This uses the pullback of the blocks to promote it to a pullback of the entire tensor. Sadly, the implementation of ChainRules.jl is not type stable since it is checking at runtime if the matrices are Hermitian, so the `rrule` is also not type stable (the pullback itself is type stable however).

Fixes #213

### DIFF
--- a/ext/TensorKitChainRulesCoreExt/linalg.jl
+++ b/ext/TensorKitChainRulesCoreExt/linalg.jl
@@ -106,3 +106,24 @@ function ChainRulesCore.rrule(::typeof(imag), a::AbstractTensorMap)
     end
     return a_imag, imag_pullback
 end
+
+function ChainRulesCore.rrule(cfg::RuleConfig, ::typeof(exp), A::AbstractTensorMap)
+    domain(A) == codomain(A) ||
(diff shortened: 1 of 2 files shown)


Jutho:
That's good enough for now; I will try to have a better version in MatrixAlgebraKit.jl at some point.

lkdvos:
I think the algorithm in Chainrules is probably sound, but it would definitely be good to make it type stable.

Jutho:
I have a more general version of this written out here:
https://github.com/Jutho/CMPSKit.jl/blob/master/src/common/exp_lazy.jl

I will integrate it into MatrixAlgebraKit.jl

PR #220, opened 2025-02-12
Add `rank` and `cond`

### LINKED ISSUE #219: MethodError in svd_pullback!
Hi Jutho and Lukas,

I am using AD in my code that involves computing the Renyi entropy from the singular values of an MPS bipartition. Due to U1 symmetry, it is possible that there are zero blocks in the two-site tensor before decomposition. In this case, the gradient of the SVD fails [here](https://github.com/Jutho/TensorKit.jl/blob/511fa902e69bfbe747cea898503b20a086e8bfdd/ext/TensorKitChainRulesCoreExt/factorizations.jl#L222).

Below you can see a MWE. If virtSpaceC is truncated, twoSiteTensor has zero blocks and the issue occurs. If I take the maximal space for virtSpaceC there are no zero blocks and it works. Locally, I have modified the code to be `r = sum(>=(tol), S)`, which seems to solve the problem.

```
function computeRenyiEntropy(twoSiteTensor::TensorMap)
    U, S, V = tsvd(twoSiteTensor, ((1, 2), (3, 4)))
    return 2 * log(tr(S))
end

function value_and_gradient(twoSiteTensor::TensorMap)
    fval = computeRenyiEntropy(twoSiteTensor)
    gval = Zygote.gradient(x -> computeRenyiEntropy(x), twoSiteTensor)[1]
    return fval, gval
end

# set nMax
nMax = 2

# initialize physical vector spaces
kL = -1
kR = +1
physSpaceL = U1Space([kL * nL => 1 for nL in 0:nMax])
physSpaceR = U1Space([kR * nR => 1 for nR in 0:nMax])

# initialize boundary virtual vector spaces
virtSpaceL = U1Space(0 => 1)
virtSpaceR = U1Space(0 => 1)

# set (truncated) central virtual space
virtSpaceC = U1Space(0 => 1, -1 => 1) # truncated space
# virtSpaceC = fuse(virtSpaceL, physSpaceL) # full space

# initialize MPS and twoSiteTensor
tensorA = randn(Float64, virtSpaceL ⊗ physSpaceL, virtSpaceC)
tensorB = randn(Float64, virtSpaceC ⊗ physSpaceR, virtSpaceR)
twoSiteTensor = permute(tensorA * permute(tensorB, ((1,), (2, 3))), ((1, 2), (3, 4)))
display(twoSiteTensor)

# test value and gradient function
fval, gval = value_and_gradient(twoSiteTensor)
println(fval)
println(gval)
```

lkdvos:
I added the utility `rank` and `cond` functions, since I thought this would help fix #219 . I overlooked that there `S` is already a vector, so that actually didn't help, but it might be useful anyways?

# To do
- [x] Add tests

### DIFF
--- a/ext/TensorKitChainRulesCoreExt/factorizations.jl
+++ b/ext/TensorKitChainRulesCoreExt/factorizations.jl
@@ -219,7 +219,7 @@ function svd_pullback!(ΔA::AbstractMatrix, U::AbstractMatrix, S::AbstractVector
     Sp = view(S, 1:p)
 
     # rank
-    r = findlast(>=(tol), S)
+    r = count(>(tol), S)
 
     # compute antihermitian part of projection of ΔU and ΔV onto U and V
     # also already subtract this projection from ΔU and ΔV
--- a/src/TensorKit.jl
+++ b/src/TensorKit.jl
@@ -72,7 +72,7 @@ export mul!, lmul!, rmul!, adjoint!, pinv, axpy!, axpby!
 export leftorth, rightorth, leftnull, rightnull,
        leftorth!, rightorth!, leftnull!, rightnull!,
        tsvd!, tsvd, eigen, eigen!, eig, eig!, eigh, eigh!, exp, exp!,
-       isposdef, isposdef!, ishermitian, sylvester
+       isposdef, isposdef!, ishermitian, sylvester, rank, cond
 export braid, braid!, permute, permute!, transpose, transpose!, twist, twist!, repartition,
        repartition!
 export catdomain, catcodomain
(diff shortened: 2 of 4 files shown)


Jutho:
Looks good to me. Is the "add tests" an actual todo?

lkdvos:
What is an "actual" to do? I would definitely say "should do"...

Jutho:
I am making some changes (the condition number of the map on an empty space should be one, not zero. Condition numbers lie in the range [1, +inf]). But now for the difficult question: should rank count the degeneracy, i.e. is it `sum(dim(c) * rank(b))` or simply `sum(rank(b))` ?

lkdvos:
With regards to the condition number, I took this from LinearAlgebra, see also [this discussion](https://github.com/JuliaLang/LinearAlgebra.jl/pull/1169) and the related [MATLAB discussion](https://www.mathworks.com/matlabcentral/answers/1744315-condition-number-of-empty-matrix). I have no opinion about this I think, whatever you feel is most consistent is good to me.

For the rank, I guess if the matrix is full rank, you would expect `rank(A) = dim(domain(A))`, so whatever is consistent with that?

Jutho:
Ok I see regarding condition number. Not my favorite pick, but I can live with it for consistency.

For the rank, that choice would include quantum dimensions.

Jutho:
@lkdvos , if the tests pass, this is good to go for me.

lkdvos:
Nightly test failures are unrelated, so I'm merging this now.

PR #224, opened 2025-02-20
insertunit and removeunit using Val

Jutho:


lkdvos:
Looks good to me!

I think the documentation is failing because now the docstrings are only attached to the `Val` functions. Perhaps we could simply have one centralized docstring somewhere, mentioning the 3 different signatures?

Jutho:
I am looking at fixing the docs (and other small issues with it).

PR #228, opened 2025-03-08
Replace `isequal` with `==` for `FusionTree`

lkdvos:
I ran into an issue where `==(::FusionTree, ::FusionTree)` did not correctly evaluate.
Since it is not defined, it falls back to `===`, which has not caused any problems for us because all our sectors are `isbits`, however this fails when this is not the case.
(In this case, we (@Yue-Zhengyuan and I) are using a `Vector{Int}` to store the partition of `N` for an `S3Irrep`).
In principle, we can work around this and not have vectors, or locally change it, but this seems like a bug that should be solved.

From the documentation of Base, it seems like we should simply define `==` instead of `isequal` for fusiontrees, because we are not dealing with floating point values. `isequal` should then fall back to `==`. Am I missing some performance implications here, or is this simply an oversight?

### DIFF
--- a/src/fusiontrees/fusiontrees.jl
+++ b/src/fusiontrees/fusiontrees.jl
@@ -114,7 +114,7 @@ function Base.hash(f::FusionTree{I}, h::UInt) where {I}
     end
     return h
 end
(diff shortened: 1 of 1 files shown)


Jutho:
Oversight or incorrect reading of the manual I guess, if there was no reason to define `isequal`.

PR #230, opened 2025-03-16
[WIP] MatrixAlgebraKit decompositions

lkdvos:
This is some preliminary code to start geting a feeling for how we could go and implement the `MatrixAlgebraKit` functions for tensors. 

### Looping over blocks

The first thing I did was try and generalize the "looping over blocks" concept, with an eye towards parallelizing that in the near future. I can come up with several different designs, which are all more or less equivalent, but look slightly different. The one I implemented here is centered around `foreach`, with a wink towards `OhMyThreads.tforeach` to include schedulers:

```julia
foreachblock(f, t, ts...)
# equivalent to
for (c, b) in blocks(t)
    f(c, (b, block.(ts, Ref(c))...))
end
```

There are several design questions here:
1. Is it fair to loop over the blockcharges of the first tensor? This can often be the case if that is an "output" tensor, but it might not cover all cases, maybe you want to loop over the union of all sectors, or the intersection?
2. Do we want `f(c, bs...)` or `f(c, bs)`?

Alternatives I can come up with are to generalize `blocks(t)` to `blocks(t1, t2, ...) -> (c => (b1, b2, ...))...`, but that does not encapsulate the threading options.

Should we also include a `mapblocks!(f, t, ts...)` and/or `mapreduceblocks(f, op, t, ts...)`?
If we also want `mapblocks(f, t, ts...)`, should the output be based on all arguments, or just the first?

### Implementing decompositions

Then, in order to define the decompositions, there are again some choices to be made.
The first is whether or not we want to define the MatrixAlgebraKit functions themselves for `AbstractTensorMap`, or only define our wrappers that dispatch to these implementations. Here, I chose to go ahead and implement the functions directly.

Secondly, the current implementation preallocates the entire output first, and then applies the relevant functions blockwise. This is particularly useful for multithreading and maximally reusing memory, but might become a bit more involved for the cases where we do not know the correct sizes a priori.

As a sidenote, currently we are not very consistent with the arrows of the factorizations. For example, `leftorth(t) -> Q, R` might have different duality of the connecting space depending on the number of indices of `t`, which can be a bit annoying when working with fermions.

Thirdly, it would be convenient to have the `default_eig_alg` function also work in the type domain, so this can be defined for a tensor without having to instantiate a block.
Similarly or additionally, a `DefaultAlgorithm` type to indicate deferring the selection until later could be useful.

---

Let me know what you think about some of this, I'll try and add some more functionality this week if we start converging on some of these design questions :)


src/tensors/matrixalgebrakit.jl
```diff
@@ -0,0 +1,49 @@
+function MAK.copy_input(::typeof(MAK.eig_full), t::AbstractTensorMap)
+    return copy_oftype(t, factorisation_scalartype(MAK.eig_full!, t))
+end
+
+function factorisation_scalartype(::typeof(MAK.eig_full!), t::AbstractTensorMap)
+    T = complex(scalartype(t))
+    return promote_type(ComplexF32, typeof(zero(T) / sqrt(abs2(one(T)))))
```
lkdvos:
```suggestion
    return promote_type(Float32, typeof(zero(T) / sqrt(abs2(one(T)))))
```

src/tensors/factorizations.jl
```diff
@@ -222,10 +222,7 @@ matrices. See the corresponding documentation for more information.
 
 See also `eigen` and `eigh`.
 """
-function eig(t::AbstractTensorMap, p::Index2Tuple; kwargs...)
-    tcopy = permutedcopy_oftype(t, factorisation_scalartype(eig, t), p)
-    return eig!(tcopy; kwargs...)
-end
+eig(t::AbstractTensorMap, p::Index2Tuple; kwargs...) = MAK.eig_full(t; kwargs...)
```
Jutho: This seems to loose the `p` argument. In first instance, I would perhaps only route the `factorization!(t)` methods through `MAK`, and then gradually extend from there.

src/tensors/matrixalgebrakit.jl
```diff
@@ -0,0 +1,210 @@
...
+     space(U) == (codomain(t) ← V_cod)) ||
+        throw(ArgumentError("`svd_full!` requires unitary tensor U with same `scalartype`"))
+    (S isa AbstractTensorMap &&
+     scalartype(S) == real(scalartype(t)) &&
+     space(S) == (V_cod ← V_dom)) ||
+        throw(ArgumentError("`svd_full!` requires rectangular tensor S with real `scalartype`"))
+    (Vᴴ isa AbstractTensorMap &&
+     scalartype(Vᴴ) == scalartype(t) &&
+     space(Vᴴ) == (V_dom ← domain(t))) ||
+        throw(ArgumentError("`svd_full!` requires unitary tensor Vᴴ with same `scalartype`"))
```
Jutho: This sounds a bit strange, as if the input tensor already needs to be unitary. Maybe something like "`svd_full!(t, (U,S,Vᴴ))` requires requires `space(Vᴴ) == fuse(P)←P` with `P=domain(t)`". That is a bit different in structure than the errors in MatrixAlgebraKit, but maybe I should also change those. Probably the check for the scalar type should also be separated from the space / size check, and MatrixAlgebraKit.jl should have a macro like this:
```julia
macro check_eltype(x, y, f=:identity, g=:eltype)
    error_message = "unexpected scalar type: "
    error_message *= string(g) * "(" * string(x) * ") != "
    if f == identity
         error_message *= string(g) * "(" * string(y) * ")"
    else
         error_message *= string(f) * "(" * string(g) * "(" * string(y) * "))"
    end
    return :($g($x) == $f($g($y)) || throw(ArgumentError($error_message)))
end
```

The error message would then not contain the method contain the name of the method which causes the error, but that information should be clear from the stack trace. I don't think it is common in Base, e.g. checking `eigen(randn(3,4))`, the non-square error does not contain a reference to `eigen`.

Anyway, maybe it is good to have some kind of conventions for the error message structure, but yeah, more conventions and interface discussions 😄

src/spaces/cartesianspace.jl
```diff
@@ -56,4 +56,9 @@ flip(V::CartesianSpace) = V
 infimum(V₁::CartesianSpace, V₂::CartesianSpace) = CartesianSpace(min(V₁.d, V₂.d))
 supremum(V₁::CartesianSpace, V₂::CartesianSpace) = CartesianSpace(max(V₁.d, V₂.d))
 
+function Base.setdiff(V::CartesianSpace, W::CartesianSpace)
```
Jutho, line 59: I am not sure if I like `setdiff` for this, since vector spaces are ultimately sets and so could have a `setdiff`, which is very different from this operation. I assume you are trying to define is a complementary space `Wc` such that `W ⊕ Wc == V`. I am not sure if there is great notation for that.

Any complement is isomorphic to the quotient space $V / W$ (not currently implemented), but the quotient space is technically a space where the elements represent equivalence classes of vectors $v \in V$ under the equivalence relation $v_1 \sim v_2$ if $v_1 - v_2 \in W$. Hence, it is not very natural to represent the quotient space with the same type of objects, and I think it would also be a bit strange if the dual operation to `V = W ⊕ Wc` would be `Wc = V / W`.

Although this would be completely not standard and I have never seen this in the literature, maybe we could use `ominus` for this, i.e. `Wc = V ⊖ W` ?

src/spaces/vectorspaces.jl
```diff
@@ -395,3 +395,11 @@ have the same value.
...
     return supremum(supremum(V₁, V₂), V₃...)
 end
+
+"""
+    setdiff(V::ElementarySpace, W::ElementarySpace)
+
+Return the set difference of two elementary spaces, i.e. an instance `X::ElementarySpace`
+such that `V = W ⊕ X`.
+"""
+Base.setdiff(V₁::S, V₂::S) where {S<:ElementarySpace}
```
Jutho: I am not sure if I like `setdiff` for this, since vector spaces are ultimately sets and so could have a `setdiff`, which is very different from this operation. I assume you are trying to define is a complementary space `Wc` such that `W ⊕ Wc == V`. I am not sure if there is great notation for that.

Any complement is isomorphic to the quotient space $V / W$ (not currently implemented), but the quotient space is technically a space where the elements represent equivalence classes of vectors $v \in V$ under the equivalence relation $v_1 \sim v_2$ if $v_1 - v_2 \in W$. Hence, it is not very natural to represent the quotient space with the same type of objects, and I think it would also be a bit strange if the dual operation to `V = W ⊕ Wc` would be `Wc = V / W`.

Although this would be completely not standard and I have never seen this in the literature, maybe we could use `ominus` for this, i.e. `Wc = V ⊖ W` ?

src/tensors/diagonal.jl
```diff
@@ -37,6 +37,16 @@ storagetype(::Type{<:DiagonalTensorMap{T,S,A}}) where {T,S,A<:DenseVector{T}} =
...
 Construct a `DiagonalTensorMap` with uninitialized data.
 """
+function DiagonalTensorMap{T}(::UndefInitializer, V::TensorMapSpace) where {T}
+    (numin(V) == numout(V) == 1 && domain(V) == codomain(V)) ||
+        throw(SpaceMismatch("DiagonalTensorMap requires a space with equal domain and codomain and 2 indices"))
+    return DiagonalTensorMap{T}(undef, domain(V))
+end
+function DiagonalTensorMap{T}(::UndefInitializer, V::ProductSpace) where {T}
+    length(V) == 1 ||
+        throw(DimensionMismatch("length(V) = $(length(V)) is not compatible with the space $V"))
```
Jutho: That is a weird error message? Don't you just want to say that you only accept product spaces with a single elementary space?

src/spaces/vectorspaces.jl
lkdvos: I'm definitely happy with `ominus`, while not standard, this is easy to remember and suggestive of what it does

src/tensors/diagonal.jl
lkdvos: Yes, I have no clue what I was thinking here...

ext/TensorKitChainRulesCoreExt/factorizations.jl
```diff
@@ -23,12 +24,11 @@ function ChainRulesCore.rrule(::typeof(TensorKit.tsvd!), t::AbstractTensorMap;
...
         Δt = similar(t)
-        for (c, b) in blocks(Δt)
-            Uc, Σc, V⁺c = block(U, c), block(Σ, c), block(V⁺, c)
-            ΔUc, ΔΣc, ΔV⁺c = block(ΔU, c), block(ΔΣ, c), block(ΔV⁺, c)
-            Σdc = view(Σc, diagind(Σc))
-            ΔΣdc = (ΔΣc isa AbstractZero) ? ΔΣc : view(ΔΣc, diagind(ΔΣc))
-            svd_pullback!(b, Uc, Σdc, V⁺c, ΔUc, ΔΣdc, ΔV⁺c)
+        foreachblock(Δt) do (c, b)
+            USVᴴc = block(U, c), block(Σ, c), block(V⁺, c)
+            ΔUSVᴴc = block(ΔU, c), block(ΔΣ, c), block(ΔV⁺, c)
```
Jutho: Not sure if the formatter agrees, but I kinda like explicit parenthesis around tuples.

src/auxiliary/linalg.jl
```diff
@@ -84,6 +84,8 @@ end
 safesign(s::Real) = ifelse(s < zero(s), -one(s), +one(s))
 safesign(s::Complex) = ifelse(iszero(s), one(s), s / abs(s))
 
+isisometry(A::StridedMatrix; kwargs...) = isapprox(A' * A, LinearAlgebra.I, kwargs...)
+
```
Jutho: Ideally, this whole file is removed and replaced with things that are available in MatrixAlgebraKit. I am fine with this as a transitional approach, but I think we want `isisometry` in MatrixAlgebraKit.jl

src/spaces/gradedspace.jl
```diff
@@ -149,6 +149,12 @@ function ⊕(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I<:Sector}
     return typeof(V₁)(dims; dual=dual1)
 end
 
+function ⊖(V::GradedSpace{I}, W::GradedSpace{I}) where {I<:Sector}
+    V ≿ W && isdual(V) == isdual(W) ||
+        throw(SpaceMismatch("$(W) is not a subspace of $(V)"))
+    return typeof(V)(c => dim(V, c) - dim(W, c) for c in sectors(V))
```
Jutho: The `isdual` flag of `V` (and `W`, as it is required to be the same) is lost in this construction. See the `⊕` implementation.

src/tensors/backends.jl
```diff
@@ -0,0 +1,32 @@
+# Scheduler implementation
+# ------------------------
+function select_scheduler(scheduler=OhMyThreads.Implementation.NotGiven(); kwargs...)
+    return if scheduler == OhMyThreads.Implementation.NotGiven() && isempty(kwargs)
+        Threads.nthreads() > 1 ? SerialScheduler() : DynamicScheduler()
```
Jutho: Is this the right order? `nthreads() > 1`, then use `SerialScheduler()`. So if `nthreads()==1`, then use `DynamicScheduler()`?

src/tensors/blockiterator.jl
```diff
@@ -13,3 +13,29 @@ Base.IteratorEltype(::BlockIterator) = Base.HasEltype()
 Base.eltype(::Type{<:BlockIterator{T}}) where {T} = blocktype(T)
 Base.length(iter::BlockIterator) = length(iter.structure)
 Base.isdone(iter::BlockIterator, state...) = Base.isdone(iter.structure, state...)
+
+# TODO: fast-path when structures are the same?
+# TODO: do we want f(c, bs...) or f(c, bs)?
```
Jutho: With `c` present (which is often useful), I like the current version `f(c, bs)`. It also fits nicely with how we keep e.g. factorization data together as a single tuple in MatrixAlgebraKit, i.e. the `USVᴴ` variable and friends.

Without the `c` argument passed on to `f`, I might prefer `f(bs...)`, but I agree that having `c` available can be necessary. 

Anyway, I don't think it matters as long as it is well documented.

src/tensors/diagonal.jl
```diff
@@ -265,6 +275,21 @@ function LinearAlgebra.mul!(dC::DiagonalTensorMap,
     return dC
 end
 
+function LinearAlgebra.lmul!(D::DiagonalTensorMap, t::AbstractTensorMap)
+    domain(D) == codomain(t) || throw(SpaceMismatch())
+    for (c, b) in blocks(t)
```
Jutho: Do we immediately want to write this using `foreachblock`?

src/tensors/diagonal.jl
```diff
@@ -265,6 +275,21 @@ function LinearAlgebra.mul!(dC::DiagonalTensorMap,
...
+        lmul!(block(D, c), b)
+    end
+    return t
+end
+function LinearAlgebra.rmul!(t::AbstractTensorMap, D::DiagonalTensorMap)
+    codomain(D) == domain(t) || throw(SpaceMismatch())
+    for (c, b) in blocks(t)
```
Jutho: Same comment/question.

src/tensors/factorizations/factorizations.jl
```diff
@@ -0,0 +1,205 @@
...
+`t` to be destroyed/overwritten, by using `isposdef!(t)`. Note that the permuted tensor on
+which `isposdef!` is called should have equal domain and codomain, as otherwise it is
+meaningless.
+"""
+function LinearAlgebra.isposdef(t::AbstractTensorMap, (p₁, p₂)::Index2Tuple)
+    tcopy = permutedcopy_oftype(t, factorisation_scalartype(isposdef, t), p)
+    return isposdef!(tcopy)
+end
+function LinearAlgebra.isposdef(t::AbstractTensorMap)
+    tcopy = copy_oftype(t, float(scalartype(t)))
```
Jutho: Should the second argument also be `factorisation_scalartype(isposdef, t)`?

lkdvos: I wanted to ask this still: do you think it is reasonable to just adopt `isposdef` in the list above, and actually define it both for `isposdef(t, p)` and `isposdef`?

lkdvos: I went ahead and implemented this as such, if you disagree I'm happy to change it back.

src/spaces/vectorspaces.jl
```diff
@@ -150,6 +150,17 @@ function ⊕ end
 ⊕(V::Vararg{VectorSpace}) = foldl(⊕, V)
 const oplus = ⊕
 
+"""
+    ⊖(V::ElementarySpace, W::ElementarySpace) -> X::ElementarySpace
+    ominus(V::ElementarySpace, W::ElementarySpace) -> X::ElementarySpace
+
+Return the set difference of two elementary spaces, i.e. an instance `X::ElementarySpace`
```
Jutho:
```suggestion
Return a space that is equivalent to the orthogonal complement of `W` in `V`, i.e. an instance `X::ElementarySpace`
```

src/TensorKit.jl
```diff
@@ -70,10 +70,10 @@ export inner, dot, norm, normalize, normalize!, tr
 
 # factorizations
 export mul!, lmul!, rmul!, adjoint!, pinv, axpy!, axpby!
-export leftorth, rightorth, leftnull, rightnull,
-       leftorth!, rightorth!, leftnull!, rightnull!,
+export leftorth, rightorth, leftnull, rightnull, leftpolar, rightpolar,
+       leftorth!, rightorth!, leftnull!, rightnull!, leftpolar!, rightpolar!,
```
Jutho: Do you want to add new methods like `leftpolar` here, or do we want to switch over to MatrixAlgebraKit conventions and use `left_` and `right_` (with underscores), and keep `leftorth` and friends for backward compatibility (with possible deprecations in a future version)?

src/tensors/factorizations/implementations.jl
```diff
@@ -0,0 +1,185 @@
+_kindof(::Union{SVD,SDD}) = :svd
+_kindof(::Union{QR,QRpos}) = :qr
+_kindof(::Union{LQ,LQpos}) = :lq
+_kindof(::Polar) = :polar
+
+for f! in (:svd_compact!, :svd_full!, :left_null_svd!, :right_null_svd!)
+    @eval function select_algorithm(::typeof($f!), t::T, alg::SVD;
+                                    kwargs...) where {T}
```
Jutho: Is this for general types `T`?

src/tensors/factorizations/matrixalgebrakit.jl
```diff
@@ -0,0 +1,528 @@
...
+end
+
+function initialize_output(::typeof(left_polar!), t::AbstractTensorMap, ::AbstractAlgorithm)
+    W = similar(t, space(t))
+    P = similar(t, domain(t) ← domain(t))
+    return W, P
+end
+
+function check_input(::typeof(right_polar!), t::AbstractTensorMap, (P, Wᴴ)::_T_PWᴴ)
+    domain(t) ≿ codomain(t) ||
```
Jutho: Is there a specific preference for writing the checks using `≿` versus `≾` ?

I know this is nitpicking, but I would either prefer consistent use of `≾` , or alternatively, consistent use of `codomain(t)` as first argument and `domain(t)` as second, with varying operator.

lkdvos:
Okay, I think this is finally ready for review + merging. I did my best to deprecate as much as possible while being reasonable about not spending too much effort dealing with things we will be deleting anyways.

There is still a section in the docs that is referring to the old factorizations, but I think we should try and tackle that when we rework the docs.

In principle it would be slightly nicer to immediately have [MatrixAlgebraKit v0.4](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/56) so we don't have to deal with renaming the truncation strategies in a release, but I would also be okay with just merging this PR, without releasing, and then fixing that before we release a new version.

lkdvos:
So this being done was clearly a lie, and I had to tweak a bunch more things. I think the main body of the PR is now figured out, I'm fighting a little bit to also port over the AD rules but this can definitely already use a review.

docs/src/lib/tensors.md
```diff
@@ -214,26 +214,21 @@ contract!
 
 ## `TensorMap` factorizations
 
-The factorisation methods come in two flavors, namely a non-destructive version where you
-can specify an additional permutation of the domain and codomain indices before the
-factorisation is performed (provided that `sectorstyle(t)` has a symmetric braiding) as
-well as a destructive version The non-destructive methods are given first:
+The factorisation methods are powered by [MatrixAlgebraKit.jl](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl)
+and all follow the same strategy. The idea is that the `TensorMap` is interpreted as a linear
+map  based on the current partition of indices between `domain` and `codomain`, and then the
```
Jutho:
```suggestion
map based on the current partition of indices between `domain` and `codomain`, and then the
```

docs/src/lib/tensors.md
```diff
@@ -214,26 +214,21 @@ contract!
...
+entire range of MatrixAlgebraKit functions can be called.
+You can specify an additional permutation of the domain and codomain indices before the
+factorisation is performed by making use of [`permute`](@ref) or [`transpose`](@ref).
```
Jutho:
```suggestion
Factorizing a tensor according to a different partition of the indices is possible
by prepending the factorization step with an explicit call to [`permute`](@ref) or [`transpose`](@ref).
```

src/tensors/linalg.jl
```diff
@@ -449,11 +455,11 @@ for f in (:cos, :sin, :tan, :cot, :cosh, :sinh, :tanh, :coth, :atan, :acot, :asi
         tf = similar(t, T)
         if T <: Real
             for (c, b) in blocks(t)
-                copy!(block(tf, c), real(MatrixAlgebra.$f(b)))
+                copy!(block(tf, c), real(MatrixAlgebraKit.$f(b)))
```
Jutho, line 458: This will not work for now, right? We haven't implemented this yet. Or am I missing something?

ext/TensorKitChainRulesCoreExt/factorizations.jl
```diff
@@ -1,481 +1,211 @@
...
+        Δt = zerovector(t)
+        ΔQ = zerovector!(similar(Q, codomain(Q) ← fuse(codomain(Q))))
+        foreachblock(ΔN) do c, (b,)
+            n = size(b, 2)
+            ΔQc = block(ΔQ, c)
+            return copy!(@view(ΔQc[:, (end - n + 1):end]), b)
         end
-        return NoTangent(), Δt
+        ΔR = ZeroTangent()
+        MatrixAlgebraKit.qr_compact_pullback!(Δt, t, (Q, R), (ΔQ, ΔR))
```
Jutho: I guess calling the pullback rule in MAK should simply have been called `qr_pullback!`, since it works for both `qr_compact!` and `qr_full!`. This was clearly an unfortunate oversight.

Jutho: Also, I assume it be easier to directly have a `qr_null_pullback!` in MAK.

Jutho: Ok, I now see that there is a `ChainRules.rrule` for `qr_null` in MAK, but not a dedicated generic pullback in the main module. So the `qr/lq_null` functions are a bit odd in that respect.

Also, for most of the other `ChainRules.rrule` definitions in MAK, they are restricted to `A::AbstractMatrix`, but they are completely general in terms of how they lower to the generic pullback in the main MAK module.

Would it have been more useful to lift this restriction, and then only overload the `MAK.f_pullback` functions here in TensorKit? This would probably allow to eliminate a lot of the replicated `NoTangent` / `ZeroTangent` code?

src/auxiliary/deprecate.jl
```diff
@@ -58,4 +35,161 @@ Base.@deprecate EuclideanProduct() EuclideanInnerProduct()
...
+function leftorth(t::AbstractTensorMap; kwargs...)
+    Base.depwarn("`leftorth` is no longer supported, use `left_orth` instead", :leftorth)
+    return leftorth!(copy_oftype(t, factorisation_scalartype(leftorth, t)); kwargs...)
+end
+function rightorth(t::AbstractTensorMap; kwargs...)
+    Base.depwarn("`rightorth` is no longer supported, use `right_orth` instead", :rightorth)
+    return rightorth!(copy_oftype(t, factorisation_scalartype(rightorth, t)); kwargs...)
+end
+function leftorth!(t::AbstractTensorMap; kwargs...)
+    Base.depwarn("`leftorth!` is no longer supported, use `left_orth!` instead", :leftorth!)
```
Jutho: I kind of prefer `... is deprecated` over `... is no longer supported`, as the latter seems to suggest that it already doesn't work anymore, whereas you've gone to great lengths to make sure they do still work.

src/tensors/factorizations/utility.jl
```diff
@@ -0,0 +1,29 @@
...
+    return esc(:($MatrixAlgebraKit.@check_scalar($x, $y, $op, $eltype)))
+end
+
+function factorisation_scalartype(t::AbstractTensorMap)
+    T = scalartype(t)
+    return promote_type(Float32, typeof(zero(T) / sqrt(abs2(one(T)))))
+end
+factorisation_scalartype(f, t) = factorisation_scalartype(t)
+
+function permutedcopy_oftype(t::AbstractTensorMap, T::Type{<:Number}, p::Index2Tuple)
```
Jutho: Is `permutedcopy_oftype` still used anywhere other than in the deprecations? If not, it could potentially also be moved to "src/auxiliary/deprecate.jl".

src/tensors/linalg.jl
lkdvos: It does work, but it's a bit of an unconventional construction, in anticipation of possibly having them there we qualify the `MatrixAlgebraKit` namespace, but these bindings actually resolve to `MatrixAlgebraKit.f === Main.f`

src/factorizations/truncation.jl
```diff
@@ -0,0 +1,230 @@
...
+    truncate_domain!(Ũ, U, ind)
+    return Ũ, ind
+end
+
+for f! in (:eig_trunc!, :eigh_trunc!)
+    @eval function MAK.truncate(::typeof($f!),
+                                (D, V)::Tuple{DiagonalTensorMap,AbstractTensorMap},
+                                strategy::TruncationStrategy)
+        ind = MAK.findtruncated(diagview(D), strategy)
+        V_truncated = spacetype(D)(c => length(I) for (c, I) in ind)
```
Jutho: Why is this one not using `truncate_space`?

lkdvos: oops 🙃

src/factorizations/truncation.jl
```diff
@@ -0,0 +1,230 @@
...
+        @assert !isnothing(I)
+        copy!(b, @view(block(tsrc, c)[I, :]))
+    end
+    return tdst
+end
+function truncate_diagonal!(Ddst::DiagonalTensorMap, Dsrc::DiagonalTensorMap, inds)
+    for (c, b) in blocks(Ddst)
+        I = get(inds, c, nothing)
+        @assert !isnothing(I)
+        copy!(diagview(b), @view(diagview(block(Dsrc, c))[I]))
```
Jutho: I find it a bit hard to parse to where the `@view` effect ranges, and I actually think the following is less character(s):


```suggestion
        copy!(diagview(b), view(diagview(block(Dsrc, c)), I))
```

PR #232, opened 2025-03-19
Rewrite special-purpose constructors to use in-place operations

lkdvos:
Refactor special-purpose constructors to utilize in-place implementations instead:
```julia
t = zeros(ComplexSpace(2), ComplexSpace(2))
unitary!(t)
```

### DIFF
--- a/src/tensors/linalg.jl
+++ b/src/tensors/linalg.jl
@@ -64,24 +64,31 @@ function one!(t::AbstractTensorMap)
     return t
 end
 
-"""
+@doc """
     id([T::Type=Float64,] V::TensorSpace) -> TensorMap
+    id!(t::AbstractTensorMap) -> AbstractTensorMap
 
 Construct the identity endomorphism on space `V`, i.e. return a `t::TensorMap` with
 `domain(t) == codomain(t) == V`, where either `scalartype(t) = T` if `T` is a `Number` type
 or `storagetype(t) = T` if `T` is a `DenseVector` type.
-"""
+
+See also [`one!`](@ref).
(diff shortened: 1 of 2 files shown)


src/tensors/linalg.jl
```diff
@@ -137,13 +147,13 @@ Construct a specific isometry between the codomain and the domain, i.e. return a
...
+""" isometry, isometry!
+
+function isometry!(t::AbstractTensorMap)
+    InnerProductStyle(t) === EuclideanInnerProduct() ||
         throw_invalid_innerproduct(:isometry)
-    domain(V) ≾ codomain(V) ||
-        throw(SpaceMismatch("$V does not allow for an isometric inclusion"))
-    t = tensormaptype(S, N₁, N₂, A)(undef, V)
+    domain(t) ≾ codomain(t) ||
+        throw(SpaceMismatch("domain and codomain are not isometrically embeddable: $t"))
```
Jutho: Is this including the whole tensor in the error message?

Jutho:
Looks good except for the comment/question.

src/tensors/linalg.jl
lkdvos: Woops, definitely did not mean to do that. Good catch, thanks!

lkdvos:
Good catch, thanks!
At some point we should maybe also normalize our error messages here as well.
I'm still not sure if we want to use `lazy` everywhere, include information, or not, and it seems like we still have `SectorMismatch` and `SpaceMismatch` both being used for similar things in certain locations.

PR #233, opened 2025-03-21
Add benchmark suite

lkdvos:
This sets up a `benchmark/benchmarks.jl` script to have a dedicated `SUITE` to run our benchmarks, similar to `BaseBenchmarks.jl`.

The goal is to start adding more coverage so we can better assess the actual performance improvements and/or regressions between versions.

(This is not meant as being illustrative for the performance of this package, for which I am creating separate scripts to generate proper figures etc)


benchmark/TensorKitBenchmarks/utils/BenchUtils.jl
```diff
@@ -0,0 +1,74 @@
...
+
+        if abs(sector.charge) > 20
+            error()
+        end
+    end
+
+    return U1Space((s => d for (s, d) in zip(sectors, dims))...)
+end
+function generate_space(::Type{SU2Irrep}, D::Int, sigma::Real=0.5)
+    poisson_pdf(x) = ceil(Int, D * exp(-sigma) * sigma^x / factorial(x))
```
Jutho: Strictly speaking, this is not correctly normalised. If we assume that `SU2Irrep(j)` is indeed generated with dimension `poisson_pdf(2*j) = D * exp(-sigma) * sigma^(2j) / factorial(2j)`, then the total dimension is, with `n=2j`

$$\sum_{n=0}^{+\infty} D * (n+1) * exp(-\sigma) * \sigma^n / n! \neq D$$

due to the extra factor $$n+1=2j+1$$, which comes from the dimension of the irrep.

Maybe `poisson_pdf(x) = ceil(Int, D * exp(-sigma) * sigma^x / factorial(x) / (x+1))`

which happens to be

Maybe `poisson_pdf(x) = ceil(Int, D * exp(-sigma) * sigma^x / factorial(x+1))`

as alternative

Jutho:
```suggestion
    poisson_pdf(x) = ceil(Int, D * exp(-sigma) * sigma^x / factorial(x+1))
```

Jutho: Also, for large sigma, the poisson distribution is peaked away from zero, so I do not know if that is really the distribution we want, versus just a half gaussian, but I guess those details don't really matter (but we probably do not want to change them afterwards).

benchmark/TensorKitBenchmarks/tensornetworks/TensorNetworkBenchmarks.jl
```diff
@@ -0,0 +1,145 @@
...
+    end
+    return nothing
+end
+
+function benchmark_mera!(bench; sigmas=nothing, T="Float64", I="Trivial", dims)
+    T_ = parse_type(T)
+    I_ = parse_type(I)
+
+    Vs = generate_space.(I_, dims, sigmas)
+    init() = init_mera_tensors(T_, Vs)
```
Jutho, line 131: I am confused how this works, as `init_mera_tensors` requires a single space `V` as second argument, and this call seems to insert a list `Vs`?

Jutho: Oh I think I understand; the incoming `dims` is a single integer and the broadcast doesn't do anything and only generate a single space, not wrapped in a list.

Jutho (approved):

benchmark/TensorKitBenchmarks/utils/BenchUtils.jl
lkdvos: I'm definitely happy to switch that, although I think the peak is reasonable too: I like poisson for SU(2) since in the Heisenberg virtual spaces often there are more values of spin 1 than spin 0.

Probably it doesn't matter too much for the actual representativeness of the benchmarks though

PR #236, opened 2025-03-31
Add `rrule` for `inv` and corresponding tests

lkdvos:


Jutho:
Do you understand the nightly tests? Otherwise looks good.

lkdvos:
Looks like some internal problems with ChainRulesTestUtils to me, although I checked and there haven't really been any changes there. My guess would be it has something to do with changes to world age things etc, but I am not really following that all too closely so I can't say for sure. In any case, looks unrelated to this change :)

Jutho:
Ok it is just annoying if nightly breaks the master branch, especially if we want to tag a version soon.

PR #240, opened 2025-04-07
Allow twisting of trivial sectors without restrictions

### LINKED ISSUE #239: `twist` for `NoBraiding` sector types where the braiding is unambiguous
For categories that have no braiding, we can still allow for some braiding rules by allowing the identity to (trivially) braid. This is something that is used in MPSKit (although this is a bit of a hidden feature: it is implemented at the level of the fusion tree manipulations, not the actual sectors)

In https://github.com/QuantumKitHub/MPSKit.jl/pull/274 I'm trying to fix the periodic boundary conditions, and getting an error on calling `twist` on a trivial space. (`PlanarTrivial` to be precise).
I was wondering if it makes sense to have similar semantics as for the braiding rules, where twisting a trivial sector is always allowed.

lkdvos:
Enable twisting of trivial sectors in cases where no braiding is defined, similar to how braiding trivial sectors is special cased.

Fixes #239

### DIFF
--- a/.github/workflows/FormatCheck.yml
+++ b/.github/workflows/FormatCheck.yml
@@ -29,7 +29,7 @@ jobs:
         #
         # julia  -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter", version="0.13.0"))'
         run: |
-          julia  -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter"))'
+          julia  -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter", version="1"))'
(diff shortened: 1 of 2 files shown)


Jutho:
Looks good; feel free to merge.

Note: it would be good to have tests for both the braiding and twisting behaviour of indices with a unit sector at some point. However, we currently don't even have any non-group symmetric tensor tests.

lkdvos:
MPSKit remains a somewhat decent test suite of this, since it's actually being used there :)

I was actually quite surprised to find that the trivial braiding thing worked, and had to dig through the code a bit to find out how this was being achieved.

PR #241, opened 2025-04-14
add rrule for flip

sanderdemeyer:
This implements the rrule for the function flip().

### DIFF
--- a/src/tensors/indexmanipulations.jl
+++ b/src/tensors/indexmanipulations.jl
@@ -5,17 +5,19 @@
 
 Return a new tensor that is isomorphic to `t` but where the arrows on the indices `i` that satisfy
 `i ∈ I` are flipped, i.e. `space(t′, i) = flip(space(t, i))`.
+
+!!! note
+    The isomorphism that `flip` applies to each of the indices `i ∈ I` is such that flipping two indices
+    that are afterwards contracted within an `@tensor` contraction will yield the same result as without
+    flipping those indices first. However, `flip` is not involutary, i.e. `flip(flip(t, I), I) != t` in
+    general. To obtain the original tensor, one can use the `inv` keyword, i.e. it holds that
+    `flip(flip(t, I), I; inv=true) == t`.
 """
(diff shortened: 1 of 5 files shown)


lkdvos:
Can you also add a test? See https://github.com/Jutho/TensorKit.jl/blob/ed09f138e41d8a69a703f6473e4acbb2838f2fb5/test/ad.jl#L224

Jutho:
It turns out that this is not the right `rrule`, because `flip` is not an involution, i.e. it is not its own inverse. More specifically, the inverse of flip is flipping the index 3 times (for fermions). Probably we want to have an associated `unflip` function? I will make some changes to this PR to make this happen.

lkdvos:
That sounds ... honestly scary 😬 It might be good to add a note of this in the docstring, since this is definitely something I would have wrongly assumed

Jutho:
It essentially amounts to saying that flipping twice corresponds exactly to how you get the Frobenius-Schur indicator, which is ±1. 

An `inv` keyword is probably the better option indeed.

src/tensors/indexmanipulations.jl
```diff
@@ -5,17 +5,19 @@
 
 Return a new tensor that is isomorphic to `t` but where the arrows on the indices `i` that satisfy
 `i ∈ I` are flipped, i.e. `space(t′, i) = flip(space(t, i))`.
+
+!!! note
+    The isomorphism that `flip` applies to each of the indices `i ∈ I` is such that flipping two indices
+    that are afterwards contracted within an `@tensor` contraction will yield the same result as without
+    flipping those indices first. However, `flip` is not involutary, i.e. `flip(flip(t, I), I) != t` in
```
```suggestion
    flipping those indices first. However, `flip` is not involutory, i.e. `flip(flip(t, I), I) != t` in
```
I'm not entirely sure about this, but involutary feels off to me. Wikipedia says involutory, but I would have guessed involutive.

lkdvos (approved):
Left a small note about typography, otherwise looks good to me!

Issue #242, opened 2025-04-15
redesign caches and clear/disable during tests


Jutho:
Currently we have the following cache variables (I hope I am not missing any):

```julia
TensorKit.GLOBAL_FUSIONBLOCKSTRUCTURE_CACHE
TensorKit.treetransposercache
TensorKit.treebraidercache
TensorKit.treepermutercache
TensorKit.transposecache
TensorKit.braidcache
```

As well as the control flags
```julia
TensorKit.usebraidcache_abelian
TensorKit.usebraidcache_nonabelian
TensorKit.usetransposecache
TensorKit.usetreetransposercache
TensorKit.usetreebraidercache
TensorKit.usetreepermutercache
```

and

```julia
abstract type CacheStyle end
struct NoCache <: CacheStyle end
struct TaskLocalCache{D<:AbstractDict} <: CacheStyle end
struct GlobalLRUCache <: CacheStyle end

function CacheStyle(I::Type{<:Sector})
    return GlobalLRUCache()
end

fusionblockstructure(W::HomSpace) = fusionblockstructure(W, CacheStyle(sectortype(W)))
```

for the fusion blockstructure caches.

We should probably make this a bit more uniform. Importantly, we should disable or empty caches while running the tests, as they in the end collect close to a gigabyte of cached data, with probably only negative effects on the test performance.

lkdvos:
I would say that the most important parts would be the ability to visualize and monitor cache usage, as well as easily clearing it. Also, I think we should be a little more conservative with the default values for the maximal sizes.

As a side note, we also have some caches for `Fsymbol` and `Rsymbol`, but only for some sector types (`SUNRepresentations.jl` mostly). Do we also want to include these in this discussion?

lkdvos:
Closed since #244

PR #244, opened 2025-04-16
make all caches behave the same

Jutho:
A first attempt at resolving #242


src/auxiliary/caches.jl
```diff
@@ -0,0 +1,148 @@
+const GLOBAL_CACHES = Any[]
+function empty_globalcaches!()
+    foreach(empty!, GLOBAL_CACHES)
+    return nothing
+end
```
lkdvos: Is there a way to have some info display function as well here? Related to this, storing the `GLOBAL_CACHES` as a vector does lose some of the information about which cache it is, it might be reasonable to use a `Dict` or `IdDict` instead to keep the name around, or store the name together with the cache

Jutho:
@lkdvos, thanks for finding the bug and fixing it. Not sure why CI didn't run on your last commit?

Jutho:
Oh I see, it is still waiting for the CI of the other PR to finish.

lkdvos:
I have absolutely no idea about the CI, I had to trigger that manually somehow

src/fusiontrees/manipulations.jl
```diff
@@ -548,28 +544,24 @@ function Base.transpose(f₁::FusionTree{I}, f₂::FusionTree{I},
...
-        return _transpose(key)
-    end
-    return d
-end
+const FSTransposeKey{I<:Sector,N₁,N₂} = Tuple{<:FusionTree{I},<:FusionTree{I},
+                                              IndexTuple{N₁},IndexTuple{N₂}}
 
-const TransposeKey{I<:Sector,N₁,N₂} = Tuple{<:FusionTree{I},<:FusionTree{I},
-                                            IndexTuple{N₁},IndexTuple{N₂}}
+Base.@pure function _fsdicttype(I, N₁, N₂)
```
lkdvos: Should we replace this with `Base.@assume_effects :foldable`? Is this required at all?

src/spaces/homspace.jl
```diff
@@ -239,18 +239,17 @@ struct FusionBlockStructure{I,N,F₁,F₂}
...
 end
 
-abstract type CacheStyle end
-struct NoCache <: CacheStyle end
-struct TaskLocalCache{D<:AbstractDict} <: CacheStyle end
-struct GlobalLRUCache <: CacheStyle end
-
-function CacheStyle(I::Type{<:Sector})
-    return GlobalLRUCache()
+function fusionblockstructuretype(W::HomSpace)
```
lkdvos, line 236: Does this also need `Base.@assume_effects`/`Base.@pure`?

src/auxiliary/caches.jl
```diff
@@ -0,0 +1,148 @@
...
+
+abstract type CacheStyle end
+struct NoCache <: CacheStyle end
+struct TaskLocalCache{D<:AbstractDict} <: CacheStyle end
+struct GlobalLRUCache <: CacheStyle end
+
+const DEFAULT_GLOBALCACHE_SIZE = Ref(10^5)
```
lkdvos: I ran a very quick benchmark case by having a VUMPS run of the isotropic spin 1 Heisenberg chain with SU2 symmetry, (along with a bit of fiddling before), and these are some relevant sizes for the tensor caches, which should be the same for any symmetry, irrespective of the bond dimension:
```julia
julia> length(TensorKit.GLOBAL_FUSIONBLOCKSTRUCTURE_CACHE)
151
julia> length(TensorKit.treepermutercache)
74
julia> length(TensorKit.treetransposercache)
7
```

For the fusion trees themselves these numbers clearly heavily depend on the symmetry, since they actually scale with the number of fusion trees that each tensor has, but at least this gives:
```julia
julia> length(TensorKit.braidcache)
238
```

In other words, I would argue that for the "tensor-level" caches `10^5` is just a huge overkill, and not something we should do.
`10^3` still seems quite generous, but actually I do think it would make sense to have the defaults here be different for the different caches: the tensor structure is expected to have more elements, so we might increase the value for that one?

For the fusion trees, `10^5` might actually sound reasonable, but it might also make sense to make this depend on the symmetry? I would probably have to investigate in a little more detail to see how this plays out for hubbard-like symmetries in PEPSKit for example.

Is there a way to actually control this with the current implementation though?

lkdvos (commented):
I think that this would actually be a nice addition, so perhaps we can push this through?

src/auxiliary/caches.jl
Jutho: There is no cache per sector type, right? There is one global cache (e.g. `transposecache` or `braidcache`), that keeps all information for all splitting-fusion tree pairs for all possible sector types. So making the cache size depend on the sector thus doesn't really make sense.

Jutho: Could you also try some PEPS simulation, e.g. some Hubbard simulation?

Jutho: I can also ask one of our Hubbard people.

lkdvos: For a CTMRG run of Hubbard with fZ2 x U1 I get the following for a fixed-space truncation:

```julia
julia> cache_info(TensorKit.GLOBAL_FUSIONBLOCKSTRUCTURE_CACHE)
CacheInfo(; hits=1985197, misses=62, currentsize=62, maxsize=10000)

julia> cache_info(TensorKit.treepermutercache)
CacheInfo(; hits=40868, misses=52, currentsize=52, maxsize=100000)
```
The other ones are empty because this counts as "abelian symmetry".

If I allow for some truncation to happen I get, after 100 iterations:
```julia
julia> cache_info(TensorKit.GLOBAL_FUSIONBLOCKSTRUCTURE_CACHE)
CacheInfo(; hits=3352630, misses=4344, currentsize=4344, maxsize=10000)

julia> cache_info(TensorKit.treepermutercache)
CacheInfo(; hits=79407, misses=2433, currentsize=2433, maxsize=100000)
```
(Keeping in mind that we currently seem to have an issue with the CTMRG run for hubbard, I'll look into that)

For contrast, I manually set the `maxsize` to `1000` and retried:
```julia
julia> cache_info(TensorKit.GLOBAL_FUSIONBLOCKSTRUCTURE_CACHE)
CacheInfo(; hits=1377722, misses=5204, currentsize=1000, maxsize=1000)

julia> cache_info(TensorKit.treepermutercache)
CacheInfo(; hits=38059, misses=2861, currentsize=1000, maxsize=1000)
```

And for `100`:
```julia
julia> cache_info(TensorKit.GLOBAL_FUSIONBLOCKSTRUCTURE_CACHE)
CacheInfo(; hits=1371502, misses=46069, currentsize=100, maxsize=100)

julia> cache_info(TensorKit.treepermutercache)
CacheInfo(; hits=19253, misses=21667, currentsize=100, maxsize=100)
```

In other words, it is very problem dependent and it might actually matter.
The main point is however that once the spaces stabilize the actual required size is very small, so maybe we can also just update our docs to make it easier to actually clear some entries (for example, it would probably be convenient to have some way of dropping all but the most recent 100 entries once you switch from dynamic to fixed bond dimension runs)

Alternatively, we could consider some more advanced eviction policies that also evict things that haven't been used in a long time, but I am guessing neither of us really has the time to experiment with that

[EDIT] the actual timings for the runs, starting from the same peps but random ctmrg environments of the same size was 6.79, 7.44 and 9.67 seconds respectively. I'd have to check more thoroughly to verify that this is actually reasonable, and start from the same initial state, but still wanted to add this information

src/fusiontrees/manipulations.jl
Jutho: It is probably no longer required with constant folding; I used to have this in any function that took in values which were returned in the type domain, i.e. the parameters `N₁` and `N₂`.

src/spaces/homspace.jl
Jutho: Here, only information from the type domain, i.e. from `typeof(W)` are used to construct the return value, so I didn't think this is necessary.

src/auxiliary/caches.jl
Jutho: So what should now be the conclusion to make this PR land? 10^5 is definitely too big, and somewhere in the 10^3 - 10^4 range is a more reasonable default? I am happy to accept/consider with more refined strategies and configuration options, but that could happen in a separate PR I guess?

lkdvos:
So as a compromise, I set the default cache size to `10^4`, resolved the comments, and added a really simple function to just print the status of all of the caches, which I think will be helpful. If tests pass, this is good to go for me.

Jutho:
> So as a compromise, I set the default cache size to `10^4`, resolved the comments, and added a really simple function to just print the status of all of the caches, which I think will be helpful. If tests pass, this is good to go for me.

Too bad real-life compromises are typically not made on a logarithmic scale 😄.

Anyway, I like the latest changes to this PR.

PR #247, opened 2025-06-13
Compatibility with MultiTensorKit

borisdevos:
This PR can be divided in three parts:

1) The largest part concerns fusion tree manipulations where the type of unit (`leftone` or `rightone`) matters in a multifusion context. As for now during the draft, I've left some comments where the type of unit was not rigorously derived, but guessed to get the code running. These will be removed when benchmarking MultiTensorKit is complete.

2) A potential hack was used to promote storage types when I was working with `ComplexF64` F-symbols in MultiTensorKit. 

3) Minor docs and docstring typos.


src/fusiontrees/manipulations.jl
```diff
@@ -230,7 +231,7 @@ function merge(f₁::FusionTree{I,N₁}, f₂::FusionTree{I,N₂},
     return insertat(f, N₁ + 1, f₂)
 end
 function merge(f₁::FusionTree{I,0}, f₂::FusionTree{I,0}, c::I, μ) where {I}
-    isone(c) ||
+    isone(c) || # I had this as Nsymbol(f₁.coupled, f₂.coupled, c) > 0, valid?
```
lkdvos: I think this is valid, and probably a slightly more rigorous check anyways. In principle `f::FusionTree{I,0}` should already satisfy `isone(f.coupled)`, so in these cases that boils down to the same thing.

```suggestion
    Nsymbol(f₁.coupled, f₂.coupled, c) == μ == 1 ||
```

I initially added the `<= mu` there, although I guess this really has to be `1` for this to make sense, hence the suggestion

src/tensors/tensoroperations.jl
lkdvos, line 1: For these changes, I think it would be nice to have a dedicated discussion (indeed in a separate PR) and some comments explaining what is going on.

The bottom line being that we want the `sectorscalartype` to mix into the actual `scalartype` in a very subtle kind of way, only really being used to promote from `<:Real` to `<:Complex` while leaving the actual precision the same. (I think?)

The implementation below is somewhat of a hack since normally `TensorOperations` is responsible for figuring out the destination scalartype `TC`, so in principle we would expect `scalartype(tensoradd_type(TC, ...)) === TC`, which is not the case here.

There are a couple options for fixing this:

1. Since `TensorOperations` uses `scalartype`, we could consider redefining `scalartype(::AbstractTensorMap)` to take the `sectorscalartype` into account. This would boil down to really having a different meaning between `eltype` and `scalartype` for TensorMaps, where the former is the type of the stored data, and the latter also contains information about the field.
2. We always require complex entries if the `sectorscalartype` is complex. This doesnt have too many performance implications since most operations on tensors would result in a complex tensor anyways, but we would have to think carefully about `DiagonalTensorMap`, since that one could have real entries (e.g. for singular value decompositions)
3. We rework the implementation of `promote_contract` in `TensorOperations` and the macros to actually work on values or types thereof instead of directly on `scalartype`s, such that we can overload `promote_contract(::AbstractTensorMap, ::AbstractTensorMap, ::Number)` or `promote_contract(::Type{<:AbstractTensorMap}, ...)`. This is quite a big internal change for TensorOperations though, since currently all macros expand to `promote_contract(scalartype(A), scalartype(B), ...)`.

src/fusiontrees/manipulations.jl
```diff
@@ -165,6 +165,7 @@ operation is the inverse of `insertat` in the sense that if
         f₂ = FusionTree{I}(f.uncoupled, f.coupled, isdual2, f.innerlines, f.vertices)
         return f₁, f₂
     elseif M === 0
+        # TODO: evaluate diagrams to see which unit is used here
```
Jutho: I would guess that is `leftone(f.uncoupled[1])`?

src/tensors/tensoroperations.jl
Jutho: Is this something that came up specifically in this PR? This seems like it should already have been an issue with existing complex sectors.

Jutho: Is this something that came up specifically in this PR? This seems like it should already have been an issue with existing complex sectors.

lkdvos: Yes, but I don't think we ever used complex sectors with non-complex tensors all that much, and MPSKit actually uses complex entries for almost everything by default.

src/fusiontrees/iterator.jl
```diff
@@ -157,7 +157,7 @@ function _fusiontree_iterate(uncoupledsectors::NTuple{N},
         nextout = iterate(outiterN, outstateN)
         nextout === nothing && return nothing
         b, outstateN = nextout
-        vertexiterN = c ⊗ dual(b)
+        vertexiterN = coupled ⊗ dual(b)
```
Jutho, line 160: Thanks for catching this. Did you encounter this by reading the code or by an actual case where the old implementation errored? Clearly there should have been a test in place to cover those lines of code. So if you have some example code that actually triggers these lines, that might be a good starting point to extract a test case from (which might be easier then engineering one from scratch).

src/fusiontrees/manipulations.jl
```diff
@@ -165,8 +165,8 @@ operation is the inverse of `insertat` in the sense that if
         f₂ = FusionTree{I}(f.uncoupled, f.coupled, isdual2, f.innerlines, f.vertices)
         return f₁, f₂
     elseif M === 0
-        f₁ = FusionTree{I}((), one(I), (), ())
-        uncoupled2 = (one(I), f.uncoupled...)
+        f₁ = FusionTree{I}((), leftone(f.uncoupled[1]), (), ())
+        uncoupled2 = (leftone(f.uncoupled[1]), f.uncoupled...)
```
Jutho:
```suggestion
        u = leftone(f.uncoupled[1])
        f₁ = FusionTree{I}((), u, (), ())
        uncoupled2 = (u, f.uncoupled...)
```

src/fusiontrees/manipulations.jl
```diff
@@ -722,14 +723,16 @@ function elementary_trace(f::FusionTree{I,N}, i) where {I<:Sector,N}
     T = sectorscalartype(I)
     F = fusiontreetype(I, N - 2)
     newtrees = FusionTreeDict{F,T}()
+    _one = f.coupled
```
Jutho: Is there a reason to make this variable here already, it only seems to be used in line 773, so would it make sense to just define it there? Or to use `f.coupled` directly? I also like `u` or `unit` (that variable is defined already below, is it the same?) for unit objects.

src/tensors/tensoroperations.jl
```diff
@@ -52,7 +52,9 @@ end
 
 function TO.tensoradd_type(TC, A::AbstractTensorMap, ::Index2Tuple{N₁,N₂},
                            ::Bool) where {N₁,N₂}
-    M = similarstoragetype(A, TC)
+    sst = sectorscalartype(sectortype(A))
+    Tnew = sst <: Real ? TC : VectorInterface.promote_add(TC, scalartype(A), sst)
+    M = similarstoragetype(A, Tnew)
```
Jutho: @lkdvos , if we can completely handle the complex `sectorscalartype` by simply modifying `tensoradd_type` and `tensorcontract_type` that would be my preferred option. That is what these functions are for, create a tensor that can support data of scalar type `TC` that will be obtained from `A` by doing some permutation. I think the fact that the data will be further scaled with complex scalar coefficients is an intrinsic part of our tensor type, and therefore this is where this should be handled. I just find it hard to reason about whether this covers all cases within the `@tensor` macro, but I think it should.

We can even have a separate promotion step of `TC` if we ever want to take more information about `sectorscalartype` into account (e.g. loss of precision etc), but for now I would keep it as simple as possible, e.g.:

```suggestion
    I = sectortype(A)
    M = similarstoragetype(A, sectorscalartype(I) <: Real ? TC : complex(TC))
```

src/tensors/tensoroperations.jl
```diff
@@ -113,9 +115,13 @@ function TO.tensorcontract_type(TC,
...
-    M = similarstoragetype(A, TC)
-    M == similarstoragetype(B, TC) ||
-        throw(ArgumentError("incompatible storage types:\n$(M) ≠ $(similarstoragetype(B, TC))"))
+    sst(x) = sectorscalartype(sectortype(x))
+    Tnew = (sst(A) <: Real && sst(B) <: Real) ? TC :
+           TO.promote_contract(TC, scalartype(A), sst(A),
+                               scalartype(B), sst(B))
+    M = similarstoragetype(A, Tnew)
+    M == similarstoragetype(B, Tnew) ||
+        throw(ArgumentError("incompatible storage types:\n$(M) ≠ $(similarstoragetype(B, Tnew))"))
```
Jutho, line 124:
```suggestion
    I = sectortype(A)
    M = similarstoragetype(A, sectorscalartype(I) <: Real ? TC : complex(TC))
    MB = similarstoragetype(B, sectorscalartype(I) <: Real ? TC : complex(TC))
    M == MB || throw(ArgumentError("incompatible storage types:\n$(M) ≠ $(MB)"))
```

src/tensors/tensoroperations.jl
```diff
@@ -113,9 +115,13 @@ function TO.tensorcontract_type(TC,
...
     spacetype(A) == spacetype(B) || throw(SpaceMismatch("incompatible space types"))
```
Jutho:
```suggestion
    spacetype(A) == spacetype(B) || throw(SpaceMismatch("incompatible space types"))
    I = sectortype(A)
    M = similarstoragetype(A, sectorscalartype(I) <: Real ? TC : complex(TC))
    MB = similarstoragetype(B, sectorscalartype(I) <: Real ? TC : complex(TC))
    M == MB || throw(ArgumentError("incompatible storage types:\n$(M) ≠ $(MB)"))
```

Jutho:
I left some comments and suggestions; the only part I want to revisit is the `elementary_trace` part in fusiontree manipulations (mostly to remind myself of how the implementation works). It is indeed very nice that such a small number of changes is sufficient to incorporate multifusion categories; even now about half of the changes are bug fixes that were needed (thanks!) even without the multifusion case. Very impressive work.

src/fusiontrees/iterator.jl
borisdevos: I found this from an error. However, I cannot find what I was doing to trigger this particular iteration... I will keep looking

src/fusiontrees/manipulations.jl
borisdevos: I use `unit` in line 789. For this particular case I'll put `f.coupled` as it's used just once.

src/fusiontrees/manipulations.jl
```diff
@@ -778,18 +781,19 @@ function elementary_trace(f::FusionTree{I,N}, i) where {I<:Sector,N}
         uncoupled_ = TupleTools.front(f.uncoupled)
         inner_ = TupleTools.front(f.innerlines)
         coupled_ = f.innerlines[end]
-        @assert coupled_ == dual(b)
+        @assert coupled_ == dual(b) # isn't this always true at this point?
```
Jutho: Yes, this was more of a security check, feel free to remove.

src/fusiontrees/manipulations.jl
```diff
@@ -778,18 +781,19 @@ function elementary_trace(f::FusionTree{I,N}, i) where {I<:Sector,N}
...
         isdual_ = TupleTools.front(f.isdual)
         vertices_ = TupleTools.front(f.vertices)
         f_ = FusionTree(uncoupled_, coupled_, isdual_, inner_, vertices_)
         fs = FusionTree((b,), b, (!f.isdual[1],), (), ())
-        for (f_′, coeff) in merge(fs, f_, one(I), 1)
-            f_′.innerlines[1] == one(I) || continue
+        unit = leftone(fs.coupled)
+        for (f_′, coeff) in merge(fs, f_, unit, 1) # coloring gets reversed here, should be the other unit
+            f_′.innerlines[1] == unit || continue # is this one valid?
```
Jutho: I think both instances of `unit` can be `leftone(b)`, which is probably the same as `leftone(fs.coupled)`, although I am not sure. `fs.coupled` itself should be `rightone(b)`, so can `leftone(rightone(b))` be different than `leftone(b)` ?

src/fusiontrees/manipulations.jl
```diff
@@ -778,18 +781,19 @@ function elementary_trace(f::FusionTree{I,N}, i) where {I<:Sector,N}
...
             uncoupled′ = Base.tail(Base.tail(f_′.uncoupled))
             isdual′ = Base.tail(Base.tail(f_′.isdual))
             inner′ = N <= 4 ? () : Base.tail(Base.tail(f_′.innerlines))
             vertices′ = N <= 3 ? () : Base.tail(Base.tail(f_′.vertices))
-            f′ = FusionTree(uncoupled′, one(I), isdual′, inner′, vertices′)
+            f′ = FusionTree(uncoupled′, unit, isdual′, inner′, vertices′) # and this one?
```
Jutho, line 795: Yes, that is the same `unit`, because this is the same coupled sector as `f_′.coupled`, i.e. `f′` is simply `f_′` but where the first two uncoupled sectors are removed (because they are traced away).

src/fusiontrees/manipulations.jl
```diff
@@ -767,7 +770,7 @@ function elementary_trace(f::FusionTree{I,N}, i) where {I<:Sector,N}
         return newtrees
     else # i == N
         if N == 2
-            f′ = FusionTree{I}((), one(I), (), (), ())
+            f′ = FusionTree{I}((), f.coupled, (), (), ()) # or leftone(f.uncoupled[1]) == rightone(f.uncoupled[2])
```
Jutho: I don't think this is right for reasons that I can explain on the blackboard. I think the coupled sector needs to be `leftone(b) = leftone(f.uncoupled[i]) = leftone(f.uncoupled[2])`, instead of `rightone(f.uncoupled[2])`. Maybe we can have `unit = leftone(b)` at the top of this `else # i == N` clause, and then use that `unit` also as the coupled sector of `f′` both here and below. I am not sure if even the special casing the `N==2` case is really necessary, but it helps a bit.

Jutho:
This looks almost complete to me, aside from one more change in `elementary_trace` and the formatting issues. It would be great to push this over the finish line and have it merged.

src/fusiontrees/manipulations.jl
borisdevos: Note to self so I don't forget in the future: `leftone(b)` is correct here because the `i==N` traces are done along the top of the fusion trees. You then want the unit to the left of the last uncoupled sector.

src/fusiontrees/manipulations.jl
borisdevos: It should indeed be `leftone(b)`. Also, `leftone(rightone(b))` is generally different: if `b` is some object in a module category I call $C_{ij}$, then `leftone(b)` is the unit in $C_{ii}$, while `leftone(rightone(b)) = one(rightone(b))` gives the unit in $C_{jj}$, a different fusion category.

Jutho:
@lkdvos, did you want to have a final review or did you have any pending comments before I merge this?

Jutho:
Thanks for the final fixes @borisdevos ; I'll merge.

PR #248, opened 2025-06-16
Default `spacetype` and `sectortype` implementations in type domain

ogauthe:
PEPSKit.jl overload `sectortype` and `spacetype` for many objects (see https://github.com/QuantumKitHub/PEPSKit.jl/pull/212/commits/af3f1409f436f0fd2fcf94501ba473f3e82a8d96 or https://github.com/QuantumKitHub/PEPSKit.jl/pull/218). Currently one has to repeat 4 lines of code for every struct in order to dispatch on type and use `sectortype(spacetype)`.

This PR proposes to set defaults in `TensorKit` such that a single line
```julia
spacetype(::Type{MyType}) = ...
```
is sufficient to use `sectortype(x::MyType)` and `spacetype(x::MyType)`.

I also defined a default for `field` as it is very similar. I kept the specializations for `AbstractTensorMap` to conserve easy access to the documentation, I do not know if there is a better way.

### DIFF
--- a/src/spaces/vectorspaces.jl
+++ b/src/spaces/vectorspaces.jl
@@ -289,8 +289,8 @@ spacetype(x) = spacetype(typeof(x))
 function spacetype(::Type{T}) where {T}
     throw(MethodError(spacetype, (T,)))
 end
-spacetype(S::Type{<:ElementarySpace}) = S
-spacetype(::Type{<:CompositeSpace{S}}) where {S} = S
+spacetype(::Type{E}) where {E<:ElementarySpace} = E
+spacetype(::Type{S}) where {E,S<:CompositeSpace{E}} = E
 
 # make ElementarySpace instances behave similar to ProductSpace instances
 blocksectors(V::ElementarySpace) = collect(sectors(V))


ogauthe:
I generalized the doc. Also generalized `sectortype` to allow `sectortype(SU2Irrep) == SU2Irrep`.

ogauthe:
> Yes this is indeed much more convenient. The only small comment is that I would streamline:
> 
> ```julia
> sectortype(::Type{S}) where {S<:Sector} = S
> spacetype(S::Type{<:ElementarySpace}) = S
> ```
> 
> so that they take the same form (both seem to do the job).

I found there are subtle differences in dispatch behavior between these 2 versions (see https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing). I now impose specialization.

lkdvos:
This is true, but often not relevant since these functions get inlined and constant folded most of the time, but I fully agree not relying on these implicit compiler details is annoying since it's hard to verify, so might as well be explicit about it

lkdvos (approved):

Issue #243, opened 2025-04-15
`flip(::DiagonalTensorMap, is)` returns `TensorMap`


lkdvos:
Currently, flipping the legs of a diagonal tensor loses that the diagonal information:

```julia
julia> d = DiagonalTensorMap{Float64}(undef, ℂ^3)
DiagonalTensorMap(ℂ^3 ← ℂ^3):
 0.0   ⋅                 ⋅ 
  ⋅   6.798626232e-314   ⋅ 
  ⋅    ⋅                0.0


julia> flip(d, 1)
TensorMap((ℂ^3)' ← ℂ^3):
 0.0  0.0               0.0
 0.0  6.798626232e-314  0.0
 0.0  0.0               0.0
```

Jutho:
Diagonal only exists if domain == codomain, so the problem is that only `flip(d, (1,2))` can preserve the `DiagonalTensorMap` structure.

Jutho:
`transpose(d)` could be used to flip the domain, but this might not have the same semantics in terms of preserving contractions.

lkdvos:
Oh, that's also true... I think we can work around this in this particular case by doing the `svd` in the "other direction", but this is definitely a bit of an annoying problem. We could consider checking that `flip(d, (1, 2))` contains either zero or both indices, and throw an error to tell the user to explicitly convert to `TensorMap` if they want a different result, but that also feels very opinionated and clunky...

lkdvos:
How about an additional `flip!(dst, src, I; inv)` method?

Jutho:
> Oh, that's also true... I think we can work around this in this particular case by doing the `svd` in the "other direction", but this is definitely a bit of an annoying problem. We could consider checking that `flip(d, (1, 2))` contains either zero or both indices, and throw an error to tell the user to explicitly convert to `TensorMap` if they want a different result, but that also feels very opinionated and clunky...

We could also support `DiagonalTensorMap` with isomorphic domain and codomain; I don't think there is a strong argument for having domain == codomain.

PR #249, opened 2025-06-21
DiagonalTensorMap performance and convenience specializations

lkdvos:
Introduce a fast path for singular/eigen value calculations, implement left and right in-place multiplication methods, and add convenience constructors for DiagonalTensorMap.

### DIFF
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -372,7 +372,6 @@ function LinearAlgebra.eigvals(d::DiagonalTensorMap)
     return SectorDict(c => LinearAlgebra.eigvals(b) for (c, b) in blocks(d))
 end
 
-
 function LinearAlgebra.cond(d::DiagonalTensorMap, p::Real=2)
     return LinearAlgebra.cond(Diagonal(d.data), p)
 end
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -368,8 +368,9 @@ end
 function LinearAlgebra.svdvals(d::DiagonalTensorMap)
     return SectorDict(c => LinearAlgebra.svdvals(b) for (c, b) in blocks(d))
 end
-function LinearAlgebra.svdvals!(d::DiagonalTensorMap)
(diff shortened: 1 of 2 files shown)


lkdvos:
Failures on nightly are unrelated, this should be good to go.

src/tensors/diagonal.jl
```diff
@@ -37,6 +37,16 @@ storagetype(::Type{<:DiagonalTensorMap{T,S,A}}) where {T,S,A<:DenseVector{T}} =
 
 Construct a `DiagonalTensorMap` with uninitialized data.
 """
+function DiagonalTensorMap{T}(::UndefInitializer, V::TensorMapSpace) where {T}
+    (numin(V) == numout(V) == 1 && domain(V) == codomain(V)) ||
+        throw(SpaceMismatch("DiagonalTensorMap requires a space with equal domain and codomain and 2 indices"))
+    return DiagonalTensorMap{T}(undef, domain(V))
+end
```
Jutho, line 44: I am wondering whether we should have this method? Do you use it? Looking at `LinearAlgebra.Diagonal`, it doesn't support an `undef` constructor, but with e.g. using `I` you can only do `Diagonal(I, 5)`, and not `Diagonal(I, (5,5))` or `Diagonal(I, 5, 5)`. Same applies to the method below, although that is more subtle because it doesn't have a counterpart to the `Diagonal` matrix case.

lkdvos: I don't remember actually. I ported over some of the changes from #230 since these are orthogonal to that, in an attempt to make that PR slightly smaller, but I can't really find anymore why I had this. I have no strong opinions here, I'm happy to remove it, although if we ever want to support isomorphic domain and codomain (#243) we would need this anyways.

Jutho: Yes that is true, I guess there are use cases for `DiagonalTensorMap` with two incoming or two outgoing arrows in the context of bond tensors for PEPS. So let's leave it like this then. The semantics are anyway different than those of `LinearAlgebra.Diagonal`, which doesn't have an `undef` constructor anyway.

lkdvos: I guess I should add some tests to at least have some coverage of this though...

PR #250, opened 2025-07-01
small type stability fix

Jutho:


lkdvos:
I'd be okay with disabling the AD tests on nightly altogether to be honest, it's very often that this causes issues

Jutho:
Do you know how to easily check whether the version it is running on is nightly or a stable version?

Jutho:
Ok, new attempt, now disabling AD tests at nightly altogether. I am not ready to give up nightly tests alltogether.

PR #254, opened 2025-07-09
Refactor repacking of transformer structure

lkdvos:
This is a small improvement based on @Jutho's suggestion that avoids an intermediate allocation.


lkdvos:
Looking at the stacktraces I think this is indeed the issue: `rmul!(stridedarray, 1)` goes through `copyto!` which checks for aliasing.

Jutho:
Strided.jl v2.3.2 will be tagged/registered soon, so we can then rerun the tests.

Issue #255, opened 2025-07-12
Contraction error after v0.14.7


Qiaoyi-Li:
Hi Jutho and Lukas,

I use TensorKit.jl to handle non-Abelian symmetries in my MPS package. After updating TensorKit to version 0.14.7, some tensor contraction processes that previously worked now return errors. The following is an instance to reproduce the issue:
```julia
using TensorKit

pspace = Rep[U₁×SU₂]((-1, 0) => 1, (0, 1 / 2) => 1)
aspace = Rep[U₁×SU₂]((1, 1 / 2) => 1)

A = ones(pspace , pspace ⊗ aspace)
B = isometry(aspace, flip(aspace))
@tensor C[a; c d] := A'[a b c] * B'[d b]
```
The error message:
```julia
ERROR: TypeError: in typeassert, expected Int64, got a value of type Nothing
Stacktrace:
  [1] TensorKit.GenericTreeTransformer(transform::TensorKit.var"#fusiontreetransform#238"{…}, p::Tuple{…}, Vdst::TensorMapSpace{…}, Vsrc::TensorMapSpace{…})
    @ TensorKit ~/.julia/packages/TensorKit/fQn8E/src/tensors/treetransformers.jl:92
  [2] TensorKit.TreeTransformer(transform::Function, p::Tuple{…}, Vdst::TensorMapSpace{…}, Vsrc::TensorMapSpace{…})
    @ TensorKit ~/.julia/packages/TensorKit/fQn8E/src/tensors/treetransformers.jl:157
  [3] _treepermuter
    @ ~/.julia/packages/TensorKit/fQn8E/src/tensors/treetransformers.jl:187 [inlined]
  [4] #236
    @ ~/.julia/packages/TensorKit/fQn8E/src/auxiliary/caches.jl:108 [inlined]
  [5] get!(default::TensorKit.var"#236#241"{…}, lru::LRUCache.LRU{…}, key::Tuple{…})
    @ LRUCache ~/.julia/packages/LRUCache/ZH7qB/src/LRUCache.jl:169
  [6] treepermuter(Vdst::TensorMapSpace{…}, Vsrc::TensorMapSpace{…}, p::Tuple{…}, ::TensorKit.GlobalLRUCache)
    @ TensorKit ./none:0
  [7] treepermuter
    @ ./none:0 [inlined]
  [8] treepermuter
[...]

PR #256, opened 2025-07-12
[Fix] Transformers for non-abelian non-selfadjoint sectors

lkdvos:
As brought up in #255 , for sectors that are non-abelian and not self-adjoint the logic for locating the fusiontrees was being handled incorrectly.
Here, I've fixed that, and added a set of spaces that have this property to the testsuite.
An alternative would be to re-enable the SU3 tests, but I think this testset is also useful by itself.


src/tensors/treetransformers.jl
```diff
@@ -116,8 +116,8 @@ function GenericTreeTransformer(transform, p, Vdst, Vsrc)
 
     @debug("TreeTransformer for $Vsrc to $Vdst via $p",
            nblocks = length(data),
-           sz_median = size(data[end ÷ 2][1], 1),
-           sz_max = size(data[1][1], 1),
+           sz_median = size(data[cld(end, 2)][1], 1),
+           sz_max = size(data[cld(end, 2)][1], 1),
```
Jutho: Are `sz_median` and `sz_max` now the same? Do you just want `data[end][1]`?

lkdvos: Oh you are right, I should have only corrected the `sz_median`. Should be resolved now.

Jutho (approved):
Looks good to merge; do you also bump the patch version?

Issue #245, opened 2025-05-08
`foldright` assumptions for `UniqueFusion`


lkdvos:
As brought up in #227, there is an inconsistency with the interaction between `FusionStyle(I) isa UniqueFusion` and the implementation of `foldright`.

After digging a bit deeper, I was able to locate the source of the inconsistencies:

```julia
using TensorKit, TensorKitSectors
using CategoryData

f₁ = FusionTree{Semion}((2, 2), 1, (true, false), ())
f₂ = FusionTree{Semion}((1,), 1, (true,), ())

TensorKitSectors.FusionStyle(::Type{Semion}) = TensorKitSectors.UniqueFusion()
only(values(TensorKit.foldright(f₁, f₂))) # 1

TensorKitSectors.FusionStyle(::Type{Semion}) = TensorKitSectors.SimpleFusion()
only(values(TensorKit.foldright(f₁, f₂))) # -1
```

Even though the fusion rules are the same as for Z2Irrep, the non-trivial Fsymbols make it such that the resulting coefficients differ by a minus sign.
Basically, bypassing the `insert` calls through https://github.com/Jutho/TensorKit.jl/blob/0c73bd2ff03dee9b6eb597863d0dbdcefce4ab3b/src/fusiontrees/manipulations.jl#L354 is not allowed.

There are some options to resolving this: we could clarify that `UniqueFusion` really assumes trivial `Fsymbol`s, therefore making the code correct again. This is slightly unfortunate given the name of `UniqueFusion`, but would definitely be an easy fix that also doesn't alter existing performance.

Alternatively, we can "fix" `foldright` by not assuming trivial Fsymbols, which might deteriorate performance for the actual trivial cases, but this might not make a large difference? This more or less reverts to getting rid of that special case.

Finally, we can consider having more traits to distinguish further between these different kinds of behavior.

I'm not sure what you think @Jutho, but I would lean slightly towards the first solution, given that it is very likely that there are more parts of the code that make the assumption that `UniqueFusion` means we can assume trivial Fsymbols. I would also be okay with expanding the set of traits, but that takes slightly more consideration.

borisdevos:
Fixed by #261

PR #260, opened 2025-07-22
Change frobeniusschur factor

lkdvos:
This is a small change that fixes some issues encountered with transposing and adjoints not commuting. We've traced the origin to being that while we aren't assuming the Frobenius Schur factor is +-1, we have no checks for that since for all our sectors this is true.

There is definitely a `conj` missing, and the one I added here seems to resolve the issue, but I absolutely did not think through whether or not it should be here, or here instead:

https://github.com/Jutho/TensorKit.jl/blob/ef03d19f8add057074c8103470dc15d759e8cd43/src/fusiontrees/manipulations.jl#L309

### DIFF
--- a/src/fusiontrees/manipulations.jl
+++ b/src/fusiontrees/manipulations.jl
@@ -347,7 +347,7 @@ function foldright(f₁::FusionTree{I,N₁}, f₂::FusionTree{I,N₂}) where {I<
     isduala = f₁.isdual[1]
     factor = sqrtdim(a)
     if !isduala
-        factor *= frobeniusschur(a)
+        factor *= conj(frobeniusschur(a))
     end
     c1 = dual(a)
     c2 = f₁.coupled


Jutho:
I am a bit confused; how did you find this one? This is unrelated to #245 , right? Since that would not be fixed by this extra `conj`? I think it is fair to assume `frobeniusschur` factors to be ±1. They need to have this value for `a == dual(a)`, and all other sectors one can in principle always choose a gauge such that they are one. I believe this is mentioned in the documentation, but it is true that we do not clearly state this or test this.

lalooten:
For Z_n, the 3-cocycles are given here: https://mathoverflow.net/a/157012, might be worth it to just implement this for general $Z_n$

lalooten:
For general abelian groups with general 3-cocycles, this is not true; there is a 1-1 correspondence between braided fusion categories $\text{Vec}_G^\omega$ (which of course requires G to be abelian) and something called pre-metric groups (finite abelian group with a quadratic form). One can construct the associator and braid matrices from the quadratic form, but this does not exhaust all possible associators meaning that not all choices of $\omega$ will allow for a braiding.

For $Z_N$, I think it is true that for any choice of 3-cocycle you can find a braiding. I had a quick look and I couldn't find a reference that gives explicit expressions for this case, but there are general expressions given a quadratic form. It's probably not too hard either to guess the solution by looking at the hexagon equations, and if you really need it for something I'm happy to look into it, but it takes a bit of work.

Jutho:
No that is fine, I think I can go with `NoBraiding` for now, or maybe add a braiding for the trivial cocycle case.

PR #261, opened 2025-07-31
Vectorize fusiontree manipulations

lkdvos:
Given the speed-up of the permutations, the bottleneck has now migrated to the actual computation of the treetransformers. (mostly for tensors with many legs again)
@ogauthe has reported cases where the first run takes ~2000seconds, while the second run is ~5seconds.

While I haven't really benchmarked or profiled myself yet (shame on me!), I immediately remembered that we are not really composing the fusion tree manipulations in a very optimal way, where we are effectively manually writing out the matrix multiplications on the one hand, but also the recursive implementation means that we are recomputing a lot of transformations that we would otherwise already have encountered. This never really showed up since these computations are somewhat fast and cached, but for large amount of fusiontrees these loops become prohibitively expensive.

This PR aims to address these issues in a very similar fashion as the previous optimization run: we consider the set of all fusion trees with equal uncoupled charges, and act on these as a block. If my back-of-the-envelope calculations are correct, if there are `N` such fusion trees and we are doing `L` substeps in the manipulation, we were previously doing $\sim N^L$ operations due to the recursive nature, while now this is $\sim L N^3$. 

This PR now contains the following changes:
- Some clean up of calling signatures (`braid` had some inconsistencies in its argument orders, `Index2Tuple` etc)
- Use `fusiontensor` instead of `convert(Array, ::FusionTreePair)` to avoid type piracy
- "vectorized" manipulations of non-`UniqueFusion` fusionstyles to avoid the exponential scaling in number of chained manipulations.
- vectorized indexmanipulations beyond `TensorMap` specializations, approaching the entire "indexmanipulations" kernels in a more unified way
- Updated the fusiontree tests to reflect these changes


src/fusiontrees/fusiontrees.jl
```diff
@@ -199,8 +201,7 @@ function Base.convert(A::Type{<:AbstractArray}, f::FusionTree{I,N}) where {I,N}
 end
 
 # TODO: is this piracy?
```
Jutho: Yes I think this is bordering on type piracy, so we should probably just give this method a dedicated name and then simply call that one.

Jutho:
I think a first general design remark is that now all the manipulations on the `OuterTreeIterator` (which is not an iterator 😄 ) have in them the lines:
```julia
    trees_src = fusiontrees(fs_src)
    trees_dst = fusiontrees(fs_dst)
    indexmap = Dict(f => ind for (ind, f) in enumerate(trees_dst))
```

If you know start composing elementary operations to make more complex ones (e.g. `bendleft/right` to implement `repartition`), there will be at least two `fusiontrees` calls on the same `OuterTreeIterator` object. In every step the previous destination is the new source for the following operation in the composition list. The `fusiontrees` method goes through `blocksectors` and the `fusiontrees(uncoupled_sectors, blocksector)`, which should be pretty efficient, but nonetheless have some overhead. 

It might thus be a better idea to store the blocksectors and associated fusion tree pairs (and there index mapping) within the `OuterTreeIterator` object. The initial data can be extracted from the data the tensor has, and then throughout the sequence of different operations, the old destination becoming new source would be able to reuse the same information. 

Of course, this way, the `OuterTreeIterator` would become a lightweight version of `FusionBlockStructure`, so maybe there is some way to recycle some functionality or even redesing `FusionBlockStructure`.

Happy to discuss when you have time.

kshyatt:
BTW does @ogauthe have a MWE of what's causing the slowdown that could be shared, perhaps to run as part of a perf regression testsuite 👀 ?

lkdvos:
I tried to finish this before the holiday, but I guess I was being a bit too ambitious...
I do think the overall idea of what I have in mind is more or less in place, and I did manage to reduce the code duplication, but there are definitely still some parts that need/could do with work.
First off, it does look like I broke something along the way, so needs some debugging.
Secondly, the fusiontensor tests are a bit painful to handle properly. There are some `isunitary` checks in there that are not fair, and should be weighted with some quantum dimensions (I think?), and the `if FusionStyle` distinctions do feel a bit clunky there...
Finally, the traces currently don't fit into the same style at all, which also makes testing them a bit hacky.

On the positive side though, I do think that I managed to unify most of the indexmanipulation code now, where also `AbstractTensorMap` can benefit from the same non-abelian speedups as `TensorMap` has.
If anyone feels like taking a stab at this, definitely be my guest!


Warning to not pull in the changed code of the last couple commits if you want a working version though 🙃

lkdvos:
OK, I think I managed to get this PR more or less in its final stages, with a few remaining comments.
The major thing is that I had to somewhat delete quite a large part of the fusiontree test suite. The problem is that we were testing quite a few functions that are defined on single fusion trees, that I didn't really bother also implementing for the `FusionTreeBlock`s.
As these are not really used I think this is fair, since it would basically require me to add a bunch of new functionality, just to then be able to test it, without this actually being used in the remainder of the package.
Therefore, I mostly kept the "double fusion tree" tests, and made sure to also test the braiding for that, since that was previously missing.

Note that I also updated the initial PR description to reflect all that got changed in this PR.

@Jutho, it would be great if you could share what you think about this, since these are some substantial changes, also surrounding what features we are offering.
In particular, technically we now have less functionality for working with individual fusion trees.

@borisdevos I have to admit that I am not sure how all of this interacts with the `GenericUnit` changes either. It seems to be working, but the test suite might really not be covering all that much and I'm not sure where to start, so feedback and input would be more than welcome.

src/tensors/indexmanipulations.jl
```diff
@@ -542,93 +533,136 @@ end
...
 end
+function add_kernel_nonthreaded!(::FusionStyle, tdst, tsrc, p, transformer, α, β, backend...)
+    # preallocate buffers
+    buffers = allocate_buffers(tdst, tsrc, transformer)
+
+    for src in fusionblocks(tsrc)
+        if length(src) == 1
+            _add_transform_single!(tdst, tsrc, p, src, transformer, α, β, backend...)
+        else
+            _add_transform_multi!(tdst, tsrc, p, src, transformer, α, β, backend...)
```
ogauthe:
```suggestion
            _add_transform_multi!(tdst, tsrc, p, src, buffers, transformer, α, β, backend...)
```

src/tensors/indexmanipulations.jl
```diff
@@ -680,6 +732,36 @@ function _add_transform_single!(
...
+    rows, cols = size(U)
+    sz_src = size(tsrc[first(fusiontrees(src))...])
+    blocksize = prod(sz_src)
+    matsize = (
+        prod(TupleTools.getindices(sz_src, codomainind(tsrc))),
+        prod(TupleTools.getindices(sz_src, domainind(tsrc))),
+    )
+
+    # Filling up a buffer with contiguous data
+    buffer_src = StridedView(buffers2, (blocksize, cols), (1, blocksize), 0)
```
ogauthe:
```suggestion
    buffer_src = StridedView(buffer2, (blocksize, cols), (1, blocksize), 0)
```

src/fusiontrees/fusiontrees.jl
```diff
@@ -156,52 +258,57 @@ function fusiontreetype(::Type{I}, N::Int) where {I <: Sector}
...
 # converting to actual array
-function Base.convert(A::Type{<:AbstractArray}, f::FusionTree{I, 0}) where {I}
-    X = convert(A, fusiontensor(unit(I), unit(I), unit(I)))[1, 1, :]
-    return X
-end
-function Base.convert(A::Type{<:AbstractArray}, f::FusionTree{I, 1}) where {I}
+Base.convert(A::Type{<:AbstractArray}, f::FusionTree) = convert(A, fusiontensor(f))
+# TODO: is this piracy?
+Base.convert(A::Type{<:AbstractArray}, (f₁, f₂)::FusionTreePair) =
+    convert(A, fusiontensor((f₁, f₂)))
```
Jutho, line 276: I would say: yes this is type piracy

src/fusiontrees/fusiontrees.jl
```diff
@@ -105,15 +105,106 @@ function FusionTree(
...
+        sort!(collect(filter(isunit, ⊗(uncoupled[2]...))))
+    elseif N₂ == 0
+        sort!(collect(filter(isunit, ⊗(uncoupled[1]...))))
+    else
+        sort!(collect(intersect(⊗(uncoupled[1]...), ⊗(uncoupled[2]...))))
+    end
+
+    for c in cs
+        for f₁ in fusiontrees(uncoupled[1], c, isdual[1]),
+                f₂ in fusiontrees(uncoupled[2], c, isdual[2])
```
Jutho: This makes `f₂` the fastest changing "index" if I am correct. Is that the order we want the trees to be in? Does it matter?

src/spaces/homspace.jl
```diff
@@ -138,6 +138,25 @@ Return the fusiontrees corresponding to all valid fusion channels of a given `Ho
...
+"""
+function fusionblocks(W::HomSpace)
+    I = sectortype(W)
+    N₁, N₂ = numout(W), numin(W)
+    isdual_src = (map(isdual, codomain(W).spaces), map(isdual, domain(W).spaces))
+    fblocks = Vector{FusionTreeBlock{I, N₁, N₂, fusiontreetype(I, N₁, N₂)}}()
+    for cod_uncoupled_src in sectors(codomain(W)), dom_uncoupled_src in sectors(domain(W))
+        fs_src = FusionTreeBlock{I}((cod_uncoupled_src, dom_uncoupled_src), isdual_src)
+        trees_src = fusiontrees(fs_src)
+        isempty(trees_src) || push!(fblocks, fs_src)
```
Jutho: Since `isempty` on `FusionBlock` was defined:

```suggestion
        isempty(fs_src) || push!(fblocks, fs_src)
```

src/tensors/braidingtensor.jl
```diff
@@ -81,21 +97,17 @@ end
...
+    d = (dims(codomain(b), f₁.uncoupled)..., dims(domain(b), f₂.uncoupled)...)
+    n1 = d[1] * d[2]
+    n2 = d[3] * d[4]
+    data = sreshape(StridedView(Matrix{eltype(b)}(undef, n1, n2)), d)
+    fill!(data, zero(eltype(b)))
+
+    r = _braiding_factor(f₁, f₂, b.adjoint)
+    isnothing(r) || @inbounds for i in axes(data, 1), j in axes(data, 2)
+        data[i, j, j, i] = r
     end
```
Jutho, line 126: I don't like the `||` short-circuit construction too much for such a complicated expression

```suggestion
    if !isnothing(r)
        for @inbounds for i in axes(data, 1), j in axes(data, 2)
            data[i, j, j, i] = r
        end
    end
```

Jutho: Something went wrong with that suggestion.

src/tensors/indexmanipulations.jl
```diff
@@ -542,93 +533,136 @@ end
...
 end
+function add_kernel_nonthreaded!(::FusionStyle, tdst, tsrc, p, transformer, α, β, backend...)
+    # preallocate buffers
+    buffers = allocate_buffers(tdst, tsrc, transformer)
+
+    for src in fusionblocks(tsrc)
+        if length(src) == 1
+            _add_transform_single!(tdst, tsrc, p, src, transformer, α, β, backend...)
+        else
+            _add_transform_multi!(tdst, tsrc, p, src, buffers, transformer, α, β, backend...)
```
ogauthe:
```suggestion
            _add_transform_multi!(tdst, tsrc, p, src, transformer, buffers, α, β, backend...)
```

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,273 @@
...
+end
+function merge(f₁::FusionTree{I, N₁}, f₂::FusionTree{I, N₂}, c::I, μ) where {I, N₁, N₂}
+    if !(c in f₁.coupled ⊗ f₂.coupled)
+        throw(SectorMismatch("cannot fuse sectors $(f₁.coupled) and $(f₂.coupled) to $c"))
+    end
+    if μ > Nsymbol(f₁.coupled, f₂.coupled, c)
+        throw(ArgumentError("invalid fusion vertex label $μ"))
+    end
+    f₀ = FusionTree{I}((f₁.coupled, f₂.coupled), c, (false, false), (), (μ,))
+    f, coeff = first(insertat(f₀, 1, f₁)) # takes fast path, single output
```
Jutho:
```suggestion
    f, coeff = only(insertat(f₀, 1, f₁)) # takes fast path, single output
```

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,273 @@
...
+    @assert coeff == one(coeff)
```
Jutho: Not sure if we need to keep this.

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,273 @@
...
+    return insertat(f, N₁ + 1, f₂)
+end
+function merge(f₁::FusionTree{I, 0}, f₂::FusionTree{I, 0}, c::I, μ) where {I}
+    Nsymbol(f₁.coupled, f₂.coupled, c) == μ == 1 ||
+        throw(SectorMismatch("cannot fuse sectors $(f₁.coupled) and $(f₂.coupled) to $c"))
+    return fusiontreedict(I)(f₁ => Fsymbol(c, c, c, c, c, c)[1, 1, 1, 1])
+end
+
+# flip a duality flag of a fusion tree
```
Jutho: `flip` really needs some comments/doc string to remember the logic. I 'll try to come up with something after rereading the PR that introduced it.

src/fusiontrees/duality_manipulations.jl
```diff
@@ -0,0 +1,792 @@
...
+
+```
+    ╰─┬─╯ |  | |   ╰─┬─╯ |  |  |
+      ╰─┬─╯  | |     ╰─┬─╯  |  |
+        ╰ ⋯ ┬╯ |       ╰ ⋯ ┬╯  |
+            |  | →         ╰─┬─╯
+        ╭ ⋯ ┴╮ |         ╭ ⋯ ╯
+      ╭─┴─╮  | |       ╭─┴─╮
+    ╭─┴─╮ |  ╰─╯     ╭─┴─╮ |
+```
```
Jutho, line 26: That's a nice drawing.

src/fusiontrees/basic_manipulations.jl
Jutho: Also, I don't think this is a very basic manipulation, as it involves both duality and braiding (twists), so not sure it is really in its place here.

Jutho: Ok, I am going to leave this as a TODO and think about it later. I forgot what the original constraints and design considerations were for `flip`.

src/fusiontrees/duality_manipulations.jl
```diff
@@ -0,0 +1,792 @@
...
+        (src.uncoupled[2]..., dual(src.uncoupled[1][end])),
+    )
+    isdual_dst = (
+        TupleTools.front(src.isdual[1]),
+        (src.isdual[2]..., !(src.isdual[1][end])),
+    )
+    I = sectortype(src)
+    N₁ = numout(src)
+    N₂ = numin(src)
+    @assert N₁ > 0
```
Jutho:
```suggestion
    I = sectortype(src)
    N₁ = numout(src)
    N₂ = numin(src)
    @assert N₁ > 0
    uncoupled_dst = (
        TupleTools.front(src.uncoupled[1]),
        (src.uncoupled[2]..., dual(src.uncoupled[1][N₁])),
    )
    isdual_dst = (
        TupleTools.front(src.isdual[1]),
        (src.isdual[2]..., !(src.isdual[1][N₁])),
    )
```

src/fusiontrees/duality_manipulations.jl
```diff
@@ -0,0 +1,792 @@
...
+            else
+                cset = ⊗(Base.tail(f₁.uncoupled)...)
+            end
+            for c in c1 ⊗ c2
+                c ∈ cset || continue
+                for μ in 1:Nsymbol(c1, c2, c)
+                    fc = FusionTree((c1, c2), c, (!isduala, false), (), (μ,))
+                    fr_coeffs = insertat(fc, 2, f₂)
+                    for (fl′, coeff1) in insertat(fc, 2, f₁)
+                        N₁ > 1 && !isone(fl′.innerlines[1]) && continue
```
borisdevos:
```suggestion
                        N₁ > 1 && !isunit(fl′.innerlines[1]) && continue
```
What's the reason why `foldleft` for the fusion tree block can't use the `foldright` results?

lkdvos: I don't think there is a fundamental reason why it couldn't, this more or less gradually came about.

I started these changes for performance reasons (when it wasn't vectorized yet), and found some cases where the bookkeeping that is required for mapping one to the other was not completely insignificant (but also not that bad), so I just changed it.

However, looking at it now, doing it on a full block anyways requires some non-trivial code since you either have to construct all the fusiontrees from scratch again, or swap the pairs and make sure they are correctly sorted, and then make sure the permutation is correctly passed on to the coefficient matrix.

TLDR this was easier in my head and I didn't think too hard to try and reduce "code duplication", because it is also slightly faster

src/fusiontrees/duality_manipulations.jl
```diff
@@ -0,0 +1,792 @@
...
+sectors to a set of outgoing uncoupled sectors, represented using the individual trees of
+outgoing (`f₁`) and incoming sectors (`f₂`) respectively (with identical coupled sector
+`f₁.coupled == f₂.coupled`). Computes new trees and corresponding coefficients obtained from
+repartitioning the tree by bending incoming to outgoing sectors (or vice versa) in order to
+have `N` outgoing sectors.
+"""
+@inline function repartition(src::Union{FusionTreePair, FusionTreeBlock}, N::Int)
+    @assert 0 <= N <= numind(src)
+    return repartition(src, Val(N))
+end
```
Jutho, line 463: I don't understand how the current implementation works. Wasn't `N` supposed to  be the number of outgoing sectors? Here, it seems like `N` is used in the implementation as the number of actual bends that need to happen.

I.e. if `N₁ == N`, repartition shouldn't be doing anything (return the input with coefficient 1).

src/fusiontrees/basic_manipulations.jl
```diff
@@ -2,211 +2,529 @@
...
+        c = uncoupled[k + 1]
+        e, d, ν = VertexGetter(k + 1)(long)
+        b, e′, κ = VertexGetter(k)(short)
+        F = Fsymbol(a, b, c, d, e, e′)
+        if FusionStyle(I) isa MultiplicityFreeFusion
+            coeff *= F
+        else
+            if k == 2
+                μ = long.vertices[1]
+                coeff = transpose(view(F, μ:μ, ν, κ, :)) * coeff
```
borisdevos, line 160: Why is this a transpose and not adjoint?

borisdevos: I think I've made sense of it by looking at the below functions

src/fusiontrees/basic_manipulations.jl
```diff
@@ -2,211 +2,529 @@
...
+which is attached via some coupled sector `b` to the final fusion vertex. The second array contains
+the corresponding expansion coefficients, either as scalars (if `FusionStyle(I) isa MultiplicityFreeFusion`)
+or as vectors of length `Nsymbol(a, b, c)`, representing the different coefficients associated
+with the different vertex labels `λ` of the topmost vertex.
+
+See also [`multi_Fmove_inv`](@ref), [`multi_associator`](@ref).
+"""
+function multi_Fmove(f::FusionTree{I, N}) where {I, N}
+    if FusionStyle(I) isa UniqueFusion
+        coupled = N == 1 ? rightunit(f.uncoupled[1]) : (N == 2:f.uncoupled[2]:f.innerlines[end])
```
borisdevos: My code suggestion button has disappeared for some reason. The second ternary condition has a typo.

src/fusiontrees/duality_manipulations.jl
```diff
@@ -0,0 +1,812 @@
...
+        end
+        ex = quote
+            dst, U = $f(src)
+            $ex_rep
+            return dst => U
+        end
+    end
+    return ex
+end
+@generated function repartition(src::Union{FusionTreePair, FusionTreeBlock}, ::Val{N}) where {N}
```
Jutho: Do we need a generated function for this? In principle we can compute the output type with `fusiontreetype`. Is that not sufficient?

src/tensors/treetransformers.jl
Jutho: I think I would prefer not having an underscore here. I know we are not very consistent with this, but I don't think this type constant is any less or more internal than e.g. `GenericTreeTransformer`.

src/tensors/treetransformers.jl
```diff
@@ -62,65 +62,82 @@ function GenericTreeTransformer(transform, p, Vdst, Vsrc)
...
+    if nthreads > 1
+        counter = Threads.Atomic{Int}(1)
+        Threads.@sync for _ in 1:min(nthreads, nblocks)
+            Threads.@spawn begin
+                while true
+                    local_counter = Threads.atomic_add!(counter, 1)
+                    local_counter > nblocks && break
+                    fs_src = fblocks[local_counter]
+                    fs_dst, U = transform(fs_src)
+                    matrix = copy(transpose(U)) # TODO: should we avoid this
```
Jutho, line 86: Can't we just apply the transpose either in constructing `U` or when applying `U`?

src/fusiontrees/duality_manipulations.jl
```diff
@@ -0,0 +1,812 @@
...
+
+return dst, U
+```
+=#
+function _repartition_body(N)
+    if N == 0
+        ex = quote
+            T = sectorscalartype(sectortype(src))
+            if FusionStyle(src) === UniqueFusion()
+                return src => one(T)
```
Jutho, line 485: This might also amount to a shortcut for `UniqueFusion` that is not really valid if the Fsymbols are nontrivial.

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,735 @@
...
+end
+@inline function (vg::VertexGetter)(f::FusionTree)
+    k = vg.k
+    k <= length(f) ||
+        throw(ArgumentError(lazy"k = $k exceeds number of uncoupled legs $(length(f))"))
+    N = length(f)
+    l = (k == 2) ? f.uncoupled[1] : f.innerlines[k - 2]
+    r = (k == N) ? f.coupled : f.innerlines[k - 1]
+    return l, r, f.vertices[k - 1]
+end
```
lkdvos, line 112: Is there a reason to not have this as a function instead?
`vertexgetter(f::FusionTree, k::Int)` seems a bit easier, and `Base.Fix2` can still be used if you need a callable struct?

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,735 @@
...
+    if FusionStyle(I) isa UniqueFusion
+        coupled = N == 1 ? rightunit(f.uncoupled[1]) : (N == 2 ? f.uncoupled[2] : f.innerlines[end])
+        f′ = FusionTree{I}(Base.tail(f.uncoupled), coupled, Base.tail(f.isdual))
+        return (f′,), (multi_associator(f, f′),)
+    end
+
+    u = rightunit(f.coupled)
+    T = typeof(Fsymbol(u, u, u, u, u, u)[1, 1, 1, 1])
+    # sectorscalartype(I) may be different if there is also braiding
+    # TODO: consider using _Fscalartype?
```
lkdvos: I would be happy to just promote `Fscalartype` to an actual function in TensorKitSectors, if we need this we should just add it?
Let me know if you agree, then I'll just go ahead and make these changes.

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,735 @@
...
+                F = F_current
+                if FusionStyle(I) isa MultiplicityFreeFusion
+                    coeffs[i] *= F
+                else
+                    coeffs[i] = transpose(view(F, :, ν, κ, :)) * coeffs[i]
+                end
+            end
+        end
+        # TODO: it would be more uniform to return this as a dictionary
+        # Would that extra step create significant extra overhead?
```
lkdvos, line 325: In the grand scheme of things, from my testing I don't think so. It does obviously cause some allocations, but I would still opt for consistency here first, if need be we can revisit this later. If we start worrying about these things, we also should start investigating reusing the intermediate arrays of coefficients, and I don't think we are at that stage just yet

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,735 @@
...
+
+"""
+    insertat(f::FusionTree{I, N₁}, i::Int, f₂::FusionTree{I, N₂})
+    -> <:AbstractDict{<:FusionTree{I, N₁+N₂-1}, <:Number}
+
+Attach a fusion tree `f₂` to the uncoupled leg `i` of the fusion tree `f₁` and bring it
+into a linear combination of fusion trees in standard form. This requires that
+`f₂.coupled == f₁.uncoupled[i]` and `f₁.isdual[i] == false`.
+"""
+@inline function insertat(f₁::FusionTree{I, N₁}, i, f₂::FusionTree{I, N₂}) where {I, N₁, N₂}
```
lkdvos: Do we need this inline? I think I care more about compile times than actual maximal performance here, and if this is a type stability issue I would rather work with `Val(i)` instead of inlining such a large function

src/fusiontrees/fusiontrees.jl
lkdvos: It doesn't matter at first glance, but I'll change this, since I think it means we are extracting the data out of the tensors in a slightly less contiguous way than we could be (although I'm not sure if this actually matters).
In any case, it's better to remain consistent.

src/fusiontrees/fusiontrees.jl
lkdvos: Given that we used to have this and I don't feel like accidentally breaking code because of it, I vote to leave this in for now but deprecate it. The correct call should be `fusiontensor`, so that should resolve the type piracy issue

src/tensors/treetransformers.jl
lkdvos: Absolutely, the main issue is that I was trying to figure out which direction makes the most sense, and then broke my brain because I didn't know if the coefficients are supposed to denote the transformation on the fusion trees or the transformation on the data, and I somehow thought that this would alter the most natural rows/cols order. In any case the `matrix` here has to be in this order for the `add_transform!` implementation to be optimal (I think, unless BLAS is actually faster to multiply `A * B^T`, but I thought that effect is completely negligeable?) so we have to keep that here.

src/fusiontrees/duality_manipulations.jl
lkdvos: I think I experimented for a little bit (but not a lot) to try and get this type stable, and this wasn't too hard to write.
Previously it was a recursive definition with `Val` constructions as well, do you think that is friendlier on the compiler?

src/fusiontrees/duality_manipulations.jl
lkdvos: I don't think this is true, this is the trivial case where no repartitioning is actually happening

src/fusiontrees/duality_manipulations.jl
lkdvos: There's a subtle change of `N` in the actual call to the generated body below, but probably it is better if I at least give the variable a different name?

src/fusiontrees/duality_manipulations.jl
Jutho: Not sure, at some point the core Julia developers started arguing very much against using generated functions, and argued for replacing some of these patterns with recursion and regular inference. I don't know the precise reasons, nor whether if this still applies and whether this applies to something like this.

src/tensors/treetransformers.jl
Jutho: Generally, I would say, if you apply a linear map, you want that

(linear map on domain basis_i) = sum_{j} A^j_i codomain basis_j 

where j is row and i is col.

If you then have a general domain vector v^i basis_i, it is transformed into a codomain vector w^j basis_j with w^j = A^j_i v^i (i.e. w = A * v)

I guess the problem is that now you have the data stored with the expansion over the different fusion trees along the columns and the rows being the extra direction coming from the sizes of the individual sectors (i.e. the blocksize). Hence, instead of a single v, you have a collection of v's stored in a matrix V, but whereas you would normally expect that the columns correspond to the different vs (leading to W = A * V), in this case I think the vectors are layed out along the rows, so that you need W = V * transpose(A).

However, I just checked and I don't think there is any noticable difference between doing V * A and V * transpose(A), so I would keep the data layout but do the transpose in the matrix multiplication call?

src/fusiontrees/basic_manipulations.jl
Jutho: I needed it very often (especially during experimenting), that having to write Base.Fix2 everytime would have been a nuisance. I also don't know what the official policy is about using `Base.Fix1` and `Base.Fix2`, as they are not exported. However, it seems Julia 1.12 received a generalization to `Base.Fix{N}`, so I guess many people are using it.

src/fusiontrees/basic_manipulations.jl
Jutho: Yes I think it makes sense to have these names. Rather than `Fscalartype` and `Rscalartype`, I was perhaps thinking about `fusionscalartype` and `braidingscalartype`. That's a bit less explicit, but maybe leads to a nicer documentation: "use this scalar type when doing fusion operations".

src/fusiontrees/basic_manipulations.jl
Jutho: The reason I don't create them as dictionaries from the start is to be able to do the `sortperm` thing, which does make a difference especially when the `Fsymbols` are not cached (`A4Irrep`).

src/fusiontrees/basic_manipulations.jl
Jutho: No, I think it has always been there, probably to make it type stable indeed. Maybe aggressive constpropagation might be a better annotation?

src/tensors/treetransformers.jl
lkdvos: I think this does make sense, since if I think about `V` as being the data, you could conceive that the tensor really is `V * F`, where `F` is the vector of different fusiontrees (which effectively labels the "basis vectors" in some sense?). `transform(F) -> A * F'` is then the transformation on the fusion trees, such that `t' = V * A * F'` and we need to multiply the data on the right. 
So I would say that it is reasonable to just change the definition and immediately compute the transpose of what is being computed now?

I'm still confused about what we are really computing in our transform functions though, and which things are active and passive basis transformations etc

src/fusiontrees/basic_manipulations.jl
lkdvos: It is marked as `public` though, so I don't think there is an issue with it not being exported

src/fusiontrees/duality_manipulations.jl
lkdvos: I just tried to remove it, and I think the type assertions on the previous recursive implementation where really necessary to make this type stable. I think in the end this is quite a valid use case for a generated function and the implementation is quite elegant, so I would rather avoid trying to make a type stable recursive implementation if you are okay with this

src/fusiontrees/basic_manipulations.jl
```diff
@@ -175,7 +175,9 @@ See also [`multi_Fmove_inv`](@ref), [`multi_associator`](@ref).
 """
 function multi_Fmove(f::FusionTree{I, N}) where {I, N}
     if FusionStyle(I) isa UniqueFusion
-        coupled = N == 1 ? rightunit(f.uncoupled[1]) : (N == 2 ? f.uncoupled[2] : f.innerlines[end])
+        coupled = N == 1 ? rightunit(f.uncoupled[1]) :
+            N == 2 ? f.uncoupled[2] :
+            only(⊗(Base.tail(f.uncoupled)...))
```
Jutho: We can probably compute this more compactly using

```suggestion
        coupled = only(dual(f.uncoupled[1]) ⊗ f.coupled)
```

src/fusiontrees/braiding_manipulations.jl
```diff
@@ -0,0 +1,377 @@
...
+        c = inner_extended[i]
+        e = inner_extended[i + 1]
+        if FusionStyle(I) isa UniqueFusion
+            c′ = first(a ⊗ d)
+            coeff = oftype(
+                oneT,
+                if inv
+                    conj(Rsymbol(d, c, e) * Fsymbol(d, a, b, e, c′, c)) * Rsymbol(d, a, c′)
+                else
+                    Rsymbol(c, d, e) * conj(Fsymbol(d, a, b, e, c′, c) * Rsymbol(a, d, c′))
```
borisdevos, line 162: Just wanted to mention here that the `a` and `d` of the last R-symbols might be switched, as that is consistent with https://quantumkithub.github.io/TensorKit.jl/stable/man/sectors/#Manipulations-on-a-fusion-tree

src/tensors/treetransformers.jl
lkdvos: Turns out that the AD errors were precisely because I did this inconsistently for the `AbstractTensorMap` vs `TensorMap` codepath.
I hope this is now fixed, and consistent, but this also shows that we are missing some test coverage, very specifically for the cases of indexmanipulations of combinations of `TensorMap` and `AdjointTensorMap`s. I will just open an issue for this here, since I really dont want to include that in this PR as well, and I think we should revisit our test suite in any case to allow us to tests GPU support in a better/more modular way.

src/auxiliary/auxiliary.jl
```diff
@@ -27,6 +27,20 @@ function permutation2swaps(perm)
...
+    @inbounds for i in 1:N
+        p[mod1(i + 1, N)] == mod1(p[i] + 1, N) || return false
+    end
+    return true
+end
+# two-argument version: check whether `v1` is a cyclic permutation of `v2`
+function iscyclicpermutation(v1, v2)
+    length(v1) == length(v2) || return false
+    return iscyclicpermutation(indexin(v1, v2))
+end
```
kshyatt, line 42: Might these two be a good upstream contribution to `Permutations.jl`?

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,577 @@
+# BASIC MANIPULATIONS:
+#----------------------------------------------
+# -> rewrite generic fusion tree in basis of fusion trees in standard form
+# -> only depend on Fsymbol
+"""
```
kshyatt, line 5: An example might be nice here 🥺

src/fusiontrees/basic_manipulations.jl
```diff
@@ -0,0 +1,577 @@
...
+    -> (::FusionTree{I, N₁ + N₂ - 1})
+
+Join fusion trees `f₁` and `f₂` by connecting the coupled sector of `f₁` to the first
+uncoupled sector of `f₂`. The resulting tree has uncoupled sectors given by those of `f₁`
+followed the remaining uncoupled sectors (except for the first) of `f₂`. This
+requires that `f₁.coupled == f₂.uncoupled[1]` and `f₂.isdual[1] == false`. This
+operation is the inverse of split, in the sense that `f == join(split(f, M)...)`
+holds for all `M` between `0` and `N`, where `N` is the number of uncoupled sectors of `f`.
+
+See also [`split`](@ref) and [`insertat`](@ref).
```
kshyatt, line 65: Another good spot for an example?

src/fusiontrees/duality_manipulations.jl
```diff
@@ -0,0 +1,739 @@
...
+            break
+        elseif mod1(q2[k] + 1, N) == q1[k]
+            i = q2[k]
+            j = q1[k]
+            break
+        else
+            k += 1
+        end
+    end
+    k > N₃ && throw(ArgumentError("Not a planar trace"))
```
kshyatt: Error message here could be better, maybe at least display k and N?

src/spaces/homspace.jl
```diff
@@ -138,6 +138,24 @@ Return the fusiontrees corresponding to all valid fusion channels of a given `Ho
 """
 fusiontrees(W::HomSpace) = fusionblockstructure(W).fusiontreelist
 
+"""
+    fusionblocks(W::HomSpace)
+
+Return the fusionblocks corresponding to all valid fusion channels of a given `HomSpace`,
```
kshyatt: Maybe "Return the `FusionTreeBlock`s"?

src/tensors/indexmanipulations.jl
```diff
@@ -516,22 +516,13 @@ end
...
-                _add_abelian_kernel_nonthreaded!(
-                    tdst, tsrc, p, transformer, α, β, backend...
-                )
-            end
+            add_trivial_kernel!(tdst, tsrc, p, transformer, α, β, backend...)
         else
+            style = FusionStyle(I)
             if use_threaded_transform(tdst, transformer)
-                _add_general_kernel_threaded!(tdst, tsrc, p, transformer, α, β, backend...)
+                add_kernel_threaded!(style, tdst, tsrc, p, transformer, α, β, backend...)
```
kshyatt, line 523: Just a general comment here but we should think how this interacts with the CUDA/AMD stuff, since threads have their own streams associated with them

src/tensors/tensoroperations.jl
```diff
@@ -224,38 +224,63 @@ function trace_permute!(
...
+
+    cod = codomain(tsrc)
+    dom = domain(tsrc)
+    n = length(cod)
+    scale!(tdst, β)
+    r₁ = (p₁..., q₁...)
+    r₂ = (p₂..., q₂...)
+
+    # TODO: Is it worth to add multithreading support?
+    if FusionStyle(I) isa UniqueFusion
```
kshyatt: Stylistic thing here but this could be refactored into two new functions, dispatching on `FusionStyle(I)`, that might be easier to read?

docs/src/man/fusiontrees.md
```diff
@@ -116,13 +116,13 @@ However, the end result should be the same, provided the pentagon and hexagon eq
...
+For `FusionStyle(I) isa UniqueFusion`, both `artin_braid` and `braid` return a single `Pair{FusionTree, <:Number}`, since there is exactly one output tree with a scalar (phase) coefficient.
+For `MultipleFusion` or `GenericFusion` sectors, use the `FusionTreeBlock`-based overloads described below.
 
 With the elementary `artin_braid`, we can then compute a more general braid.
 For this, we provide an interface
 
-[`braid(f::FusionTree{I, N}, levels::NTuple{N, Int}, permutation::NTuple{N, Int})`](@ref braid(f::FusionTree{I, N}, levels::NTuple{N, Int}, p::NTuple{N, Int}) where {I <: Sector, N})
+[`braid(f::FusionTree{I, N}, p::IndexTuple{N}, levels::IndexTuple{N})`](@ref braid(::FusionTree{I, N}, ::IndexTuple{N}, ::IndexTuple{N}) where {I, N})
 
 where the braid is specified as a permutation, such that the new sector at position `i` was originally at position `permutation[i]`, and where every uncoupled sector is also assigned a level or depth.
```
Jutho:
```suggestion
where the braid is specified as a permutation, such that the new sector at position `i` was originally at position `p[i]`, and where every uncoupled sector is also assigned a level or depth.
```

src/auxiliary/auxiliary.jl
lkdvos: I don't know if they are not already in that package, but I'm a bit hesitant to take on another dependency here for this, since this really is quite a small amount of functionality, so I think vendoring it is fine.

src/tensors/indexmanipulations.jl
lkdvos: That's definitely a good point, although I think right now not a lot of this is actually being used to multithread, so it might be fine for now.
In any case, definitely something to revisit in a separate PR :)

kshyatt: Oh yeah for sure, perhaps we should leave a comment here or something?

src/spaces/homspace.jl
```diff
@@ -138,6 +138,24 @@ Return the fusiontrees corresponding to all valid fusion channels of a given `Ho
...
+"""
+    fusionblocks(W::HomSpace)
+
+Return the [`FusionTreeBlock`](@ref)s corresponding to all valid fusion channels of a given `HomSpace`,
+grouped by their uncoupled charges.
+"""
+function fusionblocks(W::HomSpace)
+    I = sectortype(W)
+    N₁, N₂ = numout(W), numin(W)
+    isdual_src = (map(isdual, codomain(W).spaces), map(isdual, domain(W).spaces))
```
Jutho: It is annoying that `map(isdual, codomain(W))` just works, but doesn't give the right type of output. Maybe we should also define

```julia
Base.map(f, P::ProductSpace) = map(f, P.spaces)
```

test/symmetries/fusiontrees.jl
```diff
@@ -1,12 +1,20 @@
...
+using MatrixAlgebraKit: isunitary
+using LinearAlgebra
+using TupleTools
 
 # TODO: remove this once type_repr works for all included types
 using TensorKitSectors
 
+_isunitary(x::Number; kwargs...) = isapprox(x * x', one(x); kwargs...)
+_isunitary(x; kwargs...) = isunitary(x; kwargs...)
+_isone(x; kwargs...) = isapprox(x, one(x); kwargs...)
```
Jutho, line 16: I've heard someone use the expression "is one'ish" today, so maybe we should call these functions that test for properties approximately like this:
`isoneish`, `isisometricish`, ...

src/fusiontrees/braiding_manipulations.jl
```diff
@@ -0,0 +1,377 @@
...
+
+function artin_braid(src::FusionTreeBlock{I, N, 0}, i; inv::Bool = false) where {I, N}
+    1 <= i < N || throw(ArgumentError("Cannot swap outputs i=$i and i+1 out of only $N outputs"))
+    isempty(src) && return src => zeros(sectorscalartype(I), 0, 0)
+
+    uncoupled = src.uncoupled[1]
+    a, b = uncoupled[i], uncoupled[i + 1]
+    uncoupled′ = TupleTools.setindex(uncoupled, b, i)
+    uncoupled′ = TupleTools.setindex(uncoupled′, a, i + 1)
+    coupled′ = rightunit(src.uncoupled[1][N])
```
Jutho, line 89: Is there a reason that this one cannot simply remain `src.uncoupled[1].coupled` ? I don't think the multifusion case is relevant here, as these cannot be braided ? Maybe @borisdevos can chime in?

src/fusiontrees/braiding_manipulations.jl
Jutho: I think the current version is correct, but the subtlety is that the `conj` in front of `Fsymbol` also extends to this last `Rsymbol` factor. I have also tripped over this several times already, so maybe this code should be clearer if we define
```julia
Rsymbol_inv(a, b, c) = conj(Rsymbol(b, a, c))
```
and then write these lines as
```julia
if inv
   Rsymbol_inv(c, d, e) * conj(Fsymbol(d, a, b, e, c′, c)) * Rsymbol(d, a, c′)
else
    Rsymbol(c, d, e) * conj(Fsymbol(d, a, b, e, c′, c)) * Rsymbol_inv(d, a, c′)
end
```

src/fusiontrees/braiding_manipulations.jl
```diff
@@ -0,0 +1,377 @@
...
+            end
+            continue
+        end
+        # case i > 1: other naming convention
+        b = uncoupled[i]
+        d = uncoupled[i + 1]
+        a = inner_extended[i - 1]
+        c = inner_extended[i]
+        e = inner_extended[i + 1]
+        if FusionStyle(I) isa UniqueFusion
```
Jutho: Do we support `UniqueFusion` in the `FusionBlock` case? If we keep this, I would suggest to just merge it with the next case using `if FusionStyle(I) isa MultiplicityFreeFusion`, similar to `i==1` case above. I can also make and push these changes.

src/fusiontrees/braiding_manipulations.jl
borisdevos: Braided multifusion categories do exist, but it'll be a while till people can calculate that data numerically, so we don't need to worry about that. 

However, you still need to allow braiding with trivial sectors, even for the multifusion case. Even in that case, you need `coupled′` to be defined correctly.

Jutho: Yes, but can you ever braid two sectors with a different `rightunit`? In terms of the idea of diagrams with colored regions, this doesn't seem to match? Anyway, the current one is definitely fine, it was more a question out of curiosity.

src/fusiontrees/braiding_manipulations.jl
lkdvos: Great idea, feel free to go ahead, otherwise I'm also happy to do it.

src/fusiontrees/braiding_manipulations.jl
borisdevos: Yeah I actually didn't realise that the current implementation just takes the coupled sector, not worrying about any `rightunit`.

But I don't think taking the `rightunit` here is the same as the coupled sector in general, no? This is the same only for `N = 2`, otherwise you keep the left and right coloring of the coupled sector even after the braid, and this coupled sector doesn't have to be diagonal.

lkdvos: Let's keep it as-is for now, since that also avoids the need for getting the first element out of the fusiontrees, which I don't like too much either.

test/symmetries/fusiontrees.jl
lkdvos: we could, but note that this here is just a wrapper around the decision we made in MatrixAlgebraKit to call this `isunitary`, while it isn't defined for `::Number`.
I'm not sure if it really helps that much to change the name in the tests?

src/fusiontrees/braiding_manipulations.jl
lkdvos: I'm not really sure if I agree this makes it any clearer, since it then depends on knowing what `Rsymbol_inv` is defined as, which would be some internal function within TensorKit? I am happy to revisit this if we ever decide to add this to TensorKitSectors, but for now I'd vote to just leave this as-is.

test/symmetries/fusiontrees.jl
Jutho: Apparently the non-seriousness factor of this comment was not clear.

lkdvos: I actually do quite like the name :)

Jutho: Also if you have to have `isunitaryish` (which I left out on purpose).

PR #263, opened 2025-08-04
`IsingBimodule` for testing multifusion category within TensorKit

borisdevos:
This PR makes use of the newly introduced `IsingBimodule` multifusion category in QuantumKitHub/TensorKitSectors.jl#14 to test the changes in Jutho/TensorKit.jl#247 within TensorKit itself. This was needed because MultiTensorKit depends on TensorKit itself.

What's new:
- `leftoneunit` and `rightoneunit` are introduced to naturally extend `oneunit`, but not for `Type{GradedSpace{...}}` because it needs to know the objects within the `GradedSpace`.
- `artin_braid` had an `isone` check replaced with a multifusion-friendly check (although this might be excessive, since as of now we don't consider braiding within multifusion categories)
- various custom constructors to deal with not being able to evaluate `one(IsingBimodule)` 
- tests for `IsingBimodule` fusion trees, spaces and tensors
- few changes to existing tests

I left a bunch of comments everywhere to remind me of certain choices I made in the implementation. It would be nice if these were addressed. I can clean most of them afterwards, but should probably keep some of them, or at least the ones in the tests to help whoever might need to maintain those in the future.

As I started on this before #247 got merged, its commit history is here as well, and I'm not sure what kind of rebase I need to do to get rid of it. I tried something and files seemed to want to vanish, so I aborted that attempt 😅 

`IsingBimodule` still needs to get renamed to that, so currently everything is `IsingBimod`. I'm assuming we're waiting on QuantumKitHub/TensorKitSectors.jl#19 to merge to release v0.2 though.


src/spaces/gradedspace.jl
```diff
@@ -139,21 +139,15 @@ Base.oneunit(S::Type{<:GradedSpace{I}}) where {I<:Sector} = S(one(I) => 1)
 Return the corresponding vector space of type `GradedSpace{I}` that represents the trivial
 one-dimensional space consisting of the left unit of the objects in  `Sector` `I`.
 """
-function leftoneunit(S::GradedSpace{I}) where {I<:Sector}
-    sector = leftone(first(sectors(S)))
-    return spacetype(S)(sector => 1)
-end
+leftoneunit(S::GradedSpace{I}) where {I<:Sector} = oneunit(typeof(S))
```
lkdvos, line 142: What was wrong with the previous implementations? This looks like it will always fail for multifusion things

src/spaces/multifusionspace.jl
lkdvos, line 1: It's not great if we have to specialize everything specifically for `Vect[IsingBimodule]`, since this would mean we have to redo all that work for any multifusion category. In that sense, we would either need a fusion type that specifies this, such that we can specialize properly, or write these methods in a generic way that works for everything

test/tensors.jl
```diff
@@ -268,28 +272,27 @@ for V in spacelist
             end
         end
         @timedtestset "Full trace: test self-consistency" begin
-            t = rand(ComplexF64, V1 ⊗ V2' ⊗ V2 ⊗ V1')
-            t2 = permute(t, ((1, 2), (4, 3)))
```
lkdvos, line 272: Is there a reason for removing this `t2` call?

test/tensors.jl
```diff
@@ -268,28 +272,27 @@ for V in spacelist
...
-                t2 = twist!(t2, 2)
+                t2 = twist!(t, 2)
             end
             ss = tr(t2)
-            @tensor s2 = t[a, b, b, a]
-            @tensor t3[a, b] := t[a, c, c, b]
-            @tensor s3 = t3[a, a]
+            @planar s2 = t[a b; a b]
+            @planar t3[a; b] := t[a c; b c]
+            @planar s3 = t3[a; a]
```
lkdvos, line 287: It's definitely good to test both `@planar` and `@tensor`, but we probably want to test both?

test/tensors.jl
```diff
@@ -329,13 +332,15 @@ for V in spacelist
             end
         end
         @timedtestset "Index flipping: test flipping inverse" begin
+            @assert BraidingStyle(I) isa SymmetricBraiding
```
lkdvos, line 335: I don't think this is required for this test?

src/spaces/gradedspace.jl
borisdevos: I wanted `left/rightoneunit` to also work on empty spaces, because `oneunit` does so. This will indeed fail for multifusion things, hence the specialisation.

src/spaces/multifusionspace.jl
borisdevos: I'm not sure what you mean by redoing all that work. `IsingBimodule` should be the only multifusion category tested within TensorKit itself. This infrastructure already exists in MultiTensorKit (albeit in a PR for now), which will support all the other multifusion categories.

test/tensors.jl
borisdevos: I removed all the permutations in tests which weren't explicitely testing for permutations, which I had to do anyway for the multifusion tests. This would allow for testing anyonic sectors if we decide to expand the space list.

test/tensors.jl
borisdevos: This was also with expanding to non-symmetric braidings in mind. I could put a `SymmetricBraiding` check here to then evaluate `@tensor` if you want?

test/tensors.jl
borisdevos: You're right, I mistook this with just `HasBraiding`
Edit: actually I just assumed these flips required a symmetric braiding, because the next flip tests need it 🙃

test/tensors.jl
lkdvos: That's fair. I think I would like to keep both though, to actually have coverage of as much as possible. Probably it is better to just separate these things out a bit further.

src/spaces/multifusionspace.jl
lkdvos: What I mean is that if all we are testing is the specializations that are written solely for IsingBimodule, which is used mainly for testing, this doesn't actually help all that much, since it would have to be both reimplemented and retested for other multifusion categories as well

src/spaces/gradedspace.jl
lkdvos: I see, but what do you think about this then:
```julia
function leftoneunit(V::GradedSpace{I}) where {I<:Sector}
    s = sectors(V)
    u = leftoneunit(isempty(s) ? I : first(s))
    return spacetype(V)(u => 1)
end
```

src/spaces/multifusionspace.jl
```diff
@@ -0,0 +1,111 @@
...
+        return filter!(c -> hasblock(dom, c), collect(blocksectors(codom)))
+    end
+end
+
+function rightoneunit(S::Vect[IsingBimodule])
+    !isempty(sectors(S)) || throw(ArgumentError("Cannot determine type of empty space"))
+    allequal(a.col for a in sectors(S)) ||
+        throw(ArgumentError("sectors of $S do not have the same rightone"))
+
+    sector = rightone(first(sectors(S)))
```
Jutho, line 54:
```suggestion
allequal(rightone, sectors(S)) || throw(throw(ArgumentError("sectors of $S do not have the same rightone"))
return spacetype(S)(rightone(first(list))=>1)
```

PR #272, opened 2025-08-21
Excise MatrixAlgebra module entirely to use MatrixAlgebraKit

kshyatt:
Moved `one!` and `safesign` out, changed a few calls to `leftorth!`, and the `tensors` and `diagonal` tests passed locally for me. Hopefully this won't break anything downstream...


src/auxiliary/linalg.jl
```diff
@@ -84,303 +55,4 @@ end
...
-#                     resize!(work, lwork)
-#                 end
-#             end
-#             A, tau
-#         end
-#     end
-# end
-
-end
+_argmax(f, domain) = argmax(f, domain)
```
lkdvos: I think we can now simply use `argmax`, since we are requiring Julia 1.10 anyways

src/tensors/factorizations/matrixalgebrakit.jl
```diff
@@ -3,7 +3,7 @@
 for f! in
     [:svd_compact!, :svd_full!, :svd_trunc!, :svd_vals!, :qr_compact!, :qr_full!, :qr_null!,
      :lq_compact!, :lq_full!, :lq_null!, :eig_full!, :eig_trunc!, :eig_vals!, :eigh_full!,
-     :eigh_trunc!, :eigh_vals!, :left_polar!, :right_polar!]
+     :eigh_trunc!, :eigh_vals!, :left_polar!, :right_polar!, :left_orth!, :right_orth!]
```
lkdvos: Is this needed? IIRC this should just be passed on to `default_algorithm` of the underlying routine? Or is this because MatrixAlgebraKit is prematurely restricting these implementations to AbstractArray?

src/auxiliary/linalg.jl
```diff
@@ -46,35 +46,6 @@ Base.adjoint(alg::Union{SVD,SDD,Polar}) = alg
...
-# only defined in >v1.7
-@static if VERSION < v"1.7-"
-    _rf_findmax((fm, im), (fx, ix)) = isless(fm, fx) ? (fx, ix) : (fm, im)
-    _argmax(f, domain) = mapfoldl(x -> (f(x), x), _rf_findmax, domain)[2]
-else
-    _argmax(f, domain) = argmax(f, domain)
-end
-
-# TODO: define for CuMatrix if we support this
 function one!(A::StridedMatrix)
```
lkdvos: There is probably also a `one!` in MatrixAlgebraKit that we may use directly?

src/tensors/factorizations/matrixalgebrakit.jl
kshyatt: I can try to remove and see what happens!

src/auxiliary/linalg.jl
kshyatt: There is

PR #283, opened 2025-09-09
Add `absorb` for putting (part of) the contents of one tensor in another

lkdvos:
This PR adds a small utility function that embeds tensors into a different tensor, possibly of differing spaces.

The idea is that for each sector of each space, the "top-left" corner of the source data is pasted into the "top-left" corner of the destination, where the sizes are chosen such that the data fits.
In other words, this either embeds a vector space into a larger one by inclusion or into a smaller one by restriction.

The tests aren't so great, but I'm not entirely sure how to cover this better...

Additionally, this fixes an (seemingly long hidden) issue where a formatter update previously exchanged a keyword argument for an iterable, therefore losing the `dual` flag in `infimum` and `supremum` for `GradedSpace`.

### DIFF
--- a/src/spaces/gradedspace.jl
+++ b/src/spaces/gradedspace.jl
@@ -168,23 +168,19 @@ function fuse(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I<:Sector}
 end
 
 function infimum(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I<:Sector}
-    if V₁.dual == V₂.dual
-        typeof(V₁)(c => min(dim(V₁, c), dim(V₂, c))
-                   for c in
-                       union(sectors(V₁), sectors(V₂)), dual in V₁.dual)
-    else
+    Visdual = isdual(V₁)
+    Visdual == isdual(V₂) ||
         throw(SpaceMismatch("Infimum of space and dual space does not exist"))
-    end
+    return typeof(V₁)((Visdual ? dual(c) : c) => min(dim(V₁, c), dim(V₂, c))
+                      for c in intersect(sectors(V₁), sectors(V₂)); dual=Visdual)
 end
 
 function supremum(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I<:Sector}
-    if V₁.dual == V₂.dual
-        typeof(V₁)(c => max(dim(V₁, c), dim(V₂, c))
-                   for c in
-                       union(sectors(V₁), sectors(V₂)), dual in V₁.dual)
-    else
+    Visdual = isdual(V₁)
+    Visdual == isdual(V₂) ||
         throw(SpaceMismatch("Supremum of space and dual space does not exist"))
-    end
+    return typeof(V₁)((Visdual ? dual(c) : c) => max(dim(V₁, c), dim(V₂, c))
+                      for c in union(sectors(V₁), sectors(V₂)); dual=Visdual)
 end
 
 function Base.show(io::IO, V::GradedSpace{I}) where {I<:Sector}
--- a/src/tensors/linalg.jl
+++ b/src/tensors/linalg.jl
@@ -513,9 +513,10 @@ function catcodomain(t1::TT, t2::TT) where {S,N₂,TT<:AbstractTensorMap{<:Any,S
 end
 
 """
-    embed!(tdst::AbstactTensorMap, tsrc::AbstractTensorMap)
+    absorb(tdst::AbstractTensorMap, tsrc::AbstractTensorMap)
+    absorb!(tdst::AbstactTensorMap, tsrc::AbstractTensorMap)
 
-Embed the contents of `tsrc` into `tdst`, which may have different sizes of data.
+Absorb the contents of `tsrc` into `tdst`, which may have different sizes of data.
 This is equivalent to the following operation on dense arrays, but also works for symmetric
 tensors. Note also that this only overwrites the regions that are shared, and will do
 nothing on the ones that are not, so it is up to the user to properly initialize the
@@ -526,7 +527,8 @@ sub_axes = map((x, y) -> 1:min(x, y), size(tdst), size(tsrc))
 tdst[sub_axes...] .= tsrc[sub_axes...]
 ```
 """
-function embed!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap)
+absorb(tdst::AbstractTensorMap, tsrc::AbstractTensorMap) = absorb!(copy(tdst), tsrc)
+function absorb!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap)
     numin(tdst) == numin(tsrc) && numout(tdst) == numout(tsrc) ||
         throw(DimensionError("Incompatible number of indices for source and destination"))
     S = spacetype(tdst)
--- a/src/tensors/linalg.jl
+++ b/src/tensors/linalg.jl
@@ -512,6 +512,36 @@ function catcodomain(t1::TT, t2::TT) where {S,N₂,TT<:AbstractTensorMap{<:Any,S
     return t
 end
 
+"""
+    embed!(tdst::AbstactTensorMap, tsrc::AbstractTensorMap)
+
+Embed the contents of `tsrc` into `tdst`, which may have different sizes of data.
+This is equivalent to the following operation on dense arrays, but also works for symmetric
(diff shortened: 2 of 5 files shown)


lkdvos:
@Yue-Zhengyuan, @pbrehmer, @leburgel this should help with the PEPSKit expansion/shrinking procedures, but I also think this can be used to replace https://github.com/QuantumKitHub/MPSKit.jl/blob/a34817dafae7c18100e46729d179833e5c4d46a2/src/algorithms/changebonds/changebonds.jl#L23 with a more efficient version that does not require permutations.

lkdvos:
I'm happy to change the name, I just don't have that many great alternatives in mind. I think usually the only direction we really care about is embedding something small in something large, since the other direction really is just an arbitrary truncation of the "keeping the first x rows and y columns", but it can definitely be useful to just "make it fit" in some contexts.

For example, whenever a twosite update appears in the context of MPS and the truncation scheme is set to some tolerance, it is in principle possible for the "expansion" to actually become both a truncation and expansion, for example when weights are shifted between the different sectors. Then, it is convenient to not have to handle this edge-case separately, since this really is achieving the desired result. (and the resulting tensor is only used as a reasonable initial guess, not necessarily as a fully rigourous state)

Yue-Zhengyuan:
For the names, the opposite to embedding seems to be called either [projection](https://en.wikipedia.org/wiki/Projection_(mathematics)) or [retraction](https://en.wikipedia.org/wiki/Retraction_(topology)) (whose opposite is called [inclusion](https://en.wikipedia.org/wiki/Inclusion_map)). Retraction/inclusion seem to be more similar to what we are doing here.

Jutho:
I agree that retraction or projection are probably the best terminology for the operation where you loose information. For the injective operation (from smaller to larger bond dimension) I think embedding is fine. The inclusion map in mathematics is really very specific to the case where you want to make explicit that elements from some subset of a set also live in the surrounding set. If for example you have a function $f$ from some domain $D$ to some codomain $C$, and you have a subset $S \subseteq D$ of the domain, then the restriction of $f$ to S can be written as a composition $f|_S = f \circ \imath$ with $\imath$ the inclusion map from $S$ to $D$.

I am not sure if one really wants to think of smaller vector spaces as being natural subsets of the larger ones, especially when you have a tensor product of several such spaces.

leburgel:
Since, as Lukas pointed out, the action is not always purely either expanding or truncating but often both at the same time, neither `embed!` nor `project!` would be really suitable.

How about `insert!`, which then implicitly entails that only the bits that fit into the target are actually inserted? If not, then maybe `substitute!`, as in substituting overlapping parts of the source into the target? I'm definitely in favor of being pragmatic here as, setting aside mathematical rigor, both `embed!` and `insert!` really do communicate the intended effect I think, more than e.g. `project!` and even `substitute!` in my opinion.

Regardless, I think it's nice to also have an out of place version right away, in case we don't want to destroy the target.

leburgel:
If we were willing to abandon all rigor, `plop!` seems to perfectly intuit what we want this to do, at least to me.

lkdvos:
Would it be okay if I keep `embed(!)`, but add in the docstring that this function does not check for strict embeddings, and instead will just make everything fit, so it is up to the user to ensure that `all(<=, space(tsrc), space(tdst))` if they wish for a strict embedding?

Jutho:
The only potential issue I have with `embed!` is that for a long time I have been wondering if we need some utility function in TensorKit that, given a number of `ElementarySpace` objects (of the same common type), say `V1` and `V2` (but it could be extended beyond 2), creates the space `V = V1 ⊕ V2` as well as `AbstractTensorMap` objects `I1` with space `V1 → V` and `I2` with space `V2 → V` such that `I1' * I1 = id(V1)`, `I2' * I2 = id(V2)`, `I1' * I2` is a zero map, and `I1 * I1' + I2 * I2' = id(V)`. I was thinking of calling such a utility function something like `embedders`, and making them of some special purpose `AbstractTensorMap{S,1,1}` subtype such that multiplying with them with be very efficient (and just amount to taking a slice of the data of the `TensorMap` it is multiplied by.

The current `embed!` functionality could potentially even be writting as a `mul!` with these objects. I was a bit afraid of potential confusion or conflation between these methods. However, it does seem that these maps are precisely called inclusion maps in the math/category theory literature, so I guess a better name based on the `inclusion` concept would be appropriate.

Jutho:
Screenshot 2025-09-11 at 00 24 12

I like "ram in" for this concept 😄

src/tensors/linalg.jl
```diff
@@ -512,6 +512,36 @@ function catcodomain(t1::TT, t2::TT) where {S,N₂,TT<:AbstractTensorMap{<:Any,S
...
+tensors. Note also that this only overwrites the regions that are shared, and will do
+nothing on the ones that are not, so it is up to the user to properly initialize the
+destination.
+
+```julia
+sub_axes = map((x, y) -> 1:min(x, y), size(tdst), size(tsrc))
+tdst[sub_axes...] .= tsrc[sub_axes...]
+```
+"""
+function embed!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap)
```
Jutho: Is zeroing out existing data in `tdst` part of the contract, or not?

lkdvos: It's not, it is actually useful to be able to seed the destination with small random noise

src/spaces/gradedspace.jl
```diff
@@ -168,23 +168,19 @@ function fuse(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I<:Sector}
...
-        typeof(V₁)(c => max(dim(V₁, c), dim(V₂, c))
-                   for c in
-                       union(sectors(V₁), sectors(V₂)), dual in V₁.dual)
-    else
+    Visdual = isdual(V₁)
+    Visdual == isdual(V₂) ||
         throw(SpaceMismatch("Supremum of space and dual space does not exist"))
-    end
+    return typeof(V₁)((Visdual ? dual(c) : c) => max(dim(V₁, c), dim(V₂, c))
+                      for c in union(sectors(V₁), sectors(V₂)); dual=Visdual)
```
Jutho, line 183: Out of curiosity, are both implementations functionally equivalent and is it just nicer to have the error checks at the top. Or is there something technically superior to the new implementation?

lkdvos: They would be, but there is a weird error that got introduced at some point which I actually traced to a formatting change where there is a very subtle difference between:
```julia
julia> function my_function(args; kwargs...)
           @info "my args are" args
           @info "my kwargs are" kwargs
       end
my_function (generic function with 1 method)

julia> my_function(x for x in 1:4, y = 3)
┌ Info: my args are
└   args = Base.Generator{Base.Iterators.ProductIterator{Tuple{UnitRange{Int64}, Int64}}, var"#20#21"}(var"#20#21"(), Base.Iterators.ProductIterator{Tuple{UnitRange{Int64}, Int64}}((1:4, 3)))
┌ Info: my kwargs are
└   kwargs = Base.Pairs{Symbol, Union{}, Tuple{}, @NamedTuple{}}()

julia> my_function(x for x in 1:4; y = 3)
┌ Info: my args are
└   args = Base.Generator{UnitRange{Int64}, typeof(identity)}(identity, 1:4)
┌ Info: my kwargs are
│   kwargs =
│    pairs(::NamedTuple) with 1 entry:
└      :y => 3

julia> my_function(x for x in 1:4, y in 3)
┌ Info: my args are
└   args = Base.Generator{Base.Iterators.ProductIterator{Tuple{UnitRange{Int64}, Int64}}, var"#22#23"}(var"#22#23"(), Base.Iterators.ProductIterator{Tuple{UnitRange{Int64}, Int64}}((1:4, 3)))
┌ Info: my kwargs are
└   kwargs = Base.Pairs{Symbol, Union{}, Tuple{}, @NamedTuple{}}()

julia> my_function(3, y=3)
┌ Info: my args are
└   args = 3
┌ Info: my kwargs are
│   kwargs =
│    pairs(::NamedTuple) with 1 entry:
└      :y => 3
```

unfortunately both iterators and keyword arguments can be specified using `in` and `=`, so because previously we weren't being explicit about the `;` to separate arguments from keyword arguments, it seems like we messed up the `dual` as keyword argument and it became a product iterator, therefore losing the dual flag altogether.
To avoid having this confusion, I felt like instead of changing the `,` for `;` it might be more obvious to just have the error upfront

PR #284, opened 2025-09-10
Make `truncspace` explanation more specific

leburgel:
I managed to confuse myself about the intended behavior of the `truncspace` truncation scheme since dimensions were coming out smaller than I expected, and the explanation in the `tsvd` docstring didn't really help. It turned out that the truncation did behave as I would expect, but then the docstring is technically incorrect and a bit confusing.

So, I thought it would be better to make this a bit more specific.

### DIFF
--- a/docs/src/man/tensors.md
+++ b/docs/src/man/tensors.md
@@ -946,8 +946,8 @@ the keyword argument `p`. The default value `notrunc()` implies no truncation, a
 *   `truncdim(χ::Integer)`: finds the optimal truncation such that the equivalent total
     dimension of the internal vector space is no larger than `χ`;
 
-*   `truncspace(W)`: truncates such that the dimension of the internal vector space is
-    smaller than or equal to that of `W` in any sector, i.e. with
+*   `truncspace(W)`: truncates such that the dimension of the internal vector space is no
+    greater than that of `W` in any sector, i.e. with
     `W₀ = min(fuse(codomain(t)), fuse(domain(t)))` this option will result in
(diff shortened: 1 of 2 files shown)


docs/src/man/tensors.md
```diff
@@ -947,7 +947,7 @@ the keyword argument `p`. The default value `notrunc()` implies no truncation, a
     dimension of the internal vector space is no larger than `χ`;
 
 *   `truncspace(W)`: truncates such that the dimension of the internal vector space is
-    smaller than that of `W` in any sector, i.e. with
+    smaller than or equal to that of `W` in any sector, i.e. with
```
lkdvos:
```suggestion
    no bigger than that of `W` in any sector, i.e. with
```
also possible

leburgel: In that case I would go for 'no larger than', but other than that it's the same to me.

lkdvos: 🤷

Jutho: Or 'not greater than', or 'not higher than'. I now also start to wonder about the formality of 'no' versus 'not. 

It seems Julia docs describes `<=` and `>=` "Less-than-or-equals" and "greater-than-or-equals". 

I also find online that 'larger' sometimes refers to the absolute value, but since dimensions are positive that is still valid. The mysteries of the English language ...

leburgel: I went with 'no greater than'. I think 'greater' is the most appropriate one in the end, but I do think 'no' fits better than 'not' in this kind of phrase.

Jutho (approved):

Jutho:
I don't think we need CI to finish for this.

PR #290, opened 2025-09-25
Changes for TensorKitSectors v0.3

borisdevos:
This PR introduces the renamings concerning units of sectors, as well as replacing `conj` of sectors with `dual`. Additionally, `Base.oneunit` now falls back to `unitspace`. I also took the liberty of defining `zerospace`, to which `Base.zero` falls back. This is a breaking change.

A follow-up PR will then actually use `UnitStyle` etc to generalise a couple of functions to deal with multifusion categories.

Something I can still do is update the documentation related to sectors, but I know @Jutho just started on a documentation overhaul, so I haven't done so yet.

This requires QuantumKitHub/TensorKitSectors.jl#25 to be released.

### DIFF
--- a/src/factorizations/truncation.jl
+++ b/src/factorizations/truncation.jl
@@ -225,19 +225,19 @@ function MAK.findtruncated_svd(values::SectorDict, strategy::TruncationSpace)
 end
 
 function MAK.findtruncated(values::SectorDict, strategy::TruncationIntersection)
-    inds = map(Base.Fix1(MAK.findtruncated, values), strategy)
+    inds = map(Base.Fix1(MAK.findtruncated, values), strategy.components)
     return SectorDict(
         c => mapreduce(
-                Base.Fix2(getindex, c), _ind_intersect, inds;
+                Base.Fix2(getindex, c), MatrixAlgebraKit._ind_intersect, inds;
                 init = trues(length(values[c]))
             ) for c in intersect(map(keys, inds)...)
     )
 end
-function MAK.findtruncated_svd(Sd::SectorDict, strategy::TruncationIntersection)
-    inds = map(Base.Fix1(MAK.findtruncated_svd, Sd), strategy)
+function MAK.findtruncated_svd(values::SectorDict, strategy::TruncationIntersection)
+    inds = map(Base.Fix1(MAK.findtruncated_svd, values), strategy.components)
     return SectorDict(
         c => mapreduce(
-                Base.Fix2(getindex, c), _ind_intersect, inds;
+                Base.Fix2(getindex, c), MatrixAlgebraKit._ind_intersect, inds;
                 init = trues(length(values[c]))
             ) for c in intersect(map(keys, inds)...)
     )
--- a/test/factorizations.jl
+++ b/test/factorizations.jl
@@ -260,6 +260,14 @@ for V in spacelist
                 @test isisometry(U4)
                 @test isisometry(Vᴴ4; side = :right)
                 @test norm(t - U4 * S4 * Vᴴ4) <= 0.5
+
+                trunc = truncrank(dim(domain(S)) ÷ 2) & trunctol(; atol = λ - 10eps(λ))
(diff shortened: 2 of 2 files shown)


src/TensorKit.jl
```diff
@@ -9,15 +9,17 @@ module TensorKit
...
-export FusionStyle, UniqueFusion, MultipleFusion, MultiplicityFreeFusion,
-    SimpleFusion, GenericFusion
+export FusionStyle, UniqueFusion, MultipleFusion, MultiplicityFreeFusion, SimpleFusion, GenericFusion
+export UnitStyle, SimpleUnit, GenericUnit
 export BraidingStyle, SymmetricBraiding, Bosonic, Fermionic, Anyonic, NoBraiding
 export Trivial, Z2Irrep, Z3Irrep, Z4Irrep, ZNIrrep, U1Irrep, SU2Irrep, CU1Irrep
 export ProductSector
 export FermionParity, FermionNumber, FermionSpin
 export FibonacciAnyon, IsingAnyon
+export unit, rightunit, leftunit, allunits, isunit
```
Jutho: These need to be moved to below line 42 `# methods for sectors and properties thereof`

src/spaces/vectorspaces.jl
```diff
@@ -120,22 +119,27 @@ Return the sum of all degeneracy dimensions of the vector space `V`.
...
 
 Return the corresponding vector space of type `S` that represents the zero-dimensional or empty space.
-This is, with a slight abuse of notation, the zero element of the direct sum of vector spaces. 
+This is, with a slight abuse of notation, the zero element of the direct sum of vector spaces.
+`Base.zero` falls back to `zerospace`.
 """
-Base.zero(V::ElementarySpace) = zero(typeof(V))
+zerospace(V::ElementarySpace) = zerospace(typeof(V))
+Base.zero(V::ElementarySpace) = zerospace(V)
+Base.zero(::Type{V}) where {V <: ElementarySpace} = zerospace(V)
```
Jutho, line 142: I am actually wondering whether this was also considered type piracy. In the same way as overloading `Base.show(::Type{MyType})` is not a good idea, this has the same issue, that it is overloading a base method for `Type{MyType}`, and I don't know if a package is considered to be the owner of `Type{MyType}` just because it owns `MyType`. Though I don't see how packages implementing new numerical types could do without something like this.

Jutho:
This looks good to me. I've slightly reorganized the `export` section of "src/TensorKit.jl". Let's hope I didn't mess it up. Then I think that can be merged.

src/spaces/vectorspaces.jl
Jutho: Also, this is a non-blocking philosophical question. No action for this PR required.

lkdvos: Just to share some of my ideas on the topic: I've also been going back and forth on this issue...
I think one way of answering this (definitely not the only way) is to think about who owns something like `Vector{MyType}` and then claim that actually `Type{MyType}` might be the same?
For practical reasons, I think it's probably reasonable to "take ownership" of `Vector{MyType}` for your own number type if you have a specialized implementation for eg. `+`, but by that logic you should also be allowed to implement `show(::Vector{MyType})`, and therefore also `show(::Type{MyType})`.

I think that really, there's two points to type piracy. On the one hand, there is the issue with composing it across packages, you don't want to have multiple people trying to overwrite the same method. On the other hand, there's the issue with composing it in functionality, where you need the behavior of the method to remain the same as what the function intended, as otherwise you might get errors, or even silently wrong results.

The first point is clearly not an issue for things like `Vector{MyType}` or `Type{MyType}`, since you own the type so you are "the first" to be able to define a method. The second point is a bit more subtle I think, since I would argue that for `+(::Vector{MyType}, ::Vector{MyType})`, it is clear what the method should and shouldn't do, while `show` is just one of those functions where the API is really badly specified, which makes it a lot more dangerous to touch. I think that if the observable behavior of `show` would be clearly defined, we *should* be allowed to overload that.

PR #291, opened 2025-09-30
[WIP] Generalising functions to support `GenericUnit`

borisdevos:
This PR redoes parts of #263, but now with `UnitStyle` to more cleanly deal with multifusion categories.
Work in progress, I will not touch this for a month.


src/spaces/gradedspace.jl
```diff
@@ -92,10 +92,9 @@ Base.hash(V::GradedSpace, h::UInt) = hash(V.dual, hash(V.dims, h))
 field(::Type{<:GradedSpace}) = ℂ
 InnerProductStyle(::Type{<:GradedSpace}) = EuclideanInnerProduct()
 function dim(V::GradedSpace)
-    return reduce(
-        +, dim(V, c) * dim(c) for c in sectors(V);
-        init = zero(dim(unit(sectortype(V))))
-    )
+    T = Base.promote_op(*, Int, real(sectorscalartype(sectortype(V))))
+    return reduce(+, dim(V, c) * dim(c) for c in sectors(V);
+                  init=zero(T))
```
lkdvos: I don't know if we ended up trying this, but I am aware of `sum` having some special handling of `init` values.
I think we could maybe avoid the `promote_op` machinery by using something that is a little more friendly on inference: (double check type stability though!)

```suggestion
    blockdim(c) = dim(V, c) * dim(c)
    return sum(blockdim, sectors(V))
```

test/symmetries/spaces.jl
```diff
@@ -220,13 +220,21 @@ end
...
-    # space with a single sector
-    W = @constinferred GradedSpace(unit(I) => 1)
-    @test W == GradedSpace(unit(I) => 1, randsector(I) => 0)
+    # space with unit(s)
+    if isa(UnitStyle(I), GenericUnit)
+        W = @constinferred GradedSpace(unit => 1 for unit in allunits(I))
+        dict = Dict((unit => 1 for unit in allunits(I))..., randsector(I) => 0)
+        @test W == GradedSpace(dict)
+        @test @constinferred(zerospace(V)) == GradedSpace(unit => 0 for unit in allunits(I))
+        # randsector never returns trivial sector, so this cannot error
```
lkdvos: This is probably not a great internal detail to depend on. For example, in TensorKitSectors I had to change this to ensure the tests also run for `Trivial`, which would then silently break these tests

test/symmetries/spaces.jl
```diff
@@ -220,13 +220,21 @@ end
     @test eval_show(typeof(V)) == typeof(V)
     # space with no sectors
     @test dim(@constinferred(zerospace(V))) == 0
-    # space with a single sector
-    W = @constinferred GradedSpace(unit(I) => 1)
-    @test W == GradedSpace(unit(I) => 1, randsector(I) => 0)
+    # space with unit(s)
+    if isa(UnitStyle(I), GenericUnit)
```
lkdvos: Shouldn't these tests also work for the other cases? Isn't any fusion category trivially a multifusion one, even if `length(allunits) == 1` ?

test/tensors/tensors.jl
```diff
@@ -269,28 +271,27 @@ for V in spacelist
...
-                t2 = twist!(t2, 2)
+                t2 = twist!(t, 2)
             end
             ss = tr(t2)
-            @tensor s2 = t[a, b, b, a]
-            @tensor t3[a, b] := t[a, c, c, b]
-            @tensor s3 = t3[a, a]
+            @plansor s2 = t[a b; a b]
+            @plansor t3[a; b] := t[a c; b c]
+            @plansor s3 = t3[a; a]
```
lkdvos: I don't think you can simply change this to `@plansor` without affecting the results, so I think it might be better to just skip these tests if the braidingstyle doesn't allow it?

test/symmetries/spaces.jl
borisdevos: Yeah, I was planning on bringing this up later, but there are several cases where I don't really need to distinguish between the two and can immediately put the multifusion version. Up till now I've separated them (both in source and tests), though, purely for readability and to avoid having to understand multifusion stuff if it's not required. If there's no strong opinion about this, I'd like to keep it that way.

test/tensors/tensors.jl
borisdevos: I was able to put `@planar` here in #263 and that worked, so I thought `@plansor` would deal with the braiding style then. Is this incorrect?

lkdvos: I'm actually really confused `@tensor` and `@planar` give the same result to begin with, and looking at it a bit more in detail I think `tr` usually should correspond to `@planar`, so I don't know why `@tensor` was working before...

src/spaces/gradedspace.jl
```diff
@@ -134,8 +133,58 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
+
+Return the corresponding vector space of type `GradedSpace{I}` that represents the trivial
+one-dimensional space consisting of the left unit of the objects in  `Sector` `I`.
+"""
+function leftunitspace(S::GradedSpace{I}) where {I<:Sector}
+    if UnitStyle(I) isa SimpleUnit
+        return unitspace(typeof(S))
+    else
+        !isempty(sectors(S)) || throw(ArgumentError("Cannot determine type of empty space"))
+        allequal(a.row for a in sectors(S)) ||
```
Jutho: We do not want to depend on specific fields, but rather use everything through the generic interface, which is fully formulated in terms of methods:

```suggestion
        allequal(leftunit, sectors(S)) ||
```

src/spaces/gradedspace.jl
```diff
@@ -134,8 +133,58 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
+
+Return the corresponding vector space of type `GradedSpace{I}` that represents the trivial
+one-dimensional space consisting of the right unit of the objects in `Sector` `I`.
+"""
+function rightunitspace(S::GradedSpace{I}) where {I<:Sector}
+    if UnitStyle(I) isa SimpleUnit
+        return unitspace(typeof(S))
+    else
+        !isempty(sectors(S)) || throw(ArgumentError("Cannot determine type of empty space"))
+        allequal(a.row for a in sectors(S)) ||
```
```suggestion
        allequal(rightunit, sectors(S)) ||
```

src/spaces/gradedspace.jl
```diff
@@ -134,8 +133,58 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
+        return S(unit => 1 for unit in allunits(I))
+    end
+end
+function zerospace(S::Type{<:GradedSpace{I}}) where {I<:Sector}
+    if UnitStyle(I) isa SimpleUnit
+        return S(unit(I) => 0)
+    else
+        return S(unit => 0 for unit in allunits(I))
+    end
+end
```
Jutho: Not sure why the previous implementation was like that, but this should simply be
```suggestion
zerospace(S::Type{<:GradedSpace}) = S()
```

src/spaces/productspace.jl
```diff
@@ -271,19 +278,25 @@ function insertleftunit(
...
+    I = sectortype(P)
+    if UnitStyle(I) isa SimpleUnit
+        u = unitspace(spacetype(P))
+    else
+        N > 0 || throw(ArgumentError("cannot insert a sensible unit space in the empty product space"))
+        i == 0 && throw(DomainError((P, i), "cannot insert a sensible right unit space"))
+        u = rightunitspace(P[i])
+    end
     if dual
         u = TensorKit.dual(u)
```
Jutho, line 306: Not sure why the `TensorKit.` was here.
```suggestion
        u = dual(u)
```

Jutho: Same for `conj` below, but I cannot add this as a suggestion, as it is too far away from code that is changed in this PR.

src/spaces/productspace.jl
```diff
@@ -248,19 +249,25 @@ fuse(P::ProductSpace{S, 0}) where {S <: ElementarySpace} = unitspace(S)
...
+    I = sectortype(P)
+    if UnitStyle(I) isa SimpleUnit
+        u = unitspace(spacetype(P))
+    else
+        N > 0 || throw(ArgumentError("cannot insert a sensible unit space in the empty product space"))
+        i > N && throw(DomainError((P, i), "cannot insert a sensible left unit space"))
+        u = leftunitspace(P[i])
+    end
     if dual
         u = TensorKit.dual(u)
```
Jutho, line 276: Not sure why `TensorKit.` is here. Same change applies to `conj` below, but I cannot add this as a suggestion.
```suggestion
        u = dual(u)
```

src/spaces/gradedspace.jl
```diff
@@ -134,8 +133,58 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
+        return spacetype(S)(sector => 1)
+    end
+end
+
+function unitspace(S::Type{<:GradedSpace{I}}) where {I<:Sector}
+    if UnitStyle(I) isa SimpleUnit
+        return S(unit(I) => 1)
+    else
+        return S(unit => 1 for unit in allunits(I))
+    end
```
Jutho: This should probably just work generally:
```suggestion
    return S(unit => 1 for unit in allunits(I))
```

Jutho:
I am a bit torn on renaming `insertleftunit` (and `...right...`) and `removeunit` by adding `space` at the end. I agree it is consistent, but maybe also a bit excessive? Then again, it is perhaps more clear what the goal of these functions is then.

src/spaces/productspace.jl
borisdevos: Since `dual` is a variable within the function, you need to specify `TensorKit.` to call the function instead of the bool.

src/spaces/gradedspace.jl
borisdevos: I'm struggling a bit with this, because I actually need the `init` value for certain cases, e.g. calling `zerospace`. However, by adding the `init` value I get `Float64`-valued dimensions of e.g. `SU2Irrep` graded spaces, leading to `reshape`s being messy in different parts of the code. I can get around this by converting these floats back to integers, because they need to be those anyways,  but it feels somewhat messy.

test/symmetries/spaces.jl
borisdevos: I'm not sure what you mean by this. Could you clarify a little?

src/TensorKit.jl
```diff
@@ -15,10 +15,12 @@ export BraidingStyle, SymmetricBraiding, Bosonic, Fermionic, Anyonic, NoBraiding
 export Trivial, Z2Irrep, Z3Irrep, Z4Irrep, ZNIrrep, U1Irrep, SU2Irrep, CU1Irrep
 export ProductSector
 export FermionParity, FermionNumber, FermionSpin
-export FibonacciAnyon, IsingAnyon
+export FibonacciAnyon, IsingAnyon, IsingBimodule
+export unit, rightunit, leftunit, allunits, isunit
 
 # Export common vector space, fusion tree and tensor types
 export VectorSpace, Field, ElementarySpace # abstract vector spaces
+export unitspace, zerospace, leftunitspace, rightunitspace
```
lkdvos: These are already in the list a bit further down

src/spaces/gradedspace.jl
```diff
@@ -92,10 +92,8 @@ Base.hash(V::GradedSpace, h::UInt) = hash(V.dual, hash(V.dims, h))
 field(::Type{<:GradedSpace}) = ℂ
 InnerProductStyle(::Type{<:GradedSpace}) = EuclideanInnerProduct()
 function dim(V::GradedSpace)
-    return reduce(
-        +, dim(V, c) * dim(c) for c in sectors(V);
-        init = zero(dim(unit(sectortype(V))))
-    )
+    T = promote_type(Int, real(sectorscalartype(sectortype(V))))
+     return sum(dim(V, c) * dim(c) for c in sectors(V); init = zero(T))
```
lkdvos: This now no longer produces integers for SU2 if I'm not mistaken?

borisdevos: Yeah you're right, I clarify this in an above comment

src/spaces/productspace.jl
Jutho: Oh right; I overlooked that.

src/spaces/gradedspace.jl
Jutho: `sum` also supports the `init` kwarg. But then indeed you should recompute it. Though I am not sure how `real(sectorscalartype(sectortype(V)))` tells you whether the quantum dimensions are real or integer?

borisdevos: It doesn't. I was thinking at the level of the F-symbols, but for certain sectors (in particular `SU2Irrep`) the `dim(::Sector)` is specialised to return integers, while the F-symbols are floats.

src/spaces/gradedspace.jl
```diff
@@ -92,10 +92,8 @@ Base.hash(V::GradedSpace, h::UInt) = hash(V.dual, hash(V.dims, h))
 field(::Type{<:GradedSpace}) = ℂ
 InnerProductStyle(::Type{<:GradedSpace}) = EuclideanInnerProduct()
 function dim(V::GradedSpace)
-    return reduce(
-        +, dim(V, c) * dim(c) for c in sectors(V);
-        init = zero(dim(unit(sectortype(V))))
-    )
+    s = sectors(V)
+    return isempty(s) ? dim(first(allunits(sectortype(V)))) * 0 : sum(c -> dim(c) * dim(V, c), s)
```
```suggestion
    init = 0 * dim(first(allunits(sectortype(V))))
    return sum(c -> dim(c) * dim(V, c), sectors(V); init = init)
```

src/tensors/diagonal.jl
```diff
@@ -306,7 +306,7 @@ function LinearAlgebra.pinv(d::DiagonalTensorMap; kwargs...)
     if iszero(atol)
         rtol = get(kwargs, :rtol, zero(real(T)))
     else
-        rtol = sqrt(eps(real(float(oneunit(T))))) * length(d.data)
+        rtol = sqrt(eps(real(float(unitspace(T))))) * length(d.data)
```
Jutho: This seems like a replace all error. `T` is a `<:Number` type. But it is strange that `oneunit(T)` was used here instead of simply `one(T)`: @lkdvos ?

I don't think we support "dimensions" (a.k.a. units, such as SI units) in TensorKit.jl, so I think `one(T)` and `oneunit(T)` are the same in any conceivable case.

src/spaces/gradedspace.jl
lkdvos: Does this not work and produce stable results?
```julia
if isempty(sectors(V))
    return dim(first(allunits(sectortype(V)))) * 0
else
    blockdim(c) = dim(V, c) * dim(c)
    return sum(blockdim, sectors(V))
end
```

I think the trade-off is whether we want to simply always compute this `init`, and if not whether we think `Base.promote_op` is acceptable

src/tensors/diagonal.jl
lkdvos: yeah, probably messed that up in some of the merges I did when I ported Boris' changes to the testsuite

src/spaces/gradedspace.jl
lkdvos: Sorry, I missed this, I'm struggling to figure out how to navigate github's "improvements", where we now apparently need to choose between seeing only some of the comments, or some other, but not all at the same time, and definitely not in any deterministic order

test/symmetries/spaces.jl
lkdvos: I think the fact that `randsector` neve returns a trivial sector is not something we can rely on, since I don't think this is something that is guaranteed for that function.
For example, if you want to run the tests for `PlanarTrivial, or `Trivial`, we now have to choose between breaking that function or breaking the assumption you made here.

borisdevos: We can simply remove the `randsector` sector in this test then?

src/spaces/gradedspace.jl
```diff
@@ -134,8 +132,61 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
 end
 
-unitspace(S::Type{<:GradedSpace{I}}) where {I <: Sector} = S(unit(I) => 1)
-zerospace(S::Type{<:GradedSpace{I}}) where {I <: Sector} = S(unit(I) => 0)
+@static if VERSION < v"1.11"
+    _allequal(f, xs) = allequal(Base.Generator(f, xs))
+else
+    _allequal(f, xs) = allequal(f, xs)
+end
+
```
lkdvos: Would you mind moving this to the `auxiliary.jl` file, and adding a `# TODO: remove once support for v1.10 is dropped comment?

src/spaces/gradedspace.jl
```diff
@@ -134,8 +132,61 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
+"""
+    leftunitspace(S::GradedSpace{I}) where {I<:Sector} -> GradedSpace{I}
+
+Return the corresponding vector space of type `GradedSpace{I}` that represents the trivial
+one-dimensional space consisting of the left unit of the objects in  `Sector` `I`.
+"""
```
lkdvos: This function, along with the docstring should probably work for generic `ElementarySpace` instances, not just for `GradedSpace`s. I don't think you even need to change the actual implementation for this to still work.

I would therefore:
- change the signature to be `ElementarySpace`
- move the definition to `vectorspaces.jl`

src/spaces/gradedspace.jl
```diff
@@ -134,8 +132,61 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
-unitspace(S::Type{<:GradedSpace{I}}) where {I <: Sector} = S(unit(I) => 1)
-zerospace(S::Type{<:GradedSpace{I}}) where {I <: Sector} = S(unit(I) => 0)
+@static if VERSION < v"1.11"
+    _allequal(f, xs) = allequal(Base.Generator(f, xs))
+else
+    _allequal(f, xs) = allequal(f, xs)
+end
+
+"""
+    leftunitspace(S::GradedSpace{I}) where {I<:Sector} -> GradedSpace{I}
```
lkdvos:
```suggestion
    leftunitspace(S::GradedSpace{I}) where {I <: Sector} -> GradedSpace{I}
```

I know we aren't very consistent with this yet, but I like to keep the docstrings also compatible with the runic style, even though the formatter doesn't automatically handle that.

src/spaces/gradedspace.jl
```diff
@@ -134,8 +132,61 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
+        _allequal(leftunit, sectors(S)) ||
+            throw(ArgumentError("sectors of $S do not have the same left unit"))
+
+        sector = leftunit(first(sectors(S)))
+        return spacetype(S)(sector => 1)
+    end
+end
+
+"""
+    rightunitspace(S::GradedSpace{I}) where {I<:Sector} -> GradedSpace{I}
```
lkdvos: same comment here

src/spaces/gradedspace.jl
```diff
@@ -134,8 +132,61 @@ function Base.axes(V::GradedSpace{I}, c::I) where {I <: Sector}
...
+    end
+end
+
+"""
+    unitspace(S::GradedSpace{I}) where {I<:Sector} -> GradedSpace{I}
+
+Return the corresponding vector space of type `GradedSpace{I}` that represents the
+ space consisting of the unit(s) of the objects in `Sector` `I`. For `I` with simple unit,
+this is a one-dimensional space.
+"""
```
lkdvos: Do we need an additional docstring here? If you really want to include something about the unitstyle, I'd add that to the corresponding docstring for `ElementarySpace` (in the `vectorspaces.jl` file)

src/spaces/homspace.jl
```diff
@@ -96,11 +96,16 @@ function blocksectors(W::HomSpace)
     N₁ = length(codom)
     N₂ = length(dom)
     I = sectortype(W)
-    # TODO: is sort! still necessary now that blocksectors of ProductSpace is sorted?
-    if N₂ <= N₁
-        return sort!(filter!(c -> hasblock(codom, c), blocksectors(dom)))
+    if N₁ == 0 && N₂ == 0
+        return allunits(I)
+    elseif N₁ == 0
+        return filter!(isone, collect(blocksectors(dom))) # module space cannot end in empty space
```
lkdvos:
```suggestion
        return filter!(isunit, collect(blocksectors(dom))) # module space cannot end in empty space
```

src/spaces/homspace.jl
```diff
@@ -96,11 +96,16 @@ function blocksectors(W::HomSpace)
...
+    elseif N₂ == 0
+        return filter!(isone, collect(blocksectors(codom)))
```
lkdvos:
```suggestion
        return filter!(isunit, collect(blocksectors(codom)))
```

src/spaces/homspace.jl
```diff
@@ -96,11 +96,16 @@ function blocksectors(W::HomSpace)
     N₁ = length(codom)
     N₂ = length(dom)
     I = sectortype(W)
-    # TODO: is sort! still necessary now that blocksectors of ProductSpace is sorted?
```
lkdvos, line 96: Did we verify that `sort` is no longer required here?

src/spaces/homspace.jl
```diff
@@ -329,7 +336,8 @@ end
     end
 
     fusiontreeindices = sizehint!(
-        FusionTreeDict{Tuple{F₁, F₂}, Int}(), length(fusiontreelist)
+        FusionTreeDict{Tuple{F₁, F₂}, Int}(),
+        length(fusiontreelist)
```
lkdvos: I'm not sure I think this is that much easier to read here

src/spaces/productspace.jl
```diff
@@ -151,7 +151,7 @@ function blocksectors(P::ProductSpace{S, N}) where {S, N}
     end
     bs = Vector{I}()
     if N == 0
-        push!(bs, unit(I))
+        return allunits(I)
```
lkdvos: I have two remarks/questions for this:
- It is slightly odd to have a different return type for the `N == 0` case, which could be anything, instead of the sorted vector that is otherwise returned. In principle this should be fine, and the generic code paths should handle everything just fine, but it might be nicer to just be consistent and do something like the following and ensure everything is sorted
```julia
for u in allunits(I)
    push!(bs, u)
end
```
- If not, then this statement should probably be moved above the allocation of the `bs` vector

src/spaces/productspace.jl
```diff
@@ -260,7 +260,15 @@ function insertleftunit(
...
         conj::Bool = false, dual::Bool = false
     ) where {i}
-    u = unitspace(spacetype(P))
+    N = length(P)
+    I = sectortype(P)
+    if UnitStyle(I) isa SimpleUnit
+        u = unitspace(spacetype(P))
+    else
+        N > 0 || throw(ArgumentError("cannot insert a sensible unit space in the empty product space"))
+        i > N && throw(DomainError((P, i), "cannot insert a sensible left unit space"))
```
lkdvos: `DomainError` or `BoundsError`?

src/spaces/productspace.jl
```diff
@@ -283,7 +291,15 @@ function insertrightunit(
...
         conj::Bool = false, dual::Bool = false
     ) where {i}
-    u = unitspace(spacetype(P))
+    N = length(P)
+    I = sectortype(P)
+    if UnitStyle(I) isa SimpleUnit
+        u = unitspace(spacetype(P))
+    else
+        N > 0 || throw(ArgumentError("cannot insert a sensible unit space in the empty product space"))
+        i == 0 && throw(DomainError((P, i), "cannot insert a sensible right unit space"))
```
lkdvos: `BoundsError`?

src/spaces/productspace.jl
```diff
@@ -305,7 +321,15 @@ and [`insertrightunit`](@ref insertrightunit(::ProductSpace, ::Val{i}) where {i}
...
-    isisomorphic(P[i], unitspace(P[i])) || _nontrivialspaceerror(P, i)
+    I = sectortype(P)
+    if isa(UnitStyle(I), SimpleUnit)
+        isisomorphic(P[i], unitspace(P[i])) || _nontrivialspaceerror(P, i)
+    else
+        isisomorphic(P[i], leftunitspace(P[i])) ||
+            isisomorphic(P[i], rightunitspace(P[i])) ||
+            isisomorphic(P[i], unitspace(P[i])) ||
+            _nontrivialspaceerror(P, i)
+    end
```
lkdvos: This might be slightly cleaner to factor it out into a function (maybe `isunitspace`?). I know we can actually use this in some places in MPSKit as well.

src/tensors/linalg.jl
```diff
@@ -296,7 +296,7 @@ function LinearAlgebra.rank(
         t::AbstractTensorMap;
         atol::Real = 0, rtol::Real = atol > 0 ? 0 : _default_rtol(t)
     )
-    r = dim(one(sectortype(t))) * 0
+    r = zero(real(sectorscalartype(sectortype(t))))
```
lkdvos: This is probably the same story as with the `dim(::GradedSpace)` initialization, I don't think this should return a `Float64` for `SU2Irrep` (it might also be good to add tests for this)

src/TensorKit.jl
```diff
@@ -15,10 +15,12 @@ export BraidingStyle, SymmetricBraiding, Bosonic, Fermionic, Anyonic, NoBraiding
 export Trivial, Z2Irrep, Z3Irrep, Z4Irrep, ZNIrrep, U1Irrep, SU2Irrep, CU1Irrep
 export ProductSector
 export FermionParity, FermionNumber, FermionSpin
-export FibonacciAnyon, IsingAnyon
+export FibonacciAnyon, IsingAnyon, IsingBimodule
+export unit, rightunit, leftunit, allunits, isunit
 
 # Export common vector space, fusion tree and tensor types
 export VectorSpace, Field, ElementarySpace # abstract vector spaces
+export leftunitspace, rightunitspace
```
lkdvos: Can you move this to the vector space methods section instead of the types :)

src/TensorKit.jl
```diff
@@ -15,10 +15,12 @@ export BraidingStyle, SymmetricBraiding, Bosonic, Fermionic, Anyonic, NoBraiding
 export Trivial, Z2Irrep, Z3Irrep, Z4Irrep, ZNIrrep, U1Irrep, SU2Irrep, CU1Irrep
 export ProductSector
 export FermionParity, FermionNumber, FermionSpin
-export FibonacciAnyon, IsingAnyon
+export FibonacciAnyon, IsingAnyon, IsingBimodule
+export unit, rightunit, leftunit, allunits, isunit
```
lkdvos: Can you move this to the `# reexport methods for sectors and properties thereof` section?

test/symmetries/spaces.jl
lkdvos: While I'm definitely fine with this for the source, for the tests I think it's actually important to check that the fusion category still works as a special case of a multifusion one, since this checks that our code is actually generic. So, since fusion $$\subseteq$$ multifusion, it's fine to run some tests only for fusion but it isn't fine to run some tests only for multifusion 😉

test/symmetries/spaces.jl
lkdvos: works for me

lkdvos: (I didn't realize this was already like this previously by the way, it's just that my eye fell on it now)

test/tensors/factorizations.jl
```diff
@@ -9,22 +9,22 @@ spacelist = try
...
     end
 catch
-    (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂)
+    (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
 end
 
 eltypes = (Float32, ComplexF64)
 
-for V in spacelist
+for V in (VIB_M, VIB_diag)
```
lkdvos:
```suggestion
for V in spacelist
```

test/tensors/factorizations.jl
```diff
@@ -37,7 +37,7 @@ for V in spacelist
         @testset "QR decomposition" begin
             for T in eltypes,
                     t in (
-                        rand(T, W, W), rand(T, W, W)', rand(T, W, V1), rand(T, V1, W)',
+                        rand(T, W, W), rand(T, W, W)', rand(T, W, V4), rand(T, V4, W)',
```
lkdvos, line 42: Is this necessary? If so, why?

lkdvos: Same question for everything below

test/tensors/factorizations.jl
```diff
@@ -236,7 +236,8 @@ for V in spacelist
                 @test isisometry(U)
                 @test isisometry(Vᴴ; side = :right)
 
-                trunc = truncrank(dim(domain(S)) ÷ 2)
+                #FIXME: dimension of S is a float, might be a real issue if it's a decimal
+                trunc = truncrank(Int(dim(domain(S)) ÷ 2))
```
lkdvos: This can for now be changed back, same comment for the `Int` changes below.

test/tensors/tensors.jl
```diff
@@ -45,18 +46,20 @@ for V in spacelist
...
+                    @test b1 == b2
+                    @test eltype(bs) === Pair{typeof(c), typeof(b1)}
+                    @test typeof(b1) === TensorKit.blocktype(t)
+                    @test typeof(c) === sectortype(t)
+                end
             end
         end
         @timedtestset "Tensor Dict conversion" begin
-            W = V1 ⊗ V2 ⊗ V3 ← V4 ⊗ V5
+            W = V1 ⊗ V2 ← V3 ⊗ V4 ⊗ V5
```
lkdvos, line 62: Why this change?

test/tensors/tensors.jl
```diff
@@ -45,18 +46,20 @@ for V in spacelist
...
                 bs = @constinferred blocks(t)
-                (c, b1), state = @constinferred Nothing iterate(bs)
-                @test c == first(blocksectors(W))
-                next = @constinferred Nothing iterate(bs, state)
-                b2 = @constinferred block(t, first(blocksectors(t)))
-                @test b1 == b2
-                @test eltype(bs) === Pair{typeof(c), typeof(b1)}
-                @test typeof(b1) === TensorKit.blocktype(t)
-                @test typeof(c) === sectortype(t)
+                if !isempty(blocksectors(t)) # multifusion space ending on module gives empty data
```
lkdvos, line 49: We should probably craft the multifusion space examples such that this doesn't happen, to make sure these tests run?

test/tensors/tensors.jl
```diff
@@ -96,7 +99,7 @@ for V in spacelist
             end
         end
         @timedtestset "Basic linear algebra" begin
-            W = V1 ⊗ V2 ⊗ V3 ← V4 ⊗ V5
+            W = V1 ⊗ V2 ← V3 ⊗ V4 ⊗ V5
```
lkdvos, line 102: Why this change?

test/tensors/tensors.jl
```diff
@@ -134,41 +137,60 @@ for V in spacelist
                 @test dot(t2, t) ≈ conj(dot(t2', t'))
                 @test dot(t2, t) ≈ dot(t', t2')
 
-                i1 = @constinferred(isomorphism(T, V1 ⊗ V2, V2 ⊗ V1))
-                i2 = @constinferred(isomorphism(Vector{T}, V2 ⊗ V1, V1 ⊗ V2))
-                @test i1 * i2 == @constinferred(id(T, V1 ⊗ V2))
-                @test i2 * i1 == @constinferred(id(Vector{T}, V2 ⊗ V1))
+                if isa(UnitStyle(I), SimpleUnit) || !isempty(blocksectors(V2 ⊗ V1))
```
lkdvos: Can we make this work for the `isempty(blocksectors(...))` case?

test/tensors/tensors.jl
```diff
@@ -134,41 +137,60 @@ for V in spacelist
...
+                    @test i1 * i2 == @constinferred(id(T, V1 ⊗ V2))
+                    @test i2 * i1 == @constinferred(id(Vector{T}, V2 ⊗ V1))
+                end
 
-                w = @constinferred(isometry(T, V1 ⊗ (oneunit(V1) ⊕ oneunit(V1)), V1))
-                @test dim(w) == 2 * dim(V1 ← V1)
-                @test w' * w == id(Vector{T}, V1)
-                @test w * w' == (w * w')^2
+                if isa(UnitStyle(I), SimpleUnit)
+                    # FIXME?: unitspace returns all units, leads to invalid fusion channels
```
lkdvos: Can we fix this with `rightunitspace`?

test/tensors/tensors.jl
```diff
@@ -134,41 +137,60 @@ for V in spacelist
...
+                end
             end
         end
         @timedtestset "Trivial space insertion and removal" begin
             W = V1 ⊗ V2 ⊗ V3 ← V4 ⊗ V5
             for T in (Float32, ComplexF64)
                 t = @constinferred rand(T, W)
-                t2 = @constinferred insertleftunit(t)
-                @test t2 == @constinferred insertrightunit(t)
+                if isa(UnitStyle(I), SimpleUnit)
```
lkdvos: These tests are almost equal, can we merge them? Why do the multifusion ones need explicit indices?

test/tensors/tensors.jl
```diff
@@ -268,29 +292,30 @@ for V in spacelist
...
+                if !isdual(V1)
+                    t2 = twist!(t, 1)
+                end
+                if isdual(V2)
+                    t2 = twist!(t, 2)
+                end
+                ss = tr(t2)
+                @planar s2 = t[a b; a b]
+                @planar t3[a; b] := t[a c; b c]
+                @planar s3 = t3[a; a]
```
lkdvos: If this is only run for symmetric braiding, probably best to keep this to `@tensor`.

test/tensors/tensors.jl
```diff
@@ -268,29 +292,30 @@ for V in spacelist
...
         end
         @timedtestset "Partial trace: test self-consistency" begin
-            t = rand(ComplexF64, V1 ⊗ V2' ⊗ V3 ⊗ V2 ⊗ V1' ⊗ V3')
-            @tensor t2[a, b] := t[c, d, b, d, c, a]
-            @tensor t4[a, b, c, d] := t[d, e, b, e, c, a]
-            @tensor t5[a, b] := t4[a, b, c, c]
+            t = rand(ComplexF64, V1 ⊗ V2 ⊗ V3 ← V1 ⊗ V2 ⊗ V3)
+            @planar t2[a; b] := t[c d b; c d a]
+            @planar t4[a b; c d] := t[e d c; e b a]
+            @planar t5[a; b] := t4[a c; b c]
```
lkdvos: I still don't really like removing testcases like this, if you want to add more planar tests you definitely can, but not by replacing `@tensor` ones I think.

test/tensors/tensors.jl
```diff
@@ -505,8 +536,13 @@ for V in spacelist
         end
         @timedtestset "Tensor product: test via norm preservation" begin
             for T in (Float32, ComplexF64)
-                t1 = rand(T, V2 ⊗ V3 ⊗ V1, V1 ⊗ V2)
-                t2 = rand(T, V2 ⊗ V1 ⊗ V3, V1 ⊗ V1)
+                if isa(UnitStyle(I), SimpleUnit) || !isempty(blocksectors(V2 ⊗ V1))
```
lkdvos: Is it not possible to change the `V1, ...` so this does works for both?

test/setup.jl
```diff
@@ -74,13 +76,42 @@ function force_planar(tsrc::TensorMap{<:Any, <:GradedSpace})
     return tdst
 end
 
+function random_fusiontree(I::Type{<:Sector}, N::Int) # for fusion tree tests
```
lkdvos: This doesn't actually return a fusion tree, so the naming is a bit unintuitive.

test/setup.jl
```diff
@@ -74,13 +76,42 @@ function force_planar(tsrc::TensorMap{<:Any, <:GradedSpace})
...
+        try
+            in = rand(collect(⊗(out...)))
+        catch e
+            if isa(e, ArgumentError)
+                in = nothing
+            else
+                rethrow(e)
+            end
+        end
+    end
```
lkdvos: It might be slightly cleaner to add random sectors one by one, making sure the last two can properly fuse. This looks like something with a vanishingly small chance of succeeding as `N` gets larger, although we might never really care either since we might not get to that point?

test/setup.jl
```diff
@@ -74,13 +76,42 @@ function force_planar(tsrc::TensorMap{<:Any, <:GradedSpace})
...
+end
+
+# for fusion tree merge test
+function safe_tensor_product(x::I, y::I) where {I <: Sector}
+    tp = x ⊗ y
+    if isempty(tp)
+        return nothing
+    else
+        return tp
+    end
```
lkdvos: What is the difference between checking for nothing instead of checking for `isempty`?

test/symmetries/fusiontrees.jl
```diff
@@ -223,102 +241,113 @@ using .TestSetup
...
-        for in in ⊗(out...)
-            for i in 1:(N - 1)
-                for f in fusiontrees(out, in, isdual)
-                    d1 = @constinferred TK.artin_braid(f, i)
-                    @test norm(values(d1)) ≈ 1
-                    d2 = empty(d1)
-                    for (f1, coeff1) in d1
-                        for (f2, coeff2) in TK.artin_braid(f1, i; inv = true)
-                            d2[f2] = get(d2, f2, zero(coeff1)) + coeff2 * coeff1
+    if isa(UnitStyle(I), SimpleUnit)
```
lkdvos: Would checking `BraidingStyle(I)` for has braiding also work?

test/symmetries/fusiontrees.jl
```diff
@@ -332,136 +361,200 @@ using .TestSetup
...
+                        Af2 = convert(Array, f2)
+                        sz1 = size(Af1)
+                        sz2 = size(Af2)
+                        d1 = prod(sz1[1:(end - 1)])
+                        d2 = prod(sz2[1:(end - 1)])
+                        dc = sz1[end]
+                        A = reshape(
+                            reshape(Af1, (d1, dc)) * reshape(Af2, (d2, dc))',
+                            (sz1[1:(end - 1)]..., sz2[1:(end - 1)]...)
+                        )
```
lkdvos:
```suggestion
                        A = convert(Array, (f1, f2))
```

test/symmetries/fusiontrees.jl
```diff
@@ -332,136 +361,200 @@ using .TestSetup
...
+                            sz1′ = size(Af1′)
+                            sz2′ = size(Af2′)
+                            d1′ = prod(sz1′[1:(end - 1)])
+                            d2′ = prod(sz2′[1:(end - 1)])
+                            dc′ = sz1′[end]
+                            A2 += coeff * reshape(
+                                reshape(Af1′, (d1′, dc′)) *
+                                    reshape(Af2′, (d2′, dc′))',
+                                (sz1′[1:(end - 1)]..., sz2′[1:(end - 1)]...)
+                            )
```
lkdvos:
```suggestion
                            A2 .+= coeff .* convert(Array, (f1′, f2′))
```

src/spaces/productspace.jl
borisdevos: Not sure, you're not really out of bounds to put something there, but just not the left unit space.

src/spaces/productspace.jl
borisdevos: Same here, but now putting a right unit space at the far left

test/tensors/factorizations.jl
borisdevos: I designed the spaces for `IsingBimodule` such that the blocksectors of `V1 ⊗ V2` would be equal to those of `V3 ⊗ V4 ⊗ V5` and in particular of `V4`.

test/tensors/tensors.jl
borisdevos: Same answer as in the factorizations tests: the blocksectors are non-empty this way

test/tensors/tensors.jl
borisdevos: Same here

test/tensors/tensors.jl
borisdevos: Yes, but not with the spaces I made for `IsingBimodule`. The only cases where `V1 ⊗ V2` and its reverse can work for multifusion is that they either both only contain objects from one of the fusion subcategories, or one of them contains objects from one of the off-diagonal (module) categories and the other contains its dual objects. The latter won't work for these specific tests because the resulting color of `V1 ⊗ V2` is different to `V2 ⊗ V1`.

test/tensors/tensors.jl
borisdevos: The default to e.g. `insertleftunit` is to put a left unit space at the end (far right) of the domain, while for multifusion this is "ill-defined" as you should base the coloring of the inserted left unit space on the left unit of the space to its right, and not to the right unit to its left. Same for `insertrightunit` but reversed.

test/tensors/tensors.jl
borisdevos: I gave an answer somewhere else that also holds here about `V1 ⊗ V2` vs `V2 ⊗ V1`.

src/spaces/productspace.jl
borisdevos: I wondered about this as well, and indeed I've never encountered an issue with this returning a tuple in some occasions. I'll just change it to return a vector every time.

test/setup.jl
borisdevos: Testing this for large N, it gets pretty nasty around N ~ 20 already with 4GB of allocations... maybe the "add random sectors which definitely fuse" approach is more appropriate.

test/setup.jl
borisdevos: Not sure what you mean. A "forbidden" tensor product will not return type nothing, but an iterator which is empty

test/symmetries/fusiontrees.jl
borisdevos: Yes! I put it like this because I'm pretty sure a part of these tests should still work for IsingBimodule, and I think I made some in the old PR, but they were hardcoded to check braidings with unit objects.

test/tensors/tensors.jl
borisdevos: The first of the 2 multifusion spaces in `spacelist` does work here because it's essentially fusion. I can think of others involving module categories which ultimately fuse to diagonal elements, but this will probably fail in other tests below.

src/spaces/homspace.jl
borisdevos: I didn't, but I'd assume it isn't. Shall I remove the `sort!` here?

test/tensors/tensors.jl
lkdvos: That makes sense, but I'm wondering if this is an implementation issue rather than an interface issue.
The main reason I added `left-` and `right-` is mostly to allow for selecting where to put the index, and not necessarily about whether this has to be a `leftunitspace` or a `rightunitspace`.
The default value is to put the utility space at the end, where `leftunitspace` ensures it is in the domain, while `rightunitspace` might have it in the `codomain` for an `N, 0` tensor.
Unless I'm mistaken, it should be possible to have the same behavior for the multifusion case, (unless it's a `0, 0`-Tensor, in which case I don't think we can?), even though the implementation might require handling some edge cases separately.

For example, `insertleftunit(V1 * V2 <- V3, 4)` is equivalent to `insertrightunit(V1 * V2 <- V3, 3)`, and `insertleftunit(V1 * V2 <- one(), 3)` could still be defined to return the same result as `repartition(insertrightunit(V1 * V2 <- one(), 2; dual=true), 2, 1)`.
(I think?)

test/tensors/factorizations.jl
lkdvos: My main worry is that this is fundamentally changing a lot of the other tests as well, as these spaces might not necessarily be designed for this other order. It could be fine, to do this, but probably we should then more carefully outline what kind of assumptions we are making on our input spaces. For example, we could preface the entire tests with `@assert !isempty(blocksectors(...))` to indicate this?

src/spaces/homspace.jl
lkdvos: I checked, and if the `blocksectors` always returns the vector (which is sorted at the end of that function), this should be fine!

test/setup.jl
lkdvos: What is the purpose of exchanging an empty iterator with `nothing` is basically the question

test/symmetries/fusiontrees.jl
lkdvos: Ah fair point, I forgot that we can still braid units

test/setup.jl
borisdevos: I didn't think of that for some reason, will clean up

test/tensors/tensors.jl
borisdevos: I got these functions working for more edge cases, but empty spaces can't be dealt with like you suggested in general. If V1 * V2 only contains diagonal elements, then you can bring the inserted unit space to the other side, but if it's off-diagonal then the coloring isn't preserved by bringing the unit space to the other side. So for now I still make `insertleft/rightunit` error when dealing with `one(...)`.

borisdevos:
I managed to get most tests working for `GenericUnit`, so the git diff isn't that bad anymore. The only exception is the tensors tests, where I occasionally cannot test even `GenericUnit` for certain spaces while keeping everything generic 🙃 . In particular, I often check for non-empty block sectors because not every product of spaces is allowed, but a tensor map can still be constructed with these spaces and simply be empty. Then, a lot of the tests become trivial.

src/TensorKit.jl
```diff
@@ -31,10 +31,14 @@ export DiagonalTensorMap, BraidingTensor
 export TruncationScheme
 export SpaceMismatch, SectorMismatch, IndexError # error types
 
+# general vector space methods
```
lkdvos:
```suggestion
```
duplicate with line below?

src/spaces/productspace.jl
```diff
@@ -151,7 +151,10 @@ function blocksectors(P::ProductSpace{S, N}) where {S, N}
     end
     bs = Vector{I}()
     if N == 0
-        push!(bs, unit(I))
+        for u in allunits(I)
+            push!(bs, u)
+        end
+        return bs
```
lkdvos:
```suggestion
        return collect(allunits(I))
```

src/spaces/productspace.jl
```diff
@@ -196,7 +199,7 @@ hasblock(P::ProductSpace, c::Sector) = !isempty(fusiontrees(P, c))
     blockdim(P::ProductSpace, c::Sector)
 
 Return the total dimension of a coupled sector `c` in the product space, by summing over
-all `dim(P, s)` for all tuples of sectors `s::NTuple{N, <:Sector}` that can fuse to  `c`,
+all `dim(P, s)` for all tuples of sectors `s::NTuple{N, <: Sector}` that can fuse to  `c`,
```
lkdvos:
```suggestion
all `dim(P, s)` for all tuples of sectors `s::NTuple{N, Sector}` that can fuse to `c`,
```

src/spaces/vectorspaces.jl
```diff
@@ -141,6 +143,77 @@ zerospace(V::ElementarySpace) = zerospace(typeof(V))
 Base.zero(V::ElementarySpace) = zerospace(V)
 Base.zero(::Type{V}) where {V <: ElementarySpace} = zerospace(V)
 
+"""
+    leftunitspace(V::S) where {S <: ElementarySpace} -> S
+
+Return the corresponding vector space of type `ElementarySpace` that represents the trivial
```
lkdvos:
```suggestion
Return the corresponding vector space of type `S` that represents the trivial
```

src/spaces/vectorspaces.jl
```diff
@@ -141,6 +143,77 @@ zerospace(V::ElementarySpace) = zerospace(typeof(V))
...
+one-dimensional space, i.e. the space that is isomorphic to the corresponding field. For vector spaces 
+of type `GradedSpace{I}`, this corresponds to the unique left unit of the objects in `Sector` `I` present
```
lkdvos: This sentence looks slightly odd, might be a replace issue?

test/tensors/factorizations.jl
```diff
@@ -236,12 +238,15 @@ for V in spacelist
                 @test isisometry(U)
                 @test isisometry(Vᴴ; side = :right)
 
-                trunc = truncrank(dim(domain(S)) ÷ 2)
+                # dimension of S is a float for IsingBimodule
+                nvals = dim(domain(S)) ÷ 2
+                eltype(nvals) <: AbstractFloat && (nvals = Int(nvals))
```
lkdvos:
```suggestion
                nvals = round(Int, dim(domain(S)) / 2)
```

test/tensors/factorizations.jl
```diff
@@ -298,9 +303,11 @@ for V in spacelist
                 @test @constinferred isposdef(vdv)
                 t isa DiagonalTensorMap || @test !isposdef(t) # unlikely for non-hermitian map
 
-                d, v = @constinferred eig_trunc(t; trunc = truncrank(dim(domain(t)) ÷ 2))
+                nvals = dim(domain(t)) ÷ 2
+                eltype(nvals) <: AbstractFloat && (nvals = Int(nvals))
```
lkdvos:
```suggestion
                nvals = round(Int, dim(domain(t)) / 2)
```

test/tensors/tensors.jl
```diff
@@ -269,28 +278,45 @@ for V in spacelist
...
+                @tensor t3[a, b] := t[a, c, c, b]
+                @tensor s3 = t3[a, a]
+            else
+                t = rand(ComplexF64, V1 ⊗ V2 ← V1 ⊗ V2) # avoid permutes
+                ss = @constinferred tr(t)
+                @test conj(ss) ≈ tr(t')
+                @planar s2 = t[a b; a b]
+                @planar t3[a; b] := t[a c; b c]
+                @planar s3 = t3[a; a]
             end
```
lkdvos:
```suggestion
            end
            t = rand(ComplexF64, V1 ⊗ V2 ← V1 ⊗ V2) # avoid permutes
            ss = @constinferred tr(t)
            @test conj(ss) ≈ tr(t')
            @planar s2 = t[a b; a b]
            @planar t3[a; b] := t[a c; b c]
            @planar s3 = t3[a; a]
```
I think the planar tests can always run, and the other ones only conditionally?

test/tensors/tensors.jl
```diff
@@ -269,28 +278,45 @@ for V in spacelist
...
-            t = rand(ComplexF64, V1 ⊗ V2' ⊗ V3 ⊗ V2 ⊗ V1' ⊗ V3')
-            @tensor t2[a, b] := t[c, d, b, d, c, a]
-            @tensor t4[a, b, c, d] := t[d, e, b, e, c, a]
-            @tensor t5[a, b] := t4[a, b, c, c]
+            if symmetricbraiding
+                t = rand(ComplexF64, V1 ⊗ V2 ⊗ V3 ← V1 ⊗ V2 ⊗ V3)
+                @tensor t2[a; b] := t[c d b; c d a]
+                @tensor t4[a b; c d] := t[e d c; e b a]
+                @tensor t5[a; b] := t4[a c; b c]
+            else
```
lkdvos: Same comment here

src/spaces/homspace.jl
```diff
@@ -96,11 +96,16 @@ function blocksectors(W::HomSpace)
     N₁ = length(codom)
     N₂ = length(dom)
     I = sectortype(W)
-    # TODO: is sort! still necessary now that blocksectors of ProductSpace is sorted?
-    if N₂ <= N₁
-        return sort!(filter!(c -> hasblock(codom, c), blocksectors(dom)))
+    if N₁ == 0 && N₂ == 0
```
lkdvos:
```suggestion
    if N₁ == N₂ == 0
```
just cosmetic :)

src/spaces/vectorspaces.jl
borisdevos: It made sense to me the first time I typed it, but looking at it now it is confusing. I typed something new, but feel free to put whatever makes sense to you if it remains unclear.

test/tensors/factorizations.jl
```diff
@@ -329,36 +334,38 @@ for V in spacelist
...
+                        rand(T, W, V4), rand(T, V4, W),
+                        rand(T, W, V4)', rand(T, V4, W)',
                         DiagonalTensorMap(rand(T, reduceddim(V1)), V1),
                     )
 
                 d1, d2 = dim(codomain(t)), dim(domain(t))
-                @test rank(t) == min(d1, d2)
+                r = rank(t)
+                @test r == min(d1, d2)
+                @test eltype(r) == eltype(d1)
```
lkdvos:
```suggestion
                @test typeof(r) == typeof(d1)
```
This should boil down to the same thing, but `eltype` confused me for a second since this are just scalars and that seems to imply they are matrices

src/spaces/productspace.jl
```diff
@@ -151,7 +151,7 @@ function blocksectors(P::ProductSpace{S, N}) where {S, N}
     end
     bs = Vector{I}()
     if N == 0
-        push!(bs, unit(I))
+        return collect(allunits(I))
```
Jutho: This way, the `bs` has not been made in vain, and it sill goes through `sort!` at the end.
```suggestion
        append!(bs, allunits(I))
```

src/spaces/vectorspaces.jl
```diff
@@ -154,6 +156,77 @@ zerospace(V::ElementarySpace) = zerospace(typeof(V))
...
+one-dimensional space, i.e. the space that is isomorphic to the corresponding field. For vector spaces 
+of type `GradedSpace{I}`, this one-dimensional space contains the unique left unit of the objects in `Sector` `I` present
+in the vector space.
+"""
+function leftunitspace(V::ElementarySpace)
+    I = sectortype(V)
+    if UnitStyle(I) isa SimpleUnit
+        return unitspace(typeof(V))
+    else
+        !isempty(sectors(V)) || throw(ArgumentError("Cannot determine type of empty space"))
```
Jutho:
```suggestion
        !isempty(sectors(V)) || throw(ArgumentError("Cannot determine the left unit of an empty space"))
```

src/spaces/vectorspaces.jl
```diff
@@ -154,6 +156,77 @@ zerospace(V::ElementarySpace) = zerospace(typeof(V))
...
+one-dimensional space, i.e. the space that is isomorphic to the corresponding field. For vector spaces 
+of type `GradedSpace{I}`, this corresponds to the right unit of the objects in `Sector` `I` present
+in the vector space.
+"""
+function rightunitspace(V::ElementarySpace)
+    I = sectortype(V)
+    if UnitStyle(I) isa SimpleUnit
+        return unitspace(typeof(V))
+    else
+        !isempty(sectors(V)) || throw(ArgumentError("Cannot determine type of empty space"))
```
Jutho:
```suggestion
        !isempty(sectors(V)) || throw(ArgumentError("Cannot determine the right unit of an empty space"))
```

src/spaces/vectorspaces.jl
```diff
@@ -154,6 +156,77 @@ zerospace(V::ElementarySpace) = zerospace(typeof(V))
...
+trivial one-dimensional space. For vector spaces of type `GradedSpace{I}` where `Sector` `I` has a
+semi-simple unit structure, this returns `true` if `V` is isomorphic to either the left, right or
+semi-simple unit space.
+"""
+function isunitspace(V::ElementarySpace)
+    I = sectortype(V)
+    return if isa(UnitStyle(I), SimpleUnit)
+        isisomorphic(V, unitspace(V))
+    else
+        try
```
Jutho: Is the `try ... catch` block to deal with the case of an empty space `V`? Can we maybe check this beforehand (and return false), and then omit the try - catch block?

src/spaces/vectorspaces.jl
```diff
@@ -130,10 +130,12 @@ Always returns `false` for spaces where `V == conj(V)`, i.e. vector spaces over
 """ isconj(::ElementarySpace)
 
 """
-    unitspace(V::S) where {S<:ElementarySpace} -> S
+    unitspace(V::S) where {S <: ElementarySpace} -> S
 
 Return the corresponding vector space of type `S` that represents the trivial
 one-dimensional space, i.e. the space that is isomorphic to the corresponding field.
+For vector spaces of type `GradedSpace{I}` where `Sector` `I` has a semi-simple unit 
+structure, this returns a multi-dimensional space corresponding to all unit sectors in `I`.
```
Jutho:
```suggestion
For vector spaces where `I = sectortype(S)` has a semi-simple unit structure
(`UnitStyle(I) == GenericUnit()`), this returns a multi-dimensional space corresponding to all unit sectors: `dim(unitspace(V), s) == 1` for all `s in allunits(I)`. 
```

test/tensors/tensors.jl
Jutho: Could this also have been resolved by defining the spaces in the multifusion space lists, i.e. essentially amounting to
```julia
V1, V2, V3 = dual(V3), V1, V2
```
?

test/tensors/tensors.jl
Jutho: Can we simply use `symmetricbraiding` as the condition here?

src/spaces/vectorspaces.jl
borisdevos: Not only empty spaces, but also cases where there isn't a unique left or right unit. Then it throws an `ArgumentError` right after the `_allequal` check in `left/rightunitspace`.

Jutho: and how about the following check for the multistyle case:
```julia
all(isunit, sectors(V)) || return false
for s in sectors(V)
    dim(V, s) == 1 || return false
end
return true
```

This would be slightly more general in the case when there are more than 2 unit objects, since it would admit any space that has anywhere from one of the unit objects to all of the unit objects appearing exactly once. I think that is a valid generalization, if we accept spaces with a single unit object, or with all unit objects, then why not those with some of the unit objects?

Jutho: (obviously, I am also trying to eliminate the `try ... catch`, which I would prefer to avoid in these small routines acting on sectors and spaces)

borisdevos: I took inspiration from your suggestion, but had to check empty spaces explicitly. `all` returns true for empty collections, and the for-loop wouldn't check these either.

test/setup.jl
```diff
@@ -74,13 +76,30 @@ function force_planar(tsrc::TensorMap{<:Any, <:GradedSpace})
...
+    while len < N
+        r = randsector(I)
+        f = ⊗(t[len], r)
+        if !isempty(f)
+            len += 1
+            t[len] = r
+        end
+    end
+    return tuple(t...)
+end
```
Jutho: Type stability is probably not too important here, but it is also not difficult to realize this

```suggestion
function random_fusion(I::Type{<:Sector}, ::Val{N}) where {N} # for fusion tree tests
    N == 1 && return (randsector(I),)
    tail = random_fusion(I, Val(N-1))
    s = randsector(I)
    while isempty(⊗(s, first(tail)))
        s = randsector(I)
    end
    return (s, tail...)
end
```

but then you want to replace the `random_fusion(I, N)` calls with `random_fusion(I, Val(N))`.

test/symmetries/fusiontrees.jl
```diff
@@ -33,54 +33,63 @@ using .TestSetup
...
+            @test FusionStyle(f) == FusionStyle(I)
+            @test BraidingStyle(f) == BraidingStyle(I)
+
+            if FusionStyle(I) isa UniqueFusion
+                @constinferred FusionTree((), u, ())
+                @constinferred FusionTree((u,), u, (false,))
+                @constinferred FusionTree((u, u), u, (false, false))
+                @constinferred FusionTree((u, u, u), u)
+                if isa(UnitStyle(I), SimpleUnit)
+                    @constinferred FusionTree((u, u, u, u))
```
Jutho, line 65: Maybe we can test the error in the `GenericUnit` case, to get some better code coverage.
```suggestion
                    @constinferred FusionTree((u, u, u, u))
                else
                    @test_throws ArgumentError FusionTree((u, u, u, u))
```

test/symmetries/fusiontrees.jl
```diff
@@ -33,54 +33,63 @@ using .TestSetup
...
+                @test_throws MethodError FusionTree((u, u), u, (false, false, false))
+            else
+                errstr = "fusion tree requires inner lines if `FusionStyle(I) <: MultipleFusion`"
+                @test_throws errstr FusionTree((), u, ())
+                @test_throws errstr FusionTree((u,), u, (false,))
+                @test_throws errstr FusionTree((u, u), u, (false, false))
+                @test_throws errstr FusionTree((u, u, u), u)
+                if isa(UnitStyle(I), SimpleUnit)
+                    @test_throws errstr FusionTree((u, u, u, u))
+                end
```
Jutho: Can you actually replace the `error("...")` on line 95 of src/fusiontrees/fusiontrees.jl with a `throw(ArgumentError("..."))` and replace these tests with `@test_throws ArgumentError FusionTree(...)`. I think that is nicer than testing the specific error message, since we might want to tune these error messages a bit better, and so these tests are very brittle. 

Maybe then the last statement `@test_throws ArgumentError FusionTree((u, u, u, u))` is true for both `SimpleUnit` (because of the `MultipleFusion` and the lack of inner lines) and `GenericUnit` (because the lack of a single unit for the coupled sector).

test/tensors/factorizations.jl
Jutho: Same suggestion as below, could this have been solved by simply redesigning the `IsingBimodule` sectors, namely amounting to
```julia
V1new, V2new, V3new = dual(V3current), V1current, V2curren
```

test/setup.jl
borisdevos: I see, I thought I made it type-stable but I guess you have to put `N` as a type parameter so the compiler knows the length of the tuple at compile time?

test/tensors/tensors.jl
borisdevos: In general, no. For the fully diagonal case, this of course works, but in the 2nd set of spaces `V2` is colored as (1,2), while `V1` and `V3` are colored as (1,1), so you can only fuse the latter to the left of `V2`.

test/tensors/tensors.jl
borisdevos: The diagonal multifusion case would then be untested. We could accept that, since for a generic multifusion category the tested isomorphisms here don't make sense. My thought process was to test things that could work, even if it was a special case.

test/setup.jl
```diff
@@ -74,13 +76,24 @@ function force_planar(tsrc::TensorMap{<:Any, <:GradedSpace})
     return tdst
 end
 
+function random_fusion(I::Type{<:Sector}, ::Val{N}) where {N} # for fusion tree tests
+    N == 1 && return (randsector(I),)
+    tail = random_fusion(I, Val(N - 1))
+    s = randsector(I)
+    while isempty(⊗(s, first(tail)))
+        s = randsector(I)
+    end
```
Jutho: I am a bit worried about this turning into an infinite loop if `randsector(I)`, which is essentially `rand(smallset(I))`, does not have access to any valid `s` for which `!isempty(⊗(s, first(tail)))`. But I guess we can fix that problem if it ever poses itself.

test/symmetries/spaces.jl
```diff
@@ -220,13 +224,22 @@ end
     @test eval_show(typeof(V)) == typeof(V)
     # space with no sectors
     @test dim(@constinferred(zerospace(V))) == 0
-    # space with a single sector
-    W = @constinferred GradedSpace(unit(I) => 1)
-    @test W == GradedSpace(unit(I) => 1, randsector(I) => 0)
+    # space with unit(s), always test as if multifusion
+    W = @constinferred GradedSpace(unit => 1 for unit in allunits(I))
+    dict = Dict((unit => 1 for unit in allunits(I))...)
+    @test W == GradedSpace(dict)
```
Jutho: Expanding the generator in the `Dict` constructor is not necessary I think.
Maybe we can also reinstate something like the removed test `@test W == GradedSpace(unit(I) => 1, randsector(I) => 0)`:

```suggestion
    dict = Dict(unit => 1 for unit in allunits(I))
    @test W == GradedSpace(dict)
    @test W == GradedSpace(push!(dict, randsector(I) => 0))
```

borisdevos: It isn't, that was a remnant of adding `randsector(I)` to the dictionary :)

test/setup.jl
lkdvos: We could consider having a finite cutoff value with an explicit error here, which would give a more informative error message

borisdevos: I'm not sure if that problem can ever pose itself, unless we have some weird object in a category which can only fuse with the appropriate unit. So, since `randsector` doesn't select units, I get out of the while-loop manually by selecting a fusion which always works (the left unit of the tail).

test/tensors/tensors.jl
```diff
@@ -11,22 +11,23 @@ spacelist = try
...
-            (Vtr, Vℤ₂, Vfℤ₂, VU₁, VCU₁, VSU₂, VfSU₂, VSU₂U₁) #, VSU₃)
+            (Vtr, Vℤ₂, Vfℤ₂, VU₁, VCU₁, VSU₂, VfSU₂, VSU₂U₁, VIB_diag, VIB_M) #, VSU₃)
         end
     else
-        (Vtr, Vℤ₂, Vfℤ₂, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VSU₂U₁) #, VSU₃)
+        (Vtr, Vℤ₂, Vfℤ₂, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VSU₂U₁, VIB_diag, VIB_M) #, VSU₃)
     end
 catch
-    (Vtr, Vℤ₂, Vfℤ₂, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VSU₂U₁) #, VSU₃)
+    (Vtr, Vℤ₂, Vfℤ₂, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VSU₂U₁, VIB_diag, VIB_M) #, VSU₃)
```
Jutho, line 24: One comment here would be to spread out the tests a bit on the different platforms to lower the pressure on CI. So maybe `VIB_diag` on windows, `VIB_M` on apple and both on linux (the `else` case).

test/symmetries/fusiontrees.jl
```diff
@@ -33,54 +33,66 @@ using .TestSetup
...
         f2 = rand(collect(fusiontrees(out2, in2, isdual2)))
         for i in 1:N
-            out1 = ntuple(n -> randsector(I), N)
-            out1 = Base.setindex(out1, in2, i)
+            out1 = random_fusion(I, Val(N)) # guaranteed good fusion
+            out1 = Base.setindex(out1, in2, i) # can lead to poor fusion
+            while isempty(⊗(out1...))
+                out1 = random_fusion(I, Val(N))
+                out1 = Base.setindex(out1, in2, i)
+            end
```
Jutho, line 95: This is another `while` loop that could potentially become long-lived (maybe not, I really don't have any intuition about multi-fusion categories). I am definitely fine leaving it as is for the time being, as clearly the tests work.

But maybe it could be helpful to add a small `# TODO` comment here as a reminder to think about whether there is a better strategy in the future. Maybe we can implement some helpful tools to generate random fusion trees with given constraints, like a fixed sector at position i, or a fixed coupled sector, or ...

test/symmetries/fusiontrees.jl
```diff
@@ -223,13 +237,13 @@ using .TestSetup
             end
         end
     end
-    @testset "Fusion tree $Istr: elementary artin braid" begin
+    (UnitStyle(I) == SimpleUnit) && @testset "Fusion tree $Istr: elementary artin braid" begin
```
Jutho: This should really be considering the braiding style; it would also fail for something like `Vec[G]`. And since we now have `Vec[Z{N}]` in TensorKitSectors.jl, we should probably start testing things with this sector as well. Looking at the whole fusiontrees.jl test file, it is clear this file needs some massive reorganization that is too big to be fixed in this PR. 


```suggestion
    (BraidingStyle(I) isa HasBraiding) && @testset "Fusion tree $Istr: elementary artin braid" begin
```

test/symmetries/fusiontrees.jl
```diff
@@ -279,7 +293,7 @@ using .TestSetup
             end
         end
     end
-    @testset "Fusion tree $Istr: braiding and permuting" begin
+    (UnitStyle(I) == SimpleUnit) && @testset "Fusion tree $Istr: braiding and permuting" begin
```
Jutho: Same here
```suggestion
    (BraidingStyle(I) isa HasBraiding) && @testset "Fusion tree $Istr: braiding and permuting" begin
```

test/symmetries/fusiontrees.jl
```diff
@@ -332,50 +355,52 @@ using .TestSetup
...
-            for μ in 1:Nsymbol(in1, in2, c)
-                trees1 = TK.merge(f1, f2, c, μ)
-
-                # test merge and braid interplay
-                trees2 = Dict{keytype(trees1), complex(valtype(trees1))}()
-                trees3 = Dict{keytype(trees1), complex(valtype(trees1))}()
-                for ν in 1:Nsymbol(in2, in1, c)
-                    for (t, coeff) in TK.merge(f2, f1, c, ν)
-                        trees2[t] = get(trees2, t, zero(valtype(trees2))) + coeff * R[μ, ν]
+        if !isa(BraidingStyle(I), NoBraiding)
```
Jutho:
```suggestion
        if BraidingStyle(I) isa HasBraiding
```

test/symmetries/fusiontrees.jl
```diff
@@ -386,15 +411,30 @@ using .TestSetup
     else
         N = 4
     end
-    out = ntuple(n -> randsector(I), N)
-    numtrees = count(n -> true, fusiontrees((out..., map(dual, out)...)))
-    while !(0 < numtrees < 100)
-        out = ntuple(n -> randsector(I), N)
+    if isa(UnitStyle(I), SimpleUnit)
```
Jutho: I kinda like that you can `isa` as an infix operator
```suggestion
    if UnitStyle(I) isa SimpleUnit
```

test/tensors/factorizations.jl
Jutho: I am still a bit worried about the implications of these type of changes on the other sectors, where the spaces were also tuned a bit in order for the tests to be not too memory hungry.

However, if CI succeeds, I guess we are fine. I will quickly check any anomalies in runtime compared to previous CI runs, or maybe do this locally with this change.

test/tensors/factorizations.jl
```diff
@@ -9,17 +9,17 @@ spacelist = try
...
-            (Vtr, VU₁, VCU₁, VSU₂, VfSU₂)
+            (Vtr, VU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
         end
     else
-        (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂)
+        (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
     end
 catch
-    (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂)
+    (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
```
Jutho, line 22: Same comment here that I placed below:
It would be good to spread out the tests a bit on the different platforms to lower the pressure on CI. So maybe `VIB_diag` on windows, `VIB_M` on apple and both on linux (the `else` case) or if not on CI (the other `else` case).

test/tensors/tensors.jl
```diff
@@ -301,7 +328,8 @@ for V in spacelist
                 @test t3 ≈ convert(Array, t2)
             end
         end
-        @timedtestset "Trace and contraction" begin
+        #TODO: find version that works for all multifusion cases
+        (UnitStyle(I) == SimpleUnit()) && @timedtestset "Trace and contraction" begin
```
Jutho: Since this uses `@tensor`, I would suggest that this uses:
```suggestion
        symmetricbraiding && @timedtestset "Trace and contraction" begin
```

But once again, clearly a major reorganization of the tests is due.

test/tensors/tensors.jl
```diff
@@ -326,14 +354,14 @@ for V in spacelist
                 @test HrA12array ≈ convert(Array, HrA12)
             end
         end
-        @timedtestset "Index flipping: test flipping inverse" begin
+        (BraidingStyle(I) != NoBraiding()) && @timedtestset "Index flipping: test flipping inverse" begin
```
Jutho:
```suggestion
        (BraidingStyle(I) isa HasBraiding) && @timedtestset "Index flipping: test flipping inverse" begin
```

Jutho: `!= NoBraiding` is too much of a double negative for me 😄

test/symmetries/fusiontrees.jl
borisdevos: Since `random_fusion` now manually cuts off after 20 loops, the tests shouldn't get stuck here too long at any point, especially with the relatively small `N`s we consider in the tests. But a TODO is added for potential future improvements :)

test/symmetries/fusiontrees.jl
borisdevos: I planned on increasing code coverage of the TensorKit tests after QuantumKitHub/TensorKitSectors.jl#43 passes. I can then also look into adding other untested sectors (like the anyons in tensors tests).

test/symmetries/fusiontrees.jl
Jutho: I don't see how the cutoff in `random_fusion` would prevent this while loop from running an infinite number of times? But that's not important right now; it is good to have the TODO present.

borisdevos: No you're right, I confused myself.

Jutho (approved):

PR #298, opened 2025-10-08
Refactor test suite

lkdvos:
This PR separates out some of the parts of the test suite in order to increase parallelism on the github runners, and get slightly better feedback on where things went wrong if they did.

The main idea is to organize the tests into groups, as dictated by a folder structure, and then the QuantumKitHubActions workflow takes care of the rest. Additionally, you can control them from the local command line through:

```julia
import Pkg
Pkg.test("TensorKit"; test_args=["--groups", "other", "symmetries"]
```
which would run only the `other/` and `symmetries/` tests.

I've made sure that the individual files are all self-contained as well, so a workflow like this also works:

```julia
using TestEnv; TestEnv.activate()
include("test/symmetries/fusiontrees.jl")
```

Obviously the split I chose is arbitrary, and I'm happy to rethink all of the names.

---

This also tackles some issues we had with the CompatCheck workflow
This also fixes a breaking change we accidentally slipped in with the TensorKitSectors v0.3.1 changes.


test/autodiff/ad.jl
```diff
@@ -593,7 +597,7 @@ for V in spacelist
                     # TODO: I'm not sure how to properly test with spaces that might change
                     # with the finite-difference methods, as then the jacobian is ill-defined.
 
-                    trunc = truncrank(round(Int, min(dim(domain(t)), dim(codomain(t))) ÷ 2))
+                    trunc = truncrank(max(2, round(Int, min(dim(domain(t)), dim(codomain(t))) * (3 / 4))))
```
Jutho, line 600: This is interesting: given that our `dim` can return floating point numbers for the anyonic case, do we also need to allow floating point inputs for `truncrank`?

test/setup.jl
```diff
@@ -0,0 +1,168 @@
...
+using Test
+using TestExtras
+using Random
+using TensorKit
+using Combinatorics
+using TensorKit: ProductSector, fusiontensor, pentagon_equation, hexagon_equation
+using TensorKit: ℙ, PlanarTrivial
+using TensorOperations
+using Base.Iterators: take, product
+using LinearAlgebra: LinearAlgebra
```
Jutho: I think that within `module TestSetup`, the `using` list can be significantly simplified (I haven't tested this, just looked at what is being used):

```suggestion
using Random
using TensorKit
using TensorKit: PlanarTrivial, ProductSector, fusiontensor
using Base.Iterators: take, product
```

test/symmetries/spaces.jl
```diff
@@ -0,0 +1,446 @@
...
+using TensorKit
+using TensorKit: hassector, type_repr, HomSpace
+
+# TODO: remove this once type_repr works for all included types
+using TensorKitSectors
+
+@isdefined(TestSetup) || include("../setup.jl")
+using .TestSetup
+
+parse_show(x) = Meta.parse(sprint(show, x; context = (:module => @__MODULE__)))
```
Jutho: Since `parse_show` is always followed by an `eval()` call, we can maybe even include this here:

```suggestion
eval_parse_show(x) = eval(Meta.parse(sprint(show, x; context = (:module => @__MODULE__))))
```

This could also simply be called `eval_show(x)`.

test/runtests.jl
```diff
@@ -1,178 +1,72 @@
...
-# using SUNRepresentations: SUNIrrep
-# const SU3Irrep = SUNIrrep{3}
-using LinearAlgebra: LinearAlgebra
-using Zygote: Zygote
+# ARGS parsing
+# ------------
+using ArgParse: ArgParse
+using SafeTestsets: @safetestset
+# using CUDA: CUDA
+# using AMDGPU: AMDGPU
```
Jutho, line 6: This can be removed I guess?

```suggestion
```

test/runtests.jl
```diff
@@ -1,178 +1,72 @@
...
+    throw(ArgumentError("Invalid group ($group), no such folder"))
+foreach(checktestgroup, groups)
+
+# don't run all tests on GPU, only the GPU specific ones
+is_buildkite = get(ENV, "BUILDKITE", "false") == "true"
+
+# Run test groups
+# ---------------
+
+"match files of the form `*.jl`, but exclude `*setup*.jl`"
```
Jutho, line 43: Was this supposed to be a comment?
```suggestion
# match files of the form `*.jl`, but exclude `*setup*.jl`
```

test/tensors/planar.jl
```diff
@@ -1,74 +1,12 @@
-using TensorKit, TensorOperations, Test
-using TensorKit: BraidingTensor
+using Test, TestExtras
+using TensorKit
 using TensorKit: planaradd!, planartrace!, planarcontract!
 using TensorKit: PlanarTrivial, ℙ
```
Jutho: It would make sense to have the types before the methods:

```suggestion
using TensorKit: PlanarTrivial, ℙ
using TensorKit: planaradd!, planartrace!, planarcontract!
```

.github/workflows/CompatCheck.yml
```diff
@@ -20,8 +20,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        downgrade_mode: ['alldeps']
-        julia-version: ['1.10', '1']
+        downgrade_mode: ['deps']
+        julia-version: ['1.10', '1.12']
```
Jutho, line 26: Is the fixation to 1.12 necessary for a specific reason? We will forget about this in the future I am afraid.

Jutho:
Thanks for this; this looks great. I left a few small comments.

.github/workflows/CompatCheck.yml
lkdvos: Unfortunately yes: this is what was breaking the compat bounds checks before (https://github.com/julia-actions/julia-downgrade-compat/issues/25) I'll add todo/comment about this, but I'm not sure this will necessarily make us less likely to forget

test/runtests.jl
lkdvos: I left them here since the setup should already be there to handle the GPU tests, including the commented-out `CUDA.functional()` calls a little below. I'm happy to remove them, but I feel like it might be nice to just leave them there in anticipation of the actual tests.

test/runtests.jl
lkdvos: It was supposed to be a string, such that it is a docstring, but I'm happy to change it to a comment if that makes more sense to you

test/autodiff/ad.jl
lkdvos: I'm not sure, but I think for now it is fine to just leave this as-is. Given that `dim(c) >= 1` I think this never really poses any fundamental issues, `truncrank` with multiple sectors doesn't really allow you control over the distribution of sectors anyways so I think this should be granular enough. (I think most use-cases here are just to limit the total dimension, which is probably fine if we are off-by-one). There is always `truncspace` for more control?

test/runtests.jl
Jutho: Sure, that makes sense indeed.

Jutho (approved):

PR #299, opened 2025-10-10
Start on GPU extensions

kshyatt:
We have a couple `@allowscalar`s in here to allow truncation, ideally we'd remove this as we go. The AMDGPU extension needs work as well.


Project.toml
```diff
@@ -39,26 +53,31 @@ PackageExtensionCompat = "1"
 Random = "1"
 ScopedValues = "1.3.0"
 Strided = "2"
-TensorKitSectors = "0.1.4, 0.2"
+TensorKitSectors = "0.3"
```
lkdvos:
```suggestion
TensorKitSectors = "0.1.4, 0.2"
```
I don't think this update is ready yet 😉

Issue #301, opened 2025-10-10
Verbosity of Base.show for TensorMap


AFeuerpfeil:
When running code interactively, the printing of bigger TensorMaps can be annoying because it clutters the REPL, moves previous results or sometimes even takes minutes to just print a TensorMap.

I know the printing of a TensorMap is definitely debatable. For me personally, the information like all the `FusionTrees` and block entries are rarely useful, if I am not debugging.

The solution I chose in my fork is to have a minimialistic `Base.show` that only prints the space, whilst `Base.display` shows all the information.

PR #304, opened 2025-10-17
Show improvements + `subblocks` iterator

lkdvos:
These are various (WIP) opinionated show improvements, which I think improve overall quality of life. See also #301.

In the examples below, I tried to stick to `show(io, obj)` being parse-able, while `show(io, MIME"text/plain"(), obj)` being read-able. The latter is what gets shown when you show something directly to the terminal, the former is the fallback, which for example shows up when you print arrays of these objects.

Examples:

### Graded spaces

I've limited the output of the sectors, and made the printing closer to how `Base.Vector` would print:

<details><summary>Before</summary>

```julia
julia> V = U1Space(i => 1 for i in 1:100)
Rep[U₁](1=>1, 2=>1, 3=>1, 4=>1, 5=>1, 6=>1, 7=>1, 8=>1, 9=>1, 10=>1, 11=>1, 12=>1, 13=>1, 14=>1, 15=>1, 16=>1, 17=>1, 18=>1, 19=>1, 20=>1, 21=>1, 22=>1, 23=>1, 24=>1, 25=>1, 26=>1, 27=>1, 28=>1, 29=>1, 30=>1, 31=>1, 32=>1, 33=>1, 34=>1, 35=>1, 36=>1, 37=>1, 38=>1, 39=>1, 40=>1, 41=>1, 42=>1, 43=>1, 44=>1, 45=>1, 46=>1, 47=>1, 48=>1, 49=>1, 50=>1, 51=>1, 52=>1, 53=>1, 54=>1, 55=>1, 56=>1, 57=>1, 58=>1, 59=>1, 60=>1, 61=>1, 62=>1, 63=>1, 64=>1, 65=>1, 66=>1, 67=>1, 68=>1, 69=>1, 70=>1, 71=>1, 72=>1, 73=>1, 74=>1, 75=>1, 76=>1, 77=>1, 78=>1, 79=>1, 80=>1, 81=>1, 82=>1, 83=>1, 84=>1, 85=>1, 86=>1, 87=>1, 88=>1, 89=>1, 90=>1, 91=>1, 92=>1, 93=>1, 94=>1, 95=>1, 96=>1, 97=>1, 98=>1, 99=>1, 100=>1)

julia> [V]
1-element Vector{GradedSpace{U1Irrep, TensorKit.SortedVectorDict{U1Irrep, Int64}}}:
 Rep[U₁](1=>1, 2=>1, 3=>1, 4=>1, 5=>1, 6=>1, 7=>1, 8=>1, 9=>1, 10=>1, 11=>1, 12=>1, 13=>1, 14=>1, 15=>1, 16=>1, 17=>1, 18=>1, 19=>1, 20=>1, 21=>1, 22=>1, 23=>1, 24=>1, 25=>1, 26=>1, 27=>1, 28=>1, 29=>1, 30=>1, 31=>1, 32=>1, 33=>1, 34=>1, 35=>1, 36=>1, 37=>1, 38=>1, 39=>1, 40=>1, 41=>1, 42=>1, 43=>1, 44=>1, 45=>1, 46=>1, 47=>1, 48=>1, 49=>1, 50=>1, 51=>1, 52=>1, 53=>1, 54=>1, 55=>1, 56=>1, 57=>1, 58=>1, 59=>1, 60=>1, 61=>1, 62=>1, 63=>1, 64=>1, 65=>1, 66=>1, 67=>1, 68=>1, 69=>1, 70=>1, 71=>1, 72=>1, 73=>1, 74=>1, 75=>1, 76=>1, 77=>1, 78=>1, 79=>1, 80=>1, 81=>1, 82=>1, 83=>1, 84=>1, 85=>1, 86=>1, 87=>1, 88=>1, 89=>1, 90=>1, 91=>1, 92=>1, 93=>1, 94=>1, 95=>1, 96=>1, 97=>1, 98=>1, 99=>1, 100=>1)

julia> V'
Rep[U₁](1=>1, 2=>1, 3=>1, 4=>1, 5=>1, 6=>1, 7=>1, 8=>1, 9=>1, 10=>1, 11=>1, 12=>1, 13=>1, 14=>1, 15=>1, 16=>1, 17=>1, 18=>1, 19=>1, 20=>1, 21=>1, 22=>1, 23=>1, 24=>1, 25=>1, 26=>1, 27=>1, 28=>1, 29=>1, 30=>1, 31=>1, 32=>1, 33=>1, 34=>1, 35=>1, 36=>1, 37=>1, 38=>1, 39=>1, 40=>1, 41=>1, 42=>1, 43=>1, 44=>1, 45=>1, 46=>1, 47=>1, 48=>1, 49=>1, 50=>1, 51=>1, 52=>1, 53=>1, 54=>1, 55=>1, 56=>1, 57=>1, 58=>1, 59=>1, 60=>1, 61=>1, 62=>1, 63=>1, 64=>1, 65=>1, 66=>1, 67=>1, 68=>1, 69=>1, 70=>1, 71=>1, 72=>1, 73=>1, 74=>1, 75=>1, 76=>1, 77=>1, 78=>1, 79=>1, 80=>1, 81=>1, 82=>1, 83=>1, 84=>1, 85=>1, 86=>1, 87=>1, 88=>1, 89=>1, 90=>1, 91=>1, 92=>1, 93=>1, 94=>1, 95=>1, 96=>1, 97=>1, 98=>1, 99=>1, 100=>1)'
```

</details>

<details><summary>After</summary>

```julia
julia> V = U1Space(i => 1 for i in 1:100)
100-dimensional Rep[U₁]:
 (1 => 1, 2 => 1, 3 => 1, 4 => 1  …  97 => 1, 98 => 1, 99 => 1, 100 => 1)

julia> [V]
1-element Vector{GradedSpace{U1Irrep, TensorKit.SortedVectorDict{U1Irrep, Int64}}}:
 (1 => 1, 2 => 1, 3 => 1, 4 => 1  …  97 => 1, 98 => 1, 99 => 1, 100 => 1)

julia> Vect[FibonacciAnyon](:I => 1, :τ => 3)
5.85-dimensional Vect[FibonacciAnyon]:
(:I => 1, :τ => 3)

julia> V'
100-dimensional dual Rep[U₁]:
 (1 => 1, 2 => 1, 3 => 1, 4 => 1  …  97 => 1, 98 => 1, 99 => 1, 100 => 1)'

julia> show(V)
Rep[U₁](1 => 1, 2 => 1, 3 => 1, 4 => 1, 5 => 1, 6 => 1, 7 => 1, 8 => 1, 9 => 1, 10 => 1, 11 => 1, 12 => 1, 13 => 1, 14 => 1, 15 => 1, 16 => 1, 17 => 1, 18 => 1, 19 => 1, 20 => 1, 21 => 1, 22 => 1, 23 => 1, 24 => 1, 25 => 1, 26 => 1, 27 => 1, 28 => 1, 29 => 1, 30 => 1, 31 => 1, 32 => 1, 33 => 1, 34 => 1, 35 => 1, 36 => 1, 37 => 1, 38 => 1, 39 => 1, 40 => 1, 41 => 1, 42 => 1, 43 => 1, 44 => 1, 45 => 1, 46 => 1, 47 => 1, 48 => 1, 49 => 1, 50 => 1, 51 => 1, 52 => 1, 53 => 1, 54 => 1, 55 => 1, 56 => 1, 57 => 1, 58 => 1, 59 => 1, 60 => 1, 61 => 1, 62 => 1, 63 => 1, 64 => 1, 65 => 1, 66 => 1, 67 => 1, 68 => 1, 69 => 1, 70 => 1, 71 => 1, 72 => 1, 73 => 1, 74 => 1, 75 => 1, 76 => 1, 77 => 1, 78 => 1, 79 => 1, 80 => 1, 81 => 1, 82 => 1, 83 => 1, 84 => 1, 85 => 1, 86 => 1, 87 => 1, 88 => 1, 89 => 1, 90 => 1, 91 => 1, 92 => 1, 93 => 1, 94 => 1, 95 => 1, 96 => 1, 97 => 1, 98 => 1, 99 => 1, 100 => 1)
```

</details>

### Tensors

I've reworked the logic to condense some of the information, and moved printing the actual data to `blocks(t)`:

<details><summary>Before</summary>

```julia
julia> V = U1Space(i => rand(1:3) for i in 0:5);

julia> t = rand(V, V)
TensorMap(Rep[U₁](0=>1, 1=>1, 2=>1, 3=>3, 4=>1, 5=>3) ← Rep[U₁](0=>1, 1=>1, 2=>1, 3=>3, 4=>1, 5=>3)):
* Data for sector (Irrep[U₁](0),) ← (Irrep[U₁](0),):
 0.956618600870662
* Data for sector (Irrep[U₁](1),) ← (Irrep[U₁](1),):
 0.6880311699625894
* Data for sector (Irrep[U₁](2),) ← (Irrep[U₁](2),):
 0.5563263207815221
* Data for sector (Irrep[U₁](3),) ← (Irrep[U₁](3),):
 0.14172082577439749  0.9580819936113354    0.6703796017228667
 0.5327247384184031   0.010840227865541063  0.18311025114424562
 0.647277140465951    0.5369310096957065    0.19668786420897533
* Data for sector (Irrep[U₁](4),) ← (Irrep[U₁](4),):
 0.9502803020781383
* Data for sector (Irrep[U₁](5),) ← (Irrep[U₁](5),):
 0.9819862934340051    0.656975455417457   0.39031821044170756
 0.047434484466834625  0.5852165426518994  0.13176954314898348
 0.4101277094266894    0.6670030177799315  0.10052476444143021

julia> show(t)
TensorMap(Rep[U₁](0=>1, 1=>1, 2=>1, 3=>3, 4=>1, 5=>3) ← Rep[U₁](0=>1, 1=>1, 2=>1, 3=>3, 4=>1, 5=>3)):
* Data for sector (Irrep[U₁](0),) ← (Irrep[U₁](0),):
 0.6076891684855057
* Data for sector (Irrep[U₁](1),) ← (Irrep[U₁](1),):
 0.2738454610652815
* Data for sector (Irrep[U₁](2),) ← (Irrep[U₁](2),):
 0.590338319868511
* Data for sector (Irrep[U₁](3),) ← (Irrep[U₁](3),):
 0.9386484968436432  0.1546786733998191   0.5270779461416983
 0.5407179124579746  0.22422687207512249  0.3731970904347903
 0.6644337142933772  0.9491369695309702   0.7382820196608783
* Data for sector (Irrep[U₁](4),) ← (Irrep[U₁](4),):
 0.29592164796649245
* Data for sector (Irrep[U₁](5),) ← (Irrep[U₁](5),):
 0.30273515911799886  0.9038984672950175  0.6230474071503896
 0.40121087521799204  0.6100581083632737  0.045265328524538684
 0.5821099786441513   0.0462371916745955  0.2785155651386584

julia> eval(Meta.parse(sprint(show, t))) == t
ERROR: ParseError("extra token after end of expression")

julia> t = rand(V, V^2)
TensorMap(Rep[U₁](0=>1, 1=>1, 2=>1, 3=>3, 4=>1, 5=>3) ← (Rep[U₁](0=>1, 1=>1, 2=>1, 3=>3, 4=>1, 5=>3) ⊗ Rep[U₁](0=>1, 1=>1, 2=>1, 3=>3, 4=>1, 5=>3))):
* Data for sector (Irrep[U₁](0),) ← (Irrep[U₁](0), Irrep[U₁](0)):
[:, :, 1] =
 0.5666384420460484
* Data for sector (Irrep[U₁](1),) ← (Irrep[U₁](1), Irrep[U₁](0)):
[:, :, 1] =
 0.14475463836150493
* Data for sector (Irrep[U₁](1),) ← (Irrep[U₁](0), Irrep[U₁](1)):
[:, :, 1] =
 0.2249012003492179
* Data for sector (Irrep[U₁](2),) ← (Irrep[U₁](2), Irrep[U₁](0)):
[:, :, 1] =
 0.625498226574983
* Data for sector (Irrep[U₁](2),) ← (Irrep[U₁](1), Irrep[U₁](1)):
[:, :, 1] =
 0.12383739056493115
* Data for sector (Irrep[U₁](2),) ← (Irrep[U₁](0), Irrep[U₁](2)):
[:, :, 1] =
 0.1277737543797971
* Data for sector (Irrep[U₁](3),) ← (Irrep[U₁](3), Irrep[U₁](0)):
[:, :, 1] =
 0.8318555409273835   0.5939220769783348  0.714737024138462
 0.28193498695855346  0.3870108383932126  0.8311473803373162
 0.1908200693733898   0.6807316482626017  0.8038226428582901
* Data for sector (Irrep[U₁](3),) ← (Irrep[U₁](2), Irrep[U₁](1)):
[:, :, 1] =
 0.7895118457603167
 0.7631455165161088
 0.905604458975889
* Data for sector (Irrep[U₁](3),) ← (Irrep[U₁](1), Irrep[U₁](2)):
[:, :, 1] =
 0.19791450600183214
 0.01819643221870204
 0.14949751512376563
* Data for sector (Irrep[U₁](3),) ← (Irrep[U₁](0), Irrep[U₁](3)):
[:, :, 1] =
 0.6939691475636409
 0.21267003889730296
 0.2463692586921632

[:, :, 2] =
 0.20910259596135516
 0.36180523561901223
 0.6405866862351725

[:, :, 3] =
 0.44271783605601267
 0.41698433410714864
 0.2206634185456401
* Data for sector (Irrep[U₁](4),) ← (Irrep[U₁](4), Irrep[U₁](0)):
[:, :, 1] =
 0.4264392098795434
* Data for sector (Irrep[U₁](4),) ← (Irrep[U₁](3), Irrep[U₁](1)):
[:, :, 1] =
 0.00756067259883253  0.6539225068915775  0.25154536710467845
* Data for sector (Irrep[U₁](4),) ← (Irrep[U₁](2), Irrep[U₁](2)):
[:, :, 1] =
 0.2938907063748917
* Data for sector (Irrep[U₁](4),) ← (Irrep[U₁](1), Irrep[U₁](3)):
[:, :, 1] =
 0.345730580746735

[:, :, 2] =
 0.34142153430182787

[:, :, 3] =
 0.8090715031144456
* Data for sector (Irrep[U₁](4),) ← (Irrep[U₁](0), Irrep[U₁](4)):
[:, :, 1] =
 0.32129034798402445
* Data for sector (Irrep[U₁](5),) ← (Irrep[U₁](5), Irrep[U₁](0)):
[:, :, 1] =
 0.9971489141528406  0.9105834028013807   0.5333229638531825
 0.2574486953418622  0.4741899738124469   0.24857927771650168
 0.4896483412779581  0.03821790471052733  0.6804229018839635
* Data for sector (Irrep[U₁](5),) ← (Irrep[U₁](4), Irrep[U₁](1)):
[:, :, 1] =
 0.7168649333693414
 0.3235449911735686
 0.36021570969842587
* Data for sector (Irrep[U₁](5),) ← (Irrep[U₁](3), Irrep[U₁](2)):
[:, :, 1] =
 0.3946331588493833  0.17246984546352362  0.7937782983038535
 0.78787176599615    0.8386501967676255   0.8480138229512463
 0.6295917166582571  0.2971062754917273   0.19933203949092682
* Data for sector (Irrep[U₁](5),) ← (Irrep[U₁](2), Irrep[U₁](3)):
[:, :, 1] =
 0.15955470027368357
 0.08010702993426344
 0.22665875546031766

[:, :, 2] =
 0.07518711652476784
 0.2551308860396365
 0.02432166081731424

[:, :, 3] =
 0.0035578165165719433
 0.202377477304298
 0.10023799271363665
* Data for sector (Irrep[U₁](5),) ← (Irrep[U₁](1), Irrep[U₁](4)):
[:, :, 1] =
 0.5492720764972484
 0.9698841631437178
 0.5417353184198967
* Data for sector (Irrep[U₁](5),) ← (Irrep[U₁](0), Irrep[U₁](5)):
[:, :, 1] =
 0.9971684720382596
 0.08199422732278261
 0.24292479037746895

[:, :, 2] =
 0.2510089648946028
 0.5109760601553232
 0.4288262115427397

[:, :, 3] =
 0.9180936095878203
 0.65346415432928
 0.21382804192730054
```

</details>
<details><summary>After</summary>

```julia
julia> t = rand(V, V)
10←10 TensorMap{Float64, Rep[U₁], 1, 1, Vector{Float64}}:
  Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3) ← Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3)

julia> blocks(t)
blocks(::TensorMap{Float64, Rep[U₁], 1, 1, Vector{Float64}}):
 * Irrep[U₁](0) => 1×1 reshape(view(::Vector{Float64}, 1:1), 1, 1) with eltype Float64:
 0.1880971436651535

 * Irrep[U₁](1) => 1×1 reshape(view(::Vector{Float64}, 2:2), 1, 1) with eltype Float64:
 0.13790732649297244

 * Irrep[U₁](2) => 1×1 reshape(view(::Vector{Float64}, 3:3), 1, 1) with eltype Float64:
 0.4851984936122775

 * Irrep[U₁](3) => 3×3 reshape(view(::Vector{Float64}, 4:12), 3, 3) with eltype Float64:
 0.089535   0.978043  0.35284
 0.0866919  0.783924  0.45729
 0.645918   0.649325  0.11163

 * Irrep[U₁](4) => 1×1 reshape(view(::Vector{Float64}, 13:13), 1, 1) with eltype Float64:
 0.5902426467105624

 * Irrep[U₁](5) => 3×3 reshape(view(::Vector{Float64}, 14:22), 3, 3) with eltype Float64:
 0.687265  0.375243   0.22203
 0.481465  0.0316258  0.0937386
 0.311796  0.102495   0.0651239

julia> t'
10←10 adjoint(::TensorMap{Float64, Rep[U₁], 1, 1, Vector{Float64}}):
  Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3) ← Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3)

julia> blocks(t')
blocks(adjoint(::TensorMap{Float64, Rep[U₁], 1, 1, Vector{Float64}})):
 * Irrep[U₁](0) => 1×1 adjoint(reshape(view(::Vector{Float64}, 1:1), 1, 1)) with eltype Float64:
 0.1880971436651535

 * Irrep[U₁](1) => 1×1 adjoint(reshape(view(::Vector{Float64}, 2:2), 1, 1)) with eltype Float64:
 0.13790732649297244

 * Irrep[U₁](2) => 1×1 adjoint(reshape(view(::Vector{Float64}, 3:3), 1, 1)) with eltype Float64:
 0.4851984936122775

 * Irrep[U₁](3) => 3×3 adjoint(reshape(view(::Vector{Float64}, 4:12), 3, 3)) with eltype Float64:
 0.089535  0.0866919  0.645918
 0.978043  0.783924   0.649325
 0.35284   0.45729    0.11163

 * Irrep[U₁](4) => 1×1 adjoint(reshape(view(::Vector{Float64}, 13:13), 1, 1)) with eltype Float64:
 0.5902426467105624

 * Irrep[U₁](5) => 3×3 adjoint(reshape(view(::Vector{Float64}, 14:22), 3, 3)) with eltype Float64:
 0.687265  0.481465   0.311796
 0.375243  0.0316258  0.102495
 0.22203   0.0937386  0.0651239

julia> show(t)
TensorMap{Float64, Rep[U₁], 1, 1, Vector{Float64}}([0.1880971436651535, 0.13790732649297244, 0.4851984936122775, 0.08953497897401397, 0.08669186493167402, 0.6459180837983975, 0.9780429123177408, 0.7839241940197897, 0.6493251624623785, 0.3528402114107714, 0.45728993715004296, 0.11163011526213806, 0.5902426467105624, 0.6872645545383654, 0.4814646082531635, 0.31179600160823784, 0.3752434171811483, 0.031625813165556016, 0.10249535102737306, 0.22202964030310857, 0.09373862530409927, 0.06512390979102278], Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3) ← Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3))

julia> eval(Meta.parse(sprint(show,t))) == t
true

julia> t = rand(V, V^2)
10←10×10 TensorMap{Float64, Rep[U₁], 1, 2, Vector{Float64}}:
  Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3) ← (Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3) ⊗ Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3))

julia> blocks(t)
blocks(::TensorMap{Float64, Rep[U₁], 1, 2, Vector{Float64}}):
 * Irrep[U₁](0) => 1×1 reshape(view(::Vector{Float64}, 1:1), 1, 1) with eltype Float64:
 0.3207401165483481

 * Irrep[U₁](1) => 1×2 reshape(view(::Vector{Float64}, 2:3), 1, 2) with eltype Float64:
 0.603692  0.107051

 * Irrep[U₁](2) => 1×3 reshape(view(::Vector{Float64}, 4:6), 1, 3) with eltype Float64:
 0.083302  0.969771  0.745582

 * Irrep[U₁](3) => 3×8 reshape(view(::Vector{Float64}, 7:30), 3, 8) with eltype Float64:
 0.294564  0.437507  0.629868  0.732419  0.813652  0.984176  0.185027  0.434976
 0.827017  0.593513  0.622622  0.701299  0.349647  0.4101    0.786329  0.639461
 0.712637  0.450134  0.98193   0.160072  0.22098   0.248266  0.183098  0.0125587

 * Irrep[U₁](4) => 1×9 reshape(view(::Vector{Float64}, 31:39), 1, 9) with eltype Float64:
 0.362816  0.782461  0.335289  0.918599  0.515771  0.352911  0.956776  0.569801  0.653597

 * Irrep[U₁](5) => 3×14 reshape(view(::Vector{Float64}, 40:81), 3, 14) with eltype Float64:
 0.353537  0.200091  0.0498139  0.509441  0.837209  0.85076   0.426807  0.417218  0.798341  0.086593    0.702123  0.0316201  0.0633773  0.293837
 0.169678  0.166921  0.738124   0.439812  0.536551  0.777772  0.96886   0.662034  0.440518  0.00327853  0.431701  0.461115   0.488631   0.745294
 0.959431  0.951188  0.876941   0.855064  0.608388  0.349781  0.396045  0.830495  0.30937   0.89072     0.912777  0.626573   0.991999   0.807611

julia> d
10←10 DiagonalTensorMap{Float64, Rep[U₁], Vector{Float64}}:
  Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3) ← Rep[U₁](0 => 1, 1 => 1, 2 => 1, 3 => 3, 4 => 1, 5 => 3)

julia> blocks(d)
blocks(::DiagonalTensorMap{Float64, Rep[U₁], Vector{Float64}}):
 * Irrep[U₁](0) => 1×1 LinearAlgebra.Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}:
 0.40985397792974165

 * Irrep[U₁](1) => 1×1 LinearAlgebra.Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}:
 0.2531474800986825

 * Irrep[U₁](2) => 1×1 LinearAlgebra.Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}:
 0.24037982912886158

 * Irrep[U₁](3) => 3×3 LinearAlgebra.Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}:
 0.343918   ⋅         ⋅
  ⋅        0.720022   ⋅
  ⋅         ⋅        0.72227

 * Irrep[U₁](4) => 1×1 LinearAlgebra.Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}:
 0.8770955222992027

 * Irrep[U₁](5) => 3×3 LinearAlgebra.Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}:
 0.8177   ⋅         ⋅
  ⋅      0.996111   ⋅
  ⋅       ⋅        0.807667
```

</details>

I would argue that already this is showing both more information and using a whole lot less space, but obviously this is subjective. Additionally, I quite like how clean the implementation is overall.
One thing I might still add is `subblocks(t)`, as a counterpart of `blocks(t)`, which would be equivalent to `((f1, f2) => t[f1, f2] for (f1, f2) in fusiontrees(t))`, such that I can also overload the show there and have the old behavior as an option.
I also still need to tackle the adjoint and diagonal tensors :)

### DIFF
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -656,7 +656,7 @@ function Base.show(io::IO, ::MIME"text/plain", t::AbstractTensorMap)
     # 3) [optional]: show data
     get(io, :compact, true) && return nothing
     ioc = IOContext(io, :typeinfo => sectortype(t))
-    println(io, "\n\n blocks(t):")
+    println(io, "\n\n blocks: ")
     show_blocks(io, MIME"text/plain"(), blocks(t))
     return nothing
 end


Jutho:
> I also still need to tackle the

You are keeping us in suspense ...

I very much like the tensor printing improvements. The space printing I have no strong opinions about; I don't think the new scheme satisfies the "uses less space" criterion, but if this is more readable then I am definitely ok with this. I liked the aspect that copy-paste of standard REPL output yielded valid code, but I don't know how much I really rely on this.

lkdvos:
I'm also the least hung up on the graded spaces to be honest. Note that the main point was that the default printing truncates overly long outputs, and scales this with your current screen size, not necessarily that this is now vertical instead of horizontal.

The REPL output being parse-able is still something that we have, you just have to call the default `show`:

```julia
julia> show(V)
Rep[U₁](1 => 1, 2 => 1, 3 => 1, 4 => 1, 5 => 1, 6 => 1, 7 => 1, 8 => 1, 9 => 1, 10 => 1, 11 => 1, 12 => 1, 13 => 1, 14 => 1, 15 => 1, 16 => 1, 17 => 1, 18 => 1, 19 => 1, 20 => 1, 21 => 1, 22 => 1, 23 => 1, 24 => 1, 25 => 1, 26 => 1, 27 => 1, 28 => 1, 29 => 1, 30 => 1, 31 => 1, 32 => 1, 33 => 1, 34 => 1, 35 => 1, 36 => 1, 37 => 1, 38 => 1, 39 => 1, 40 => 1, 41 => 1, 42 => 1, 43 => 1, 44 => 1, 45 => 1, 46 => 1, 47 => 1, 48 => 1, 49 => 1, 50 => 1, 51 => 1, 52 => 1, 53 => 1, 54 => 1, 55 => 1, 56 => 1, 57 => 1, 58 => 1, 59 => 1, 60 => 1, 61 => 1, 62 => 1, 63 => 1, 64 => 1, 65 => 1, 66 => 1, 67 => 1, 68 => 1, 69 => 1, 70 => 1, 71 => 1, 72 => 1, 73 => 1, 74 => 1, 75 => 1, 76 => 1, 77 => 1, 78 => 1, 79 => 1, 80 => 1, 81 => 1, 82 => 1, 83 => 1, 84 => 1, 85 => 1, 86 => 1, 87 => 1, 88 => 1, 89 => 1, 90 => 1, 91 => 1, 92 => 1, 93 => 1, 94 => 1, 95 => 1, 96 => 1, 97 => 1, 98 => 1, 99 => 1, 100 => 1)
```

I'm happy to remove the vertical printing and just do the horizontal truncated outputs instead, if that makes more sense? I would like to keep the little summary at the top though, the `d-dimensional dual spacetype:` actually gives quite a lot of information at a glance already.

src/spaces/gradedspace.jl
```diff
@@ -197,24 +197,72 @@ function supremum(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I <: Sector
...
+    return nothing
+end
+
+function Base.show(io::IO, V::GradedSpace)
+    pre = (get(io, :typeinfo, Any)::DataType == typeof(V) ? "" : type_repr(typeof(V)))
+
+    io = IOContext(io, :typeinfo => Pair{sectortype(V), Int})
+
+    pre *= "("
+    post = isdual(V) ? ")'" : ")"
```
Jutho: How about replacing `V` by `dual(V)` if `isdual(V)`, then you do not need to think about dual spaces anymore in the rest of this function.

lkdvos: I'm not sure I fully understood what you meant, but I attempted to add this.

src/spaces/gradedspace.jl
```diff
@@ -197,24 +197,83 @@ function supremum(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I <: Sector
...
+    isdual(V) && print(io, ")")
+    return nothing
+end
+
+function Base.summary(io::IO, V::GradedSpace)
+    d = reduceddim(V)
+    print(io, d, "-element ")
+    Base.showarg(io, V, true)
+    return nothing
+end
```
Jutho: I am not too fond of this one. Our spaces are supposed to be simple types that generalize just using integers for dimensions. It seems that for most simply types in Base, e.g. for all `Number` subtypes, `Base.summary` just print out their type.

Jutho: Also, the `n-element SomeType` summary is used for arrays, where `n` then refers to the number of entries in the array. So while I support something like this in the `summary` string of a `TensorMap`, it doesn't really make sense in `GradedSpace`. If anything like this appears, it would just need to be the number of sectors present. But I would simply print the type for the summary (which is probably the Julia default and doesn't even need a definition), or we could simply have

```julia
Base.summary(V::ElementarySpace) = type_repr(typeof(V))
```

lkdvos:
@Jutho this should now also be ready for review. I think overall this is a lot more convenient, and we should probably include this before the docs rewrite as well

src/spaces/gradedspace.jl
```diff
@@ -197,21 +197,75 @@ function supremum(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I <: Sector
...
+
+function Base.show(io::IO, ::MIME"text/plain", V::GradedSpace)
+    # print small summary, e.g.
+    # d-element Vect[I] or d-element dual(Vect[I])
+    d = reduceddim(V)
+    print(io, d, "-element ")
+    isdual(V) && print(io, "dual(")
+    print(io, type_repr(typeof(V)))
+    isdual(V) && print(io, ")")
```
Jutho: I still think this `d-element` printing style is not really appropriate. When is this method called; is this in notebooks?

How about some printing style for an object like `V = SU2Space(0=>3, 1=>2, 2=>1)`:

"Rep[SU₂] object with dimension 14 (reduced dimension 6)"

or for V'

"Rep[SU₂] dual object with dimension 14 (reduced dimension 6)"

where also the extra information in parentheses can be omitted if `dim(V) == reduceddim(V)`.

Just a suggestion, let me know what you think.

src/tensors/abstracttensor.jl
```diff
@@ -295,6 +301,145 @@ function blocktype(::Type{T}) where {T <: AbstractTensorMap}
...
+See also [`subblock`](@ref), [`fusiontrees`](@ref), and [`hassubblock`](@ref).
+"""
+subblocks(t::AbstractTensorMap) = SubblockIterator(t, fusiontrees(t))
+
+const _doc_subblock = """
+Return a view into the data of `t` corresponding to the splitting - fusion tree pair
+`(f₁, f₂)`. In particular, this is an `AbstractArray{T}` with `T = scalartype(t)`, of size
+`(dims(codomain(t), f₁.uncoupled)..., dims(codomain(t), f₂.uncoupled)...)`.
+
+Whenever `FusionStyle(sectortype(t))`, it is also possible to provide only the external
```
Jutho: This sentence is missing a part about being `UniqueFusion`

src/tensors/abstracttensor.jl
```diff
@@ -295,6 +301,145 @@ function blocktype(::Type{T}) where {T <: AbstractTensorMap}
...
+    into the tensor data. As a result, `getindex` and `view` have the same behavior.
+
+See also [`subblock`](@ref), [`subblocks`](@ref) and [`fusiontrees`](@ref).
+""" Base.view(::AbstractTensorMap, ::Tuple{I, Vararg{I}}) where {I <: Sector},
+    Base.view(::AbstractTensorMap, ::FusionTree, ::FusionTree)
+
+@inline Base.view(t::AbstractTensorMap, sectors::Tuple{I, Vararg{I}}) where {I <: Sector} =
+    subblock(t, sectors)
+@inline Base.view(t::AbstractTensorMap, f₁::FusionTree, f₂::FusionTree) =
+    subblock(t, (f₁, f₂))
```
Jutho: Do you think it is useful to have `view` methods? I kind of consider this to be `AbstractArray` specific. I am a bit hesitant to provide to many alternative interfaces.

src/tensors/abstracttensor.jl
```diff
@@ -499,3 +648,33 @@ function Base.convert(::Type{Array}, t::AbstractTensorMap)
...
+# ----------------
+Base.dims2string(V::HomSpace) = join(dim.(codomain(V)), '×') * "←" * join(dim.(domain(V)), '×')
+
+Base.summary(io::IO, t::AbstractTensorMap) = tensor_summary(io, t, space(t))
+
+function tensor_summary(io::IO, t, V)
+    print(io, Base.dims2string(V), " ")
+    Base.showarg(io, t, true)
+    return nothing
+end
```
Jutho: Is it useful to split this up and not just define `Base.summary` in this way? There don't seem to be any other call sites of `tensor_summary`. Is there any use case for not having `V = space(t)` ?

src/tensors/abstracttensor.jl
```diff
@@ -480,6 +625,10 @@ end
 
 # Conversion to Array:
 #----------------------
+Base.ndims(t::AbstractTensorMap) = numind(t)
+Base.size(t::AbstractTensorMap) = ntuple(Base.Fix1(size, t), numind(t))
+Base.size(t::AbstractTensorMap, i) = dim(space(t, i))
```
Jutho: I am still not convinced about the usefulness of this. This is mostly aimed at making it more friendly for newcomers? But shouldn't we stimulate them to just learn from the start that `TensorMap`s are not just `AbstractArray`s and that they should just check the documentation to learn about the interface. I am happy to for example have a summary of the interface in the README or whathever can help.

One reason why I think this might not be a good idea is that this forces `length(t) = prod(size(t)) = dim(codomain) * dim(domain)`. But there might be an argument to be made for `length(t) = dim(codomain ← domain) = dim(t)`. So once we establish this is part of our public interface, it will be very hard to revert this if we ever come up with a better use case for these definitions.

src/tensors/abstracttensor.jl
```diff
@@ -499,3 +648,33 @@ function Base.convert(::Type{Array}, t::AbstractTensorMap)
...
+
+# Human-readable:
+function Base.show(io::IO, ::MIME"text/plain", t::AbstractTensorMap)
+    # 1) show summary: typically d₁×d₂×… ← d₃×d₄×… $(typeof(t)):
+    summary(io, t)
+    println(io, ":")
+
+    # 2) show spaces
+    # println(io, " space(t):")
+    print(io, "  ", space(t))
```
Jutho: would it be an option to show domain and codomain separately?
```suggestion
    print(io, "codomain: ", codomain(t))
    print(io, "domain:   ", domain(t))
```

src/tensors/abstracttensor.jl
```diff
@@ -499,3 +648,33 @@ function Base.convert(::Type{Array}, t::AbstractTensorMap)
         return A
     end
 end
+
+# Show and friends
+# ----------------
+Base.dims2string(V::HomSpace) = join(dim.(codomain(V)), '×') * "←" * join(dim.(domain(V)), '×')
```
Jutho: This might need special casing for when `N1` or `N2` is 0.

src/spaces/gradedspace.jl
lkdvos: Doing another deep dive into the implementations and documentation, this is what I can find:

The different show methods are *mostly* used like this:
- `show(io, MIME"text/plain"(), x)` is what gets called mostly in interactive usage, i.e. it ends up being the implementation when you `display` something to an output that can support text. (I.e. REPL, or notebook output, ...). This is supposed to be for human consumption
- Container types generally implement 3-argument show by calling `show(io, MIME"text/plain"(), x)` for elements `x`, with `:compact => true` set. In this case you should avoid `\n` to not break the printing of the container.
- `:limit => true` can be set to print `…` in place of most elements in a container.
- `show(io, x)` is the default implementation/fallback, which is typically used for computer consumption (i.e. parseable)

For example `Array` handles this for the 3-arg version by writing a summary first, and then, on new line(s) prints the data. If `compact => true`, the summary is omitted and the `[x; y;; ...]` single-line notation is used to show the data. 

So, what I would like is:
- The 3-arg show for a space gives me a quick summary with the most important data, which is, in my opinion, the sectortype, some measure of size and the duality. Then, the sectors and dimensions can be used afterwards to fully specify the space
- The 3-arg show with `:compact => true` should not contain any `\n`. We can decide to either show `Vect[I](x => d, ...)` or show the summary, depending on what we think is the most useful information to show at a glance in a container of spaces. This is the thing that would get called e.g. when showing the spaces of an MPS, which you could think of as a container space

I don't actually feel to strongly about how this information is worded exactly, but I do think showing the dimension somewhere is a strict improvement, for example in MPS or PEPS methods, the relevant spaces to inspect are the virtual spaces, which tend to become completely unreadable when there are many sectors. In this case, simply truncating the sectors and going for the `…` reduces the screen clutter, but then there is no more information on the screen.

I have no real preference about the dim vs reduced dim, I argue that the information is mostly in comparing different values anyways (I want to see if my dimension increased or decreased during a DMRG run), but given that we are typically not using `reduceddim` in many places I feel like to not make this number confusing we would have to be a little explicit about it.

`"Rep[SU₂] dual object with dimension 14 (reduced dimension 6)"` seems a bit long, I think one of the dimensions is sufficient? I would also prefer to avoid the categoric "object" language when this isn't necessary, so maybe to suggest either alternative:

`"(dual) Rep[SU₂] of dim d"`
`"(dual) Rep[SU₂] of reduceddim d"`

This gives all information, would be `Float` dim-friendly since the floating point number is at the end, avoids (more) use of category language and is reasonably short. `dim` and `reduceddim` are idiomatic since we are already using that as a function name, so it cannot be confusing what this number is referring to.

src/tensors/abstracttensor.jl
lkdvos: I think if I would redesign everything right now, I would have gone with `view` instead of `getindex` to begin with, since we are technically performing "method-piracy" on `Base.getindex`, which is supposed to yield a copy.
I don't want to change this now (that would be really painful), but I do find it easier to say "we have made the decision to make `Base.getindex(::AbstractTensorMap)` default to behave like `Base.view(::AbstractTensorMap)`", since these two concepts (as words) already exist and have meaning to me.
This is more or less the only reason why I added this, I don't really see many people using it, but I also don't think it hurts to have this here.

src/tensors/abstracttensor.jl
lkdvos: I definitely agree with the sentiment to steer people towards our interface, but I don't think we currently have one for this exact concept, and `(dims(codomain(t))..., dims(domain(t))...)` is a bit less convenient in my opinion. Since I think it is quite obvious what `size` would return, it is a bit weird to me to get a `MethodError`. I don't think we as a library should be deciding if it is allowed to use `size` or not, and because of type piracy reasons this is basically the only place this method can be defined, so we might as well add it?

I have to say that I am not sure I agree that your reason is really related to this. I don't think `length(t) = prod(size(t))` is expected, and I don't see why we should hold off on defining `size` just because there might be other functions that could have multiple meanings.

Ultimately, it's mostly just about being pragmatic. It's useful functionality, it's a short, easily available function name that is easy to remember and it's clear what it does, and I see very little harm in overloading it

src/tensors/abstracttensor.jl
lkdvos: I just copied the logic from how `AbstractArray` does this.
I think there the idea is that you probably want to specialize the implementation on `axes(A)` instead of on the array type, but I don't think this is too relevant here so I'm happy to remove that layer if you prefer

src/spaces/gradedspace.jl
Jutho: The use of 'object' was supposed to be such that only category-minded people (like you) think about categorical objects 😄 , and other people think of a computer-science object of type `Rep[SU₂]`. Though the possible occurance of `dual` then spoils that a little bit. I just think the one guideline should be not to take `Array` as an example, as really spaces are supposed to be like small bits types (even if `GradedSpace` isn't really. Furthermore, `Array` still prints out all elements even with `:compact => true`, so by that guideline we can still print all the sectors. But I agree that this is not what you want to see when an MPS is printed. However, I do think it is strange if `ℂ^4` is printed just as this, and `Rep[SU₂](...)` is printed as `dual Rep[SU₂] of dim d`. I kind of dislike the implicit implication that it is an instance and not a type.

How about the following compromise: With `:compact => true`, a `V::GradedSpace` is printed as

`type_repr(typeof(V)) * "(…)" * (isdual(V) ? "'" : "") * " with dim $(dim(V))"` 

Depending on the length of the information that you want to show, you can add additional information such as `" with $(length(sectors(V)) sectors and dim $(dim(V))"` but this is really up to you.

Jutho: I am still confused by the `show` methods. If you just have an interactive REPL session, it goes via `display(V)` and ends up calling `show(stdout, MIME"text/plain", V)`. Is that than also the version that should check for `:limit`? So we might have `:compact=>false` but `:limit=>true`, so we would have a second line printing the sectors? In fact, for `show(stdout, MIME"text/plain", V)`, I wouldn't mind if the sectors are printed underneath each other as in your original design. 

The version without MIME type, `show(io, V)`, is only called if you call explicitly `show(V)`. That one, even for `Array`, prints out all information and is supposed to be parseable. So can that just remain the current version that prints out all the sectors? For `Array`, the only difference between `:compact` true or false in that case is that it truncates floating point numbers to less accuracy, but that doesn't really make sense for sector dimensions.

Jutho: Oh wait, `show(io, V)` does also listen to `:limit`. Pff, this is annoying.

lkdvos: I think (this is my interpretation, it's all very badly defined/documented) that `:compact` means no linebreaks and optionally more dense representations (less floating point digits, or `x=>y` instead of `x => y`), and `:limit` means respecting the `displaysize`.
The version without MIME type is only called if you explicitly `show(V)` or `print(io, V)`, where I think it is supposed to be parseable but can still be `:limit`ted if you like

src/tensors/abstracttensor.jl
Jutho: Would using `dims(::HomSpace)` be an adequate solution, or you really think users will want to use `size`? I have to say that this has never come up as an open issue on Github, but maybe users complain to you directly?

src/tensors/tensor.jl
```diff
@@ -495,105 +479,32 @@ See also [`Base.setindex!(::TensorMap{T,S,N₁,N₂}, ::Any, ::FusionTree{I,N₁
...
-end
-@propagate_inbounds function Base.getindex(t::TensorMap, sectors::Tuple)
-    return t[map(sectortype(t), sectors)]
-end
-
 # Show
 #------
-function Base.summary(io::IO, t::TensorMap)
-    return print(io, "TensorMap(", space(t), ")")
+function type_repr(::Type{TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A}
```
Jutho, line 495: `type_repr` comes from TensorKitSectors.jl, right? Do we import this somewhere? Or does TensorKit.jl has its own `type_repr` (which doesn't sound right).

lkdvos: We do: 

https://github.com/QuantumKitHub/TensorKit.jl/blob/9ba3b6c7ad4738ad9b25d6f2527a9a2767927a4a/src/TensorKit.jl#L122

Happy to qualify if you think that is more obvious, but I think it is really fair to say we "take ownership" of these functions in TensorKitSectors so this is kind of fine to me

Jutho: Ok thanks; I had checked before posing the question, but had overlooked (internet is working reasonably well but not superfast on the train). I am fine with keeping it like this.

src/tensors/abstracttensor.jl
lkdvos: I have had a couple people mention that they found it surprising that it doesn't work (technically, I quote: "this is why no one is using TensorKit")
I'm okay with `dims(::HomSpace)`, (can we also have `dims(::AbstractTensorMap)`?), but still have a hard time motivating "we didn't implement size because or tensor type would look to much like an array which we want to avoid as it might cause confusion". Maybe I'm just reading too much into it, but this seems like a very easy thing to fix and just provide?

src/tensors/abstracttensor.jl
Jutho: First an aside, your remark about the "method piracy" on `Base.getindex` is probably valid for `StridedView`, though in that case I do think it is well documented that slicing it will produce a view instead of a new copy.

For `AbstractTensorMap`, I do not agree, because I do not consider `Base.getindex` as a method that is specific to `AbstractArray`, and therefore I do not think it should follow the same rules. Many types have a `getindex` that have nothing to do with `AbstractArray`. If you would have some kind of dictionary that stores blocks as values with fusion tree pairs as keys, it would be very natural that indexing with the key would return the original block, without making a copy. I think this is the analogy I have in my head when I think of `(Abstract)TensorMap`, so I really don't think of them as `AbstractArray`s. It seems `ITensor`s have the same behavior: `T[Block((1,3))]` returns an `AbstractArray` subtype that is actually providing a view into the data.

In fact, I believe returning a view is necessary if we want some syntax like `T[(f1, f2)] .= 1` to work. Here, it becomes messy with `Array`, since `A[1:2, 1:3] .= value` will also overwrite the data, but I really have no idea what this is lowered to. I assume something like `setindex!.(Ref(A), value, ...)` where I don't know what `...` needs to be so that it is not considered as just a one-dimensional broadcast. But this anyways cannot work for our case.

So while I don't think our `TensorMap` `getindex` is problematic, I agree that using a separate method like `subblock` makes the distinction even more clear.

However, also recycling `view` for this is again hitting the fundamental disagreement we seem to have. I don't like to think of `(Abstract)TensorMap` as an `AbstractArray` subtype, and so I had opted not to use any methods that seem to be specific to [`AbstractArray`](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array) in Julia base (`ndims`, `size`, `axes`, `view`, ...) for `AbstractTensorMap`. However, `AbstractTensorMap` definitely is a container type, and so the [basic methods for general containers](https://docs.julialang.org/en/v1/manual/interfaces/#Indexing) (`getindex` and `setindex!`) are available to us, with the key types and behavior that seems appropriate for our use case.

It seems `ITensor`s are a bit in a mixed category: they listen to `size` and `ndims`, and you can index them with integer indices, but not with ranges and you cannot apply `view`. Do you have any insight into the user experience and typical user requests in that regard?

src/tensors/abstracttensor.jl
Jutho: Sure, if `dims(::HomSpace)` work, than `dims(::AbstractTensorMap)` must return the same. But I guess I will have to give up my fight against `size` then. Maybe we need an `allowsize` like `CUDA.allowscalar` ? 😄 .

Jutho: I especially look forward to the warning on the first call:
```julia-repl
julia >size(::TensorMap)
WARNING: a TensorMap is not an AbstractArray and has a different interface; expect non-integer dimensions and other unexpected behavior. RTFM!
```

src/tensors/abstracttensor.jl
lkdvos: I think most of the complaints surrounding `ITensor`s seem to land in the same category: the lack of support for slicing is the most-often quoted nuisance, and the opaque manner in which the data is stored when introducing symmetries is not great.
Overall, my feeling is that people tend to expect these things to "just work", and they care less about the performance than they do about the convenience of not having to alter their habits of how to code or interact with arrays. I would say that people are receptive to the "this works, but if you want a fast/better implementation you have to work a little for it"

src/tensors/abstracttensor.jl
lkdvos: Jokes aside, I would actually be quite happy with a `TensorKit.allowscalar` to allow scalar `getindex`/`setindex` for abelian and `getindex` for non-abelian tensors too 😁

src/spaces/gradedspace.jl
```diff
@@ -197,21 +197,53 @@ function supremum(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I <: Sector
...
+    io = IOContext(io, :typeinfo => eltype(v))
+
+    if limited && length(v) > 20
+        axs1 = axes(v, 1)
+        f, l = first(axs1), last(axs1)
+        Base.show_delim_array(io, v, opn, ",", "", false, f, f + 9)
+        print(io, "  …  ")
+        Base.show_delim_array(io, v, "", ",", cls, false, l - 9, l)
+    else
+        Base.show_delim_array(io, v, opn, ",", cls, false)
```
Jutho, line 225: I guess this is in principle a private function, though I am not opposed to using it. But it could easily be broken in a next Julia release.

src/tensors/abstracttensor.jl
```diff
@@ -499,3 +625,38 @@ function Base.convert(::Type{Array}, t::AbstractTensorMap)
...
+
+    # 2) show spaces
+    # println(io, " space(t):")
+    println(io, " codomain: ", codomain(t))
+    println(io, " domain: ", domain(t))
+
+    # 3) [optional]: show data
+    get(io, :compact, true) && return nothing
+    ioc = IOContext(io, :typeinfo => sectortype(t))
+    println(io, "\n\n blocks(t):")
```
Jutho: On the one hand I like showing that `blocks` is a function call, so that new users learn about this method, so I would like to propose the same for `domain` and `codomain` printing in step 2. On the other hand, the `t` name is completely arbitrary, and could confuse people that don't call tensors `t`. So I am a bit torn between those two choices. But I think not using an undefined `t` dominates in importance.

src/spaces/gradedspace.jl
lkdvos: Yeah, I agree that in principle we should probably vendor this ourselves. Unfortunately it is actually quite ugly/involved, so I would really prefer to not have to do so, and it might actually require less effort to simply update this function if Julia breaks it...

PR #305, opened 2025-10-26
Avoid copy in `twist` for tensors with bosonic braiding

leburgel:
When profiling some AD-based optimizations, I noticed the `twist`s that occur in the `rrule`s of the tensor operations show up quite prominently sometimes. This was caused by the fact that by default `twist` copies its input before calling `twist!`, even for tensors with bosonic braiding for which `twist!` does nothing in the end.

Here I just added the check for trivial twists from `twist!` to `twist` to avoid a copy for trivial twists. I don't know if anything really relies on the fact that `twist` makes a copy though. If this is the case, maybe there's a different way to handle trivial twists in the tensor operations pullbacks themselves?

### DIFF
--- a/src/spaces/gradedspace.jl
+++ b/src/spaces/gradedspace.jl
@@ -200,7 +200,7 @@ end
 Base.summary(io::IO, V::GradedSpace) = print(io, type_repr(typeof(V)))
 
 function Base.show(io::IO, V::GradedSpace)
-    opn = (get(io, :typeinfo, Any)::DataType == typeof(V) ? "" : type_repr(typeof(V)))
+    opn = (get(io, :typeinfo, Any)::Type == typeof(V) ? "" : type_repr(typeof(V)))
     opn *= "("
     if isdual(V)
         cls = ")'"
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -644,20 +644,26 @@ function Base.summary(io::IO, t::AbstractTensorMap)
 end
 
 # Human-readable:
-function Base.show(io::IO, ::MIME"text/plain", t::AbstractTensorMap)
-    # 1) show summary: typically d₁×d₂×… ← d₃×d₄×… $(typeof(t)):
+function Base.show(io::IO, mime::MIME"text/plain", t::AbstractTensorMap)
+    # 1) show summary: typically d₁×d₂×… ← d₃×d₄×… $(typeof(t))
     summary(io, t)
-    println(io, ":")
 
+    # case without `\n`:
+    if get(io, :compact, true)
+        print(io, "(…, ")
+        show(io, mime, space(t))
+        print(io, ')')
+        return nothing
+    end
+
+    # case with `\n`
     # 2) show spaces
-    # println(io, " space(t):")
+    println(io, ':')
     println(io, " codomain: ", codomain(t))
     println(io, " domain: ", domain(t))
 
     # 3) [optional]: show data
-    get(io, :compact, true) && return nothing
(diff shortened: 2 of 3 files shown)


lkdvos:
This feels like a pretty big breaking change and I'm a little scared of this introducing silent and subtle bugs. 

In principle, for the other "indexmanipulations" we have a `copy = false` keyword to indicate if data sharing with the input is allowed, which is probably the easiest solution. Unfortunately, if we then want to make this non-breaking we would have to set that to true by default, which is not in line with the other default values. 

The fastest way forward for tousle specific case is probably to just special case it in the reverse rules, which is in line with the forward rules special case. 

However, I think it would be nicest in the long run to just additionally include the keyword as a breaking change, and then possibly simplify the use in the contractions, with the next 0.16 release. I don't know what you think @Jutho ?

Jutho:
Does anyone understand how this change causes the autodiff tests to fail?

lkdvos:
I am actually guessing it's an RNG thing, it looks like these tests are still a bit brittle :(

src/tensors/indexmanipulations.jl
```diff
@@ -294,15 +294,19 @@ function twist!(t::AbstractTensorMap, is; inv::Bool = false)
...
 Apply a twist to the `i`th index of `tsrc` and return the result as a new tensor.
-If `inv=true`, use the inverse twist.
+If `inv = true`, use the inverse twist.
+If `copy = false`, `tdst` might share data with `tsrc` whenever possible. Otherwise, a copy is always made.
 
 See [`twist!`](@ref) for storing the result in place.
 """
-twist(t::AbstractTensorMap, i; inv::Bool = false) = twist!(copy(t), i; inv)
+function twist(t::AbstractTensorMap, is; inv::Bool = false, copy::Bool = false)
+    !copy && (BraidingStyle(sectortype(t)) == Bosonic() || isempty(is)) && return t
```
Jutho: This is becoming a bit long, and we can even add more cases:

```suggestion
    if !copy
        BraidingStyle(sectortype(t)) == Bosonic() && return t
        all(isunit ∘ Base.Fix1(space, t), is) && return t
    end
```

Jutho: the second condition contains `isempty(inds)`

Jutho: I've corrected the suggestion to `inds` -> `is`, although I actually prefer the slightly longer `inds` 😄

lkdvos: unfortunately `isunit` isn't defined for spaces right now. I'm happy to add it, but does that matter?

lkdvos: I factored out the code that is responsible for this, so we can easily add more cases later. I additionally caught the `NoBraiding` case I think, if we ever care about this :)

src/tensors/indexmanipulations.jl
```diff
@@ -262,47 +262,59 @@ To repartition into an existing destination, see [repartition!](@ref).
...
+    if BraidingStyle(I) == NoBraiding()
+        for i in inds
+            cs = sectors(space(t, i))
+            all(isone, cs) || throw(SectorMismatch(lazy"Cannot twist sectors $cs"))
+        end
+        return true
+    elseif BraidingStyle(I) == Bosonic()
+        return true
+    else
+        return isempty(inds)
```
Jutho: We could even have
```suggestion
        for i in inds
            cs = sectors(space(t, i))
            all(isone ∘ twist, cs) || return false
        end
        return true
```

Jutho (approved):

PR #306, opened 2025-10-27
avoid twist allocating in rrules when not required

lkdvos:
This is a temporary non-breaking solution before #305 fully resolves the issue, to already include in a patch release.

@leburgel tagging you here


ext/TensorKitChainRulesCoreExt/tensoroperations.jl
```diff
@@ -188,3 +188,10 @@ function ChainRulesCore.rrule(::typeof(TensorKit.scalar), t::AbstractTensorMap)
     end
     return val, scalar_pullback
 end
+
+# temporary function to avoid copies when not needed
+# TODO: remove once `twist(t; copy=false)` is defined
+function _twist_nocopy(t, inds; kwargs...)
+    (BraidingStyle(sectortype(t)) isa Fermionic && !isempty(inds)) || return t
```
Jutho: I think we only use this in cases that are either `Bosonic` or `Fermionic`, but nonetheless I would find it safer to use:

```suggestion
    (BraidingStyle(sectortype(t)) isa Bosonic || isempty(inds)) && return t
```

PR #309, opened 2025-11-05
Error message improvements

lkdvos:
These are some small changes and updates to the space checks.
I think after the "treetransformer" rewrite, because we create the transformer first and only then apply the indexmanipulation, we generated some very cryptic messages from deep inside the transformer code, which I now should have resolved a little earlier in the callchain.
While I was at it, I did the same for the tensor contractions.

Mostly I followed the same path of defining appropriate methods on the spaces themselves, but I'm happy to change things or move things around if that is appropriating a bit too much of these functions.


src/spaces/homspace.jl
```diff
@@ -50,6 +50,9 @@ const TensorMapSpace{S <: ElementarySpace, N₁, N₂} = HomSpace{
     S, ProductSpace{S, N₁},
     ProductSpace{S, N₂},
 }
+numout(::Type{TensorMapSpace{S, N₁, N₂}}) where {S, N₁, N₂} = N₁
+numin(::Type{TensorMapSpace{S, N₁, N₂}}) where {S, N₁, N₂} = N₂
+numind(T::Type{TensorMapSpace{S, N₁, N₂}}) where {S, N₁, N₂} = numout(T) + numin(T)
```
Jutho: These definitions already exist on line 44 - 46

lkdvos: There is a slight difference that this is defined for types, rather than for values though

Jutho: Ah yes. Maybe these definitions can be moved together, and the value ones are no longer necessary, i.e. if we have a general `numout(x) = numout(typeof(x))` style for both tensormaps and homspaces?

src/spaces/homspace.jl
```diff
@@ -139,18 +142,33 @@ fusiontrees(W::HomSpace) = fusionblockstructure(W).fusiontreelist
...
 
+_transpose_indices(W::HomSpace) = (reverse(domainind(W)), reverse(codomainind(W)))
+
+function LinearAlgebra.transpose(W::HomSpace, (p₁, p₂)::Index2Tuple = _transpose_indices(W))
+    p = linearizepermutation(p₁, p₂, numout(W), numin(W))
+    iscyclicpermutation(p) || throw(ArgumentError(lazy"$((p₁, p₂)) is not a cyclic permutation for $W"))
+    return select(W, (p₁, p₂))
+end
+
+function braid(W::HomSpace, (p₁, p₂)::Index2Tuple, lvls::IndexTuple)
```
Jutho: `lvls` instead of `levels`? -2 characters, -100% readability 😞

src/tensors/indexmanipulations.jl
```diff
@@ -447,19 +478,16 @@ See also [`transpose`](@ref), [`transpose!`](@ref), [`add_permute!`](@ref), [`ad
         tdst::AbstractTensorMap, tsrc::AbstractTensorMap, p::Index2Tuple,
         α::Number, β::Number, backend::AbstractBackend...
     )
+    @boundscheck spacecheck_transform(transpose, tdst, tsrc, p)
     transformer = treetransposer(tdst, tsrc, p)
-    return add_transform!(tdst, tsrc, p, transformer, α, β, backend...)
+    return @inbounds add_transform!(tdst, tsrc, p, transformer, α, β, backend...)
 end
 
 function add_transform!(
```
Jutho: I think the `@inbounds` in the calls above will only have effect if you declare this with `Base.@propagate_inbounds`, as this function is probably to large to be inlined.

lkdvos:
I think I addressed all comments, this is ready for review/merging

Jutho (approved):

PR #310, opened 2025-11-05
remove`'\n'` in compact printing of tensors

lkdvos:
Additionally relaxed a `DataType` that should really have been a `Type`.

### DIFF
--- a/test/symmetries/spaces.jl
+++ b/test/symmetries/spaces.jl
@@ -454,10 +454,10 @@ end
 
 @timedtestset "show and friends" begin
     V = U1Space(i => 1 for i in 1:3)
-    @test string(V) == "Rep[U₁](1 => 1, 2 => 1, 3 => 1)"
-    @test string(V') == "Rep[U₁](1 => 1, 2 => 1, 3 => 1)'"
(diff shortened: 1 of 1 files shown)


src/spaces/gradedspace.jl
```diff
@@ -200,7 +200,7 @@ end
 Base.summary(io::IO, V::GradedSpace) = print(io, type_repr(typeof(V)))
 
 function Base.show(io::IO, V::GradedSpace)
-    opn = (get(io, :typeinfo, Any)::DataType == typeof(V) ? "" : type_repr(typeof(V)))
+    opn = (get(io, :typeinfo, Any)::Type == typeof(V) ? "" : type_repr(typeof(V)))
```
Jutho, line 203: Is this type constraint really helpful? What is its specific purpose?

lkdvos: It comes from me copying without thinking:https://github.com/JuliaLang/julia/blob/6f2f7f5dbdcb26f6ca6ae1e487b7927894323196/base/arrayshow.jl#L571-L574

I think the reason is that the `Any` that comes out of the `get` call might lead to invalidations or something similar, but I definitely don't pretend to fully understand the details.

Jutho: Ok, might be worth checking if we have other patterns like this one that do not yet have the constraint, but maybe you did that already.

PR #312, opened 2025-11-10
Updates for MatrixAlgebraKit v0.6

lkdvos:
This PR brings TensorKit up to speed with the MatrixAlgebraKit v0.6.

The main changes are the addition of the new projection functionality, and a rework of the orthnull interface.
The former is mostly just implemented here, while the latter means that a lot of code could be removed from TensorKit instead.

I additionally took the time to hopefully stabilize some of the truncated SVD AD tests, as these struggled with finite differences in combination with spaces that change.

### DIFF
--- a/src/factorizations/diagonal.jl
+++ b/src/factorizations/diagonal.jl
@@ -81,41 +81,8 @@ for f! in
         :eigh_trunc!, :right_orth!, :left_orth!,
     )
     @eval function MAK.$f!(d::DiagonalTensorMap, F, alg::DiagonalAlgorithm)
-        MAK.check_input($f!, d, F, alg)
         $f!(_repack_diagonal(d), _repack_diagonal.(F), alg)
         return F
-    end
-end
-
-for f! in (:qr_full!, :qr_compact!)
-    @eval function MAK.check_input(
-            ::typeof($f!), d::AbstractTensorMap, QR, ::DiagonalAlgorithm
-        )
-        Q, R = QR
-        @assert d isa DiagonalTensorMap
-        @assert Q isa DiagonalTensorMap && R isa DiagonalTensorMap
-        @check_scalar Q d
-        @check_scalar R d
-        @check_space(Q, space(d))
-        @check_space(R, space(d))
-
-        return nothing
-    end
-end
-
-for f! in (:lq_full!, :lq_compact!)
-    @eval function MAK.check_input(
-            ::typeof($f!), d::AbstractTensorMap, LQ, ::DiagonalAlgorithm
-        )
-        L, Q = LQ
-        @assert d isa DiagonalTensorMap
-        @assert Q isa DiagonalTensorMap && L isa DiagonalTensorMap
-        @check_scalar Q d
-        @check_scalar L d
-        @check_space(Q, space(d))
-        @check_space(L, space(d))
-
-        return nothing
     end
 end
 
@@ -126,10 +93,8 @@ end
 
 # f_vals
 # ------
-
 for f! in (:eig_vals!, :eigh_vals!, :svd_vals!)
     @eval function MAK.$f!(d::AbstractTensorMap, V, alg::DiagonalAlgorithm)
-        MAK.check_input($f!, d, V, alg)
         $f!(_repack_diagonal(d), diagview(_repack_diagonal(V)), alg)
         return V
     end
@@ -139,65 +104,4 @@ for f! in (:eig_vals!, :eigh_vals!, :svd_vals!)
         data = MAK.initialize_output($f!, _repack_diagonal(d), alg)
         return DiagonalTensorMap(data, d.domain)
     end
-end
-
-function MAK.check_input(::typeof(eig_full!), t::AbstractTensorMap, DV, ::DiagonalAlgorithm)
-    domain(t) == codomain(t) ||
-        throw(ArgumentError("Eigenvalue decomposition requires square input tensor"))
-
-    D, V = DV
-
-    @assert D isa DiagonalTensorMap
-    @assert V isa AbstractTensorMap
-
-    # scalartype checks
-    @check_scalar D t
-    @check_scalar V t
-
-    # space checks
-    @check_space D space(t)
-    @check_space V space(t)
-
-    return nothing
-end
-
-function MAK.check_input(::typeof(eigh_full!), t::AbstractTensorMap, DV, ::DiagonalAlgorithm)
-    domain(t) == codomain(t) ||
-        throw(ArgumentError("Eigenvalue decomposition requires square input tensor"))
-
-    D, V = DV
-
-    @assert D isa DiagonalTensorMap
-    @assert V isa AbstractTensorMap
-
-    # scalartype checks
-    @check_scalar D t real
-    @check_scalar V t
-
-    # space checks
-    @check_space D space(t)
-    @check_space V space(t)
-
-    return nothing
-end
-
-function MAK.check_input(::typeof(eig_vals!), t::AbstractTensorMap, D, ::DiagonalAlgorithm)
-    @assert D isa DiagonalTensorMap
-    @check_scalar D t
-    @check_space D space(t)
-    return nothing
-end
-
-function MAK.check_input(::typeof(eigh_vals!), t::AbstractTensorMap, D, ::DiagonalAlgorithm)
-    @assert D isa DiagonalTensorMap
-    @check_scalar D t real
-    @check_space D space(t)
-    return nothing
-end
-
-function MAK.check_input(::typeof(svd_vals!), t::AbstractTensorMap, D, ::DiagonalAlgorithm)
-    @assert D isa DiagonalTensorMap
-    @check_scalar D t real
-    @check_space D space(t)
-    return nothing
 end
--- a/src/factorizations/matrixalgebrakit.jl
+++ b/src/factorizations/matrixalgebrakit.jl
@@ -31,8 +31,6 @@ for f! in (
         :left_polar!, :right_polar!,
     )
     @eval function MAK.$f!(t::AbstractTensorMap, F, alg::AbstractAlgorithm)
-        MAK.check_input($f!, t, F, alg)
-
         foreachblock(t, F...) do _, (tblock, Fblocks...)
             Fblocks′ = $f!(tblock, Fblocks, alg)
             # deal with the case where the output is not in-place
@@ -41,7 +39,6 @@ for f! in (
             end
             return nothing
         end
-
         return F
     end
 end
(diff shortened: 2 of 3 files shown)


lkdvos:
I'm linking #313 and #314 here, as this should hopefully be easier to resolve with the updated orthnull interfaces

test/autodiff/ad.jl
```diff
@@ -119,9 +121,9 @@ function remove_eiggauge_dependence!(
     return ΔV
 end
 function remove_eighgauge_dependence!(
-        ΔV, D, V; degeneracy_atol = MatrixAlgebraKit.default_pullback_gaugetol(D)
+        ΔV, D, V; degeneracy_atol = MatrixAlgebraKit.default_pullback_degeneracy_atol(D)
     )
-    gaugepart = V' * ΔV
+    gaugepart = project_antihermitian!(V' * ΔV)
     gaugepart = (gaugepart - gaugepart') / 2
```
Jutho:
```suggestion
```

test/autodiff/ad.jl
```diff
@@ -594,49 +593,46 @@ for V in spacelist
...
 
-                    # TODO: I'm not sure how to properly test with spaces that might change
-                    # with the finite-difference methods, as then the jacobian is ill-defined.
-
-                    trunc = truncrank(max(2, round(Int, min(dim(domain(t)), dim(codomain(t))) * (3 / 4))))
+                    # Testing truncation with finitedifferences is RNG-prone since the
+                    # Jacobian changes size if the truncation space changes, causing errors.
+                    # So, first test the fixed space case, then do more limited testing on
+                    # some gradients and compare to the fixed space case
+                    V_trunc = spacetype(t)(c => ceil(Int, min(size(b)...) / 2) for (c, b) in blocks(t))
```
Jutho:
```suggestion
                    V_trunc = spacetype(t)(c => div(min(size(b)...), 2) for (c, b) in blocks(t))
```

src/factorizations/factorizations.jl
```diff
@@ -61,14 +55,17 @@ LinearAlgebra.svdvals!(t::AbstractTensorMap) = diagview(svd_vals!(t))
...
-    return true
+function MAK.ishermitian(t::AbstractTensorMap; kwargs...)
+    return InnerProductStyle(t) === EuclideanInnerProduct() &&
+        domain(t) == codomain(t) &&
+        all(cb -> MatrixAlgebraKit.ishermitian(cb[2]; kwargs...), blocks(t))
+end
+function MAK.isantihermitian(t::AbstractTensorMap; kwargs...)
+    return InnerProductStyle(t) === EuclideanInnerProduct() &&
+        domain(t) == codomain(t) &&
+        all(cb -> MatrixAlgebraKit.isantihermitian(cb[2]; kwargs...), blocks(t))
```
Jutho: Is it a deliberate choice to use `MAK` in the definition of the tensor function and `MatrixAlgebraKit` when calling the corresponding matrix function ?

src/factorizations/factorizations.jl
```diff
@@ -84,15 +81,19 @@ function LinearAlgebra.isposdef!(t::AbstractTensorMap)
...
     return all(f, blocks(t))
 end
-function MatrixAlgebraKit.is_right_isometry(t::AbstractTensorMap; kwargs...)
+function MAK.is_right_isometric(t::AbstractTensorMap; kwargs...)
     domain(t) ≿ codomain(t) || return false
-    f((c, b)) = MatrixAlgebraKit.is_right_isometry(b; kwargs...)
+    f((c, b)) = MAK.is_right_isometric(b; kwargs...)
     return all(f, blocks(t))
 end
+function MAK.isisometric(t::AbstractTensorMap, (p₁, p₂)::Index2Tuple; kwargs...)
```
Jutho: Do we still support a permutation argument here? Didn't we remove this in most other places?

Jutho:
What change made it such that so much code in "src/factorizations/factorizations.jl" can be removed?

kshyatt:
What's the status here? This is blocking the GPU updates PR :)

src/factorizations/factorizations.jl
```diff
@@ -61,14 +55,17 @@ LinearAlgebra.svdvals!(t::AbstractTensorMap) = diagview(svd_vals!(t))
...
+    return InnerProductStyle(t) === EuclideanInnerProduct() &&
+        domain(t) == codomain(t) &&
+        all(cb -> MAK.ishermitian(cb[2]; kwargs...), blocks(t))
+end
+function MAK.isantihermitian(t::AbstractTensorMap; kwargs...)
+    return InnerProductStyle(t) === EuclideanInnerProduct() &&
+        domain(t) == codomain(t) &&
+        all(cb -> MAK.isantihermitian(cb[2]; kwargs...), blocks(t))
 end
+LinearAlgebra.ishermitian(t::AbstractTensorMap) = MAK.ishermitian(t)
```
Jutho, line 74: I cannot ask a question about the code below since it is not changed, but still:
Is there a reason to not handle `isposdef!` via `all(..., blocks(t))`? I think that still short-circuits, and there is not really a rule against modifying the blocks?

lkdvos: I think this is just historical, that code already existed and I didn't update this when adding the other MatrixAlgebraKit routines, even though I slightly prefer the code style with `all`.

src/factorizations/adjoint.jl
```diff
@@ -60,12 +66,12 @@ for (left_f!, right_f!) in zip(
     end
 
     @eval function MAK.$left_f!(t::AdjointTensorMap, F, alg::AbstractAlgorithm)
-        $right_f!(adjoint(t), reverse(adjoint.(F)), _adjoint(alg))
-        return F
+        F′ = $right_f!(adjoint(t), reverse(adjoint.(F)), _adjoint(alg))
+        return reverse(adjoint.(F′))
```
Jutho, line 72: These were incorrect before my latest commit, so reminder here to figure out why the tests didn't pick this up.

src/factorizations/adjoint.jl
```diff
@@ -81,8 +87,8 @@ for f! in (:svd_full!, :svd_compact!, :svd_trunc!)
         return reverse(adjoint.(MAK.initialize_output($f!, adjoint(t), _adjoint(alg))))
     end
     @eval function MAK.$f!(t::AdjointTensorMap, F, alg::AbstractAlgorithm)
-        $f!(adjoint(t), reverse(adjoint.(F)), _adjoint(alg))
-        return F
+        F′ = $f!(adjoint(t), reverse(adjoint.(F)), _adjoint(alg))
+        return reverse(adjoint.(F′))
     end
 
     # disambiguate by prohibition
```
Jutho, line 98: Another out-of-place question due to the limitation to only add comments near edited code: line 113 below, namely
```julia
function MAK.svd_compact!(t::AdjointTensorMap, USVᴴ, alg::DiagonalAlgorithm)
    return MAK.svd_compact!(t, USVᴴ, alg.alg)
end
```
This doesn't seem correct. I don't think `DiagonalAlgorithm` has an `alg.alg` field. Even if this is only to fix an ambiguity and is never called, we should try to make it some valid code. But I am a bit confused about this method definition.

Jutho:
Ok, I went over this PR (and the whole `src/factorizations/` code more generally) during the past few days. I made a few changes already, but also left some more comments.

I think this is almost ready, but it would be good to have a discussion about the deprecation path of the old interface. For me, the MatrixAlgebraKit function naming was always a convenient way to implement all these methods as a library implementer. And I think it is very useful that they all exist at the level of TensorMaps as well.

But a library as TensorKit might also want to have an additional more convenient interface with shorter function names (better discoverable) and more general behavior, where truncation is simply controlled by a keyword argument, such as the earlier `tsvd` (I guess that is the most important one). It would be good to exchange ideas about this.

src/factorizations/adjoint.jl
lkdvos: I think I now corrected this, but in general the whole `Adjoint` part of this interface could definitely use some love. I feel like it might make sense to have an `AdjointAlgorithm` wrapper in MatrixAlgebraKit that could alleviate a bunch of these issues, but haven't gotten around to that just yet.

lkdvos:
I'm definitely happy to discuss having higher-level interfaces in TensorKit, but probably that does not have to hold back this PR, and should be handled separately.
I am hoping I resolved the autodiff error, this was a slight hiccup with blocks that might be fully truncated.

As mentioned in one of the comments above, in the long run I would like to have a better way to deal with the `AdjointTensorMap` implementations, preferably already in `MatrixAlgebraKit`, but that probably has to wait for a bit anyways, so I might advocate trying to merge this PR first.

src/factorizations/pullbacks.jl
```diff
@@ -33,18 +33,13 @@ for pullback! in (:svd_pullback!, :eig_pullback!, :eigh_pullback!)
             kwargs...
         )
         foreachblock(Δt, t) do c, (Δb, b)
+            haskey(inds, c) || return nothing
+            ind = inds[c]
             Fc = block.(F, Ref(c))
             ΔFc = block.(ΔF, Ref(c))
-            ind = inds[c]
             MAK.$pullback!(Δb, b, Fc, ΔFc, ind; kwargs...)
+            return nothing
```
Jutho, line 41: I am confused; was this commented out in a previous state of this PR? Was this why the forgotten `reverse` on the adjoints was not detected in the tests?

Jutho:
I think this is probably ready?

src/factorizations/pullbacks.jl
lkdvos: I think you commented this out when you changed this to use `foreachblock` instead of a simple for loop, and I then added the `haskey` clause since some blocks can be fully truncated.

lkdvos:
I agree to merge this ASAP.

I had a more in-depth look at the adjoint implementations, and it turns out that `copy_input` requires `f` instead of `f!`, so our specializations were just never called since the fallback of `copy_input` instantiates a regular tensor instead of an adjoint one.
I now fixed this, and had to then make some additional changes to ensure that the tests pass, which I hope now they do again (and this should increase coverage a bit).

One last thing I have been considering and wanted to bring up, although this can be handled in a separate PR as well since it wouldn't be breaking, is how much we value the `check_input` functions at the tensor level.
Given the current MatrixAlgebraKit implementations, most of the checks will actually be repeated again at the block level, and this is quite a bit of code, just to have the error appear one layer sooner.
It would be straightforward to simply delete all of this code, and not have a call to `check_input` in the tensor implementations.
Of course it doesn't make a big difference now since we have all of these functions written anyways, but keeping less code does seem like a viable argument.
For the sake of merging this ASAP, I've gone ahead and simply implemented this, but it is easy to revert this last commit.

For the release, I think this has to count as breaking again, both because of #305 as well as the `left_orth` and `right_orth` changes, and the `isisometry` to `isisometric`. Are there other changes we want to include in this release as well?
I can think of #291 and #261, although the latter might not be considered breaking per se.

Jutho:
I was wondering about the double action of `check_input`, but mostly from a performance concern. However, I assume these checks are mostly very lightweight, so I am wondering if removing them from the tensormap level is such a good idea. I guess it's a matter of weighing the benefits of less code to maintain vs clearer error messages.

Then again, mostly these checks are there for the output tensors, and if most users do not actually allocate these themselves, the checks are also quite useless, since they just check that our `initialize_output` works as intended. So maybe we can indeed go ahead, and add specific more informative checks at the tensor level when the need presents itself. 

So I will approve and then I think automerge will kick in.

PR #289, opened 2025-09-24
WIP: Documentation update/overhaul

Jutho:
Very much work in progress. 

@leburgel , one of the things I would like to do, is to include your "Symmetric tensor deepdive" as an extra appendix in the documentation. Would you be ok with that? Then we would have two appendices. One for the theorists, one for the practicioners.

PR #316, opened 2025-11-17
Add symmetric tensor tutorial to docs as appendix

leburgel:
Adds the symmetric tensor tutorial from the [quantumghent tensor tutorials](https://quantumghent.github.io/TensorTutorials/5-Tutorials/SymmetricTensors.html) to the docs as an appendix, as discussed in #289.

In the process, I add `subblock` and `subblocks` to the library docs, since these were missing before so I couldn't reference their docstrings.

### DIFF
--- a/docs/src/appendix/symmetric_tutorial.md
+++ b/docs/src/appendix/symmetric_tutorial.md
@@ -19,6 +19,14 @@ intuition behind them to your specific problem at hand.
     [TensorKitTensors.jl package](https://github.com/QuantumKitHub/TensorKitTensors.jl), in
     which case we basically provide a narrated walk-through of the corresponding code.
 
+#### Contents of the tutorial
+
+```@contents
+Pages = ["symmetric_tutorial.md"]
+Depth = 2:3
+```
+
+#### Setup
 
 ```@example symmetric_tutorial
 using LinearAlgebra
@@ -26,11 +34,6 @@ using TensorKit
 using WignerSymbols
 using SUNRepresentations
 using Test # for showcase testing
-```
-
-```@contents
-Pages = ["symmetric_tutorial.md"]
-Depth = 2:3
 ```
 
 ## Level 0: The transverse-field Ising model


docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1222 @@
...
+```
+To see how these fusion rules arise, we can consider the action of the symmetry
+transformation $XX$ on the possible two-site basis states, each of which corresponds to a
+tensor product of representations. We can see that $XX$ acts trivially on both $\ket{+}
+\otimes \ket{+}$ and $\ket{-} \otimes \ket{-}$ meaning these transform under the trivial
+representation, which gives the first and last entries of the fusion rules. Similarly, $XX$
+acts with a minus sign on both $\ket{+} \otimes \ket{-}$ and $\ket{-} \otimes \ket{+}$,
+meaning these transform under the sign representation, which gives the second and third
+entries of the fusion rules. Having introduced this notion of 'fusing' irreps, we can now
+associate associate a well-defined *coupled irrep* to each of the four two-site basis
```
lkdvos:
```suggestion
associate a well-defined *coupled irrep* to each of the four two-site basis
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1222 @@
...
+intuition. The creation operator $a^+$ violates particle number conservation by mapping the
+occupation number $n$ to $n + 1$. From the point of view of representation theory, this
+process can be thought of as the *fusion* of an `U1Irrep(n)` with an `U1Irrep(1)`, naturally
+giving the fusion product `U1Irrep(n + 1)`. This means we can represent $a^+$ as a
+`TensorMap(..., V ← V ⊗ A)`, where the auxiliary vector space `A` contains the $+1$ irrep
+with degeneracy 1, `A = U1Space(1 => 1)`. Similarly, the decrease in occupation number when
+acting with $a^-$ can be thought of as the *splitting* of an `U1Irrep(n)` into an
+`U1Irrep(n - 1)` and an `U1Irrep(1)`, leading to a representation in terms of a
+`TensorMap(..., A ⊗ V ← V)`. Based on these observations, we can represent the matrix
+elements {eq}`eq:bosonopmatel` as blocks labeled by the $\mathrm{U}(1)$ fusion trees
```
lkdvos: equation reference

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1222 @@
...
+refered to as the
+[*quadratic Casimir*](https://en.wikipedia.org/wiki/Representation_theory_of_SU(2)#The_Casimir_element),
+commutes with all generators. By
+[Schur's lemma](https://en.wikipedia.org/wiki/Schur%27s_lemma), it must then act
+proportionally to the identity on every irrep, where the corresponding eigenvalue is
+determined by the spin irrep label. In particular, we have for each irrep $l$
+```math
+\vec{S}^2 \ket{l,m} = l(l+1) \ket{l,m}.
+```
+It then follows from Eq. {eq}`eq:casimir_decomp` that the reduced matrix elements of the
```
lkdvos: equation reference broken

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1222 @@
...
+However, it should be noted that for general groups the Clebsch-Gordan coefficients may not
+be as easy to compute (in general, no closed formulas exist). In addition, the procedure for
+manually projecting out the reduced matrix elements requires being particularly careful
+about the correspondence between the basis states used to define the original matrix
+elements and those implied by the Clebsch-Gordan coefficients. Therefore, it is often easier
+to directly construct the symmetric tensor based on some representation theory, as we will
+see below.
+
+
+### The 'generic' approach to the spin-1 Heisenberg model: Wigner-Eckart in action
```
lkdvos, line 1025: Can we add here that this can be done automatically for the entire array with the `TensorMap` constructor as well? I think the explicit example is still relevant since it demonstrates what is going on, but we should probably get the message across that it isn't needed to do this manually (provided you have the correct basis for the dense array)

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1222 @@
...
+    l4 = f.uncoupled[2]
+    k = f.coupled
+    TT[s, f] .= (casimir(k) - casimir(l3) - casimir(l4)) / 2
+end
+subblocks(TT)
+```
+Circling back to our earlier remark, we clearly see that the fusion trees of this tensor
+indeed have non-trivial vertex labels.
+```@example symmetric_tutorial
+f = collect(fusiontrees(TT))[3][2]
```
lkdvos: This particular example seems to not have non-trivial vertex labels

lkdvos (commented):
Will try and go through in more detail, some small comments:

It might make the notation a bit lighter if we replace `ComplexF64` with `Float64` for the cases where this is applicable, I know we don't default to this in MPSKit for various reasons, but here it seems a bit much to me.

For git reasons it might be convenient to break lines at sentences instead of splitting at linewidth. It also tends to make paragraphs less jumpy for small changes/comments

docs/src/appendix/symmetric_tutorial.md
leburgel: Huh, that's strange, the fusion trees must be in a different order than when I ran the original tutorial. Is the fusion tree order deterministic given a specific homspace, and could it be that this order changed between March 2024 and now?

lkdvos: it is deterministic, and I would have to double check the exact timing but we had indeed one change (IIRC in v0.14)

docs/src/appendix/symmetric_tutorial.md
leburgel: Ah, I didn't think of this in the original tutorial, but it does seem very relevant here. The only issue is that I'm not really sure myself what the right dense array basis is for different symmetries. Is there any documentation at all on what the chosen ordering of sectors is for each sector type that's implemented? Because if this information can't be reliably found, it feels a bit strange to showcase this feature if there's no way to apply it reliably in the general case.

For SU(2) here I'm sure I can guess the irrep ordering so people could adapt that to their own SU(2)-symmetric operators, but for most other symmetries (even U(1) for example I think) I would be rather at a loss how to initialize the dense array in the first place.

leburgel: Also, I don't want to be too annoying, but if the internal sector ordering changes as any point like it did for SU(3) that would give extremely annoying breakages for people who use this approach.

So thinking about this more broadly, is calling the constructor for homspaces with non-trivial symmetries directly on a dense data array really something we want people to be doing at all?

lkdvos: There is some functionality for figuring this out:
- They are concatenated in the order determined by `Base.isless(x::I, y::I) where {I <: Sector}`
- `axes(V, sector)` will tell you the range of values, so for example `axes(SU2Space(0 => 2, 1/2 => 2), SU2Irrep(1/2)) == 3:6`.
In general, I think we are committed to not changing this, as this would indeed be a massive breaking change.

This is actually unrelated to the fusiontree ordering that happened (apparently in v0.13, I checked), which actually only shuffled the order of the internal data in the `.data` field.
I agree that these are extremely annoying breakages, which we try to avoid at all costs, but this would have been unaffected here. (and we typically have more support for upgrading old code, see e.g. https://github.com/QuantumKitHub/TensorKit.jl#transferring-tensormap-data-from-older-versions-to-v013)

I definitely understand the problem though, but I would argue that the hardest part is our (currently undocumented) basis choice of the clebsch gordan coefficients, not the order in which different sectors are concatenated, which is actually quite well-determined (although definitely also underdocumented).
Here this issue is just not showing up since we only have tensors with a single sector per leg...

leburgel: I don't know if I just never knew about the `axes` functionality for (product)spaces or I just forgot, but this is very useful. This then solves the problem of finding the right dense array basis for all Abelian groups at least. This is actually documented [here](https://github.com/QuantumKitHub/TensorKit.jl/blob/38f879810d5fc7728d1df859fc407fe32096b6fd/docs/src/man/tensors.md?plain=1#L355-L371), which is already very nice. However, the statement
> Currently, the only non-abelian sectors are `Irrep[SU₂]` and `Irrep[CU₁]`, for which the internal structure is the natural one.

is not actually that helpful, where I take some particular offense to the "the natural one" part.

I was indeed confusing two things before. I also had the basis choice for the Clebsch-Gordan coefficients in mind, in particular the ordering of the 'magnetic quantum number' within each irrep. Even here, while I think I know what the 'natural one' is, this really might depend on person to person. So I think the bottom line is that a user can only use the constructor from a dense data array for sector types for which they know exactly what the corresponding `fusiontensor` is, up to the basis choice inside each representation space.

Going back to being annoying then, even for the SU(2) example as it is in the tutorial, I already just assumed that the `fusiontensor` for SU(2) was a `WignerSymbols.clebschgordan`. I know this, but the only way to actually find out is to hunt down the [source](https://github.com/QuantumKitHub/TensorKitSectors.jl/blob/03ad9ed3f0d2375e023b5ea3f423182297aa9d0a/src/irreps/su2irrep.jl#L72-L83). So for one, I should probably switch to using `fusiontensor` here instead of going through `WignerSymbols.clebschgordan` in the example. Then I can manually state what the choice of basis within each representation space is, solving the problem for SU(2) at least. And then I can state that you can do the same just using the constructor from a dense array, provided you know exactly what `fusiontensor` actually does.

lkdvos: I agree with you, both that "natural" is not the best wording and that we should absolutely document this. Concretely, I was thinking of starting documentation for TensorKitSectors that at least states the basis for the `fusiontensor` there. For TensorKit there then needs to be slightly more explanation to show how to go from trivalent `fusiontensor` to the full fusiontree. Just obtaining the numerical tensor can actually be done with `convert(Array, (f1, f2))`, which is also not a great interface, so there I would even suggest overloading `fusiontensor(f1, f2)` instead?

PS, just wanted to mention that on top of the ordering of the magnetic quantum number there is an additional sign degree of freedom for each of these basis states, again fixed by an (undocumented) convention, so this also has to be specified. In general though, I am actually not entirely sure how to do this, i.e. it still requires a "reference point" to write down a basis?

leburgel: I tried to rewrite/add things to address this as best I could. I hope this resolves things more or less.

> PS, just wanted to mention that on top of the ordering of the magnetic quantum number there is an additional sign degree of freedom for each of these basis states

I would think that this doesn't matter in this context as long as people either use `fusiontensor` to manually project out reduced tensor elements, or just go through the `TensorMap` constructor from the dense data array. I would think this only becomes a problem when you try to manually project out Clebsch-Gordans that you defined yourself, where you run the risk of not matching the TensorKit internal sign convention.

docs/src/index.md
```diff
@@ -49,6 +49,6 @@ Depth = 2
 ## Appendix
 
 ```@contents
-Pages = ["appendix/categories.md"]
-Depth = 3
+Pages = ["appendix/categories.md", "appendix/symmetric_tutorial.md"]
```
Jutho: Maybe the "symmetric tutorial" can even be the first appendix, and the "categories" the second one?

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1289 @@
...
+using SUNRepresentations
+using Test # for showcase testing
+```
+
+
+## Level 0: The 1-dimensional Ising model
+
+As the most basic example, we will consider the
+[1-dimensional transverse-field Ising model](https://en.wikipedia.org/wiki/Transverse-field_Ising_model),
+whose Hamiltonian is given by
```
Jutho, line 43: I always doubt about whether "whose" can be used with reference to subjects that are not persons but objects, or whether instead one should use:

"the Ising model, the Hamiltonian of which is given by"

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1289 @@
...
+```math
+\begin{equation}
+\label{eq:z2gen}
+U = \prod_i X_i.
+\end{equation}
+```
+
+We will circle back to the implications of this symmetry later.
+
+As a warmup we will implement the Hamiltonian \eqref{eq:isingham} in the standard way by
```
Jutho: Unlike hyperlinks, equation references are printed in dark blue, even in dark mode. This makes them very hard to read. Even in the light theme, they clearly have a distinct blue color from hyperlinks, which is not very pretty. Not sure if that can easily be changed.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1289 @@
...
+
+```@raw html
+<center><img src="../img/symmetric_tutorial/ZZX_symm.svg" alt="ZZX_symm" class="color-invertible" style="zoom: 170%"/></center>
+```
+
+These identitities precisely mean that these local tensors transform trivially under a
+tensor product representation of $\mathbb{Z}_2$. This implies that, recalling [the
+introduction on symmetries](@ref ss_symmetries), in an appropriate basis
+for the local physical vector space, our local tensors would become block-diagonal where
+each so-called *matrix block* is labeled by a $\mathbb{Z}_2$ irrep.The appropriate local basis transformation is precisely the one that brings the local representation $X$ into block-diagonal form. Clearly, this transformation is nothing more than the Hadamard transformation which maps the computational basis of $Z$ eigenstates $\{\ket{\uparrow}, \ket{\downarrow}\}$ to that of the $X$
```
Jutho:
```suggestion
each so-called *matrix block* is labeled by a $\mathbb{Z}_2$ irrep. The appropriate local basis transformation is precisely the one that brings the local representation $X$ into block-diagonal form. Clearly, this transformation is nothing more than the Hadamard transformation which maps the computational basis of $Z$ eigenstates $\{\ket{\uparrow}, \ket{\downarrow}\}$ to that of the $X$
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1289 @@
...
+product of two irreps can be decomposed into a direct sum of irreps. This decomposition is
+encoded in what are often called the
+[*fusion rules*](https://en.wikipedia.org/wiki/Fusion_rules),
+```math
+a \otimes b \cong \bigoplus_c N_c^{ab} c,
+```
+where $N_{ab}^c$ encodes the number of times the irrep $c$ occurs in the tensor product of
+irreps $a$ and $b$. These fusion rules are called *Abelian* if the tensor product of any two
+irreps corresponds to exactly one irrep. We will return to the implications of *non-Abelian*
+fusion rules [later](@ref ss_non_abelian). For the case of the $\mathbb{Z}_2$ irreps, the fusion
```
Jutho: In the `FusionStyle` trait, we are not actually using the "abelian" vs "non-abelian" nomenclature, because at some point it became confusing if you also want to consider Vec[G]. For a non-abelian group, the "fusion" of the simple objects of Vec[G] is unique. However, it does depend on the order: `a ⊗ b != b ⊗ a`. Hence, there cannot be a braiding. On the other hand, for `Rep[G]`, it does hold that `a ⊗ b == b ⊗ a`, also for a non-abelian group.

So then, what does abelian mean: does it mean unique fusion, or independent of the order fusion? 

Note that I have no problem with calling the group abelian or non-abelian; but I would prefer to avoid calling fusion rules abelian.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1289 @@
...
+corresponds a certain reduced matrix element**. In TensorKit.jl, these reduced tensor
+elements corresponding to the fusion trees of a `TensorMap` can be accessed through the
+[`subblocks`](@ref) method.
+
+!!! note
+    In general, such a reduced matrix element is not necessarily a scalar, but rather an array
+    whose size is determined by the degeneracy of the irreps in the codomain and domain of the
+    fusion tree. For this reason, a reduced matrix element associated to a given fusion tree is
+    also referred to as an *array block*. In the following we will use terms 'reduced matrix
+    element', 'array block' or just 'block' interchangeably. However, it should be remembered
```
Jutho: Can we not consistently use subblock, or reduced matrix element, instead of simply block, and reserve the latter for the diagonal matrix blocks?

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1289 @@
...
+    that these are distinct from the matrix blocks in the block-diagonal decomposition of the
+    tensor.
+
+
+### [Fusion trees and how to use them](@id sss_fusion_trees)
+
+This view of the underlying symmetry structure in terms of fusion trees and corresponding
+array blocks is a very convenient way of working with the `TensorMap` type. Consider a
```
Jutho: Would there be a reason not to use:

```suggestion
subblocks is a very convenient way of working with the `TensorMap` type. Consider a
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1289 @@
...
+generic fusion tree of the form
+
+```@raw html
+<center><img src="../img/symmetric_tutorial/fusiontree.svg" alt="fusiontree" class="color-invertible" style="zoom: 170%"/></center>
+```
+
+which can be used to label a block of a `TensorMap` corresponding to a two-site operator.
```
Jutho:
```suggestion
which can be used to label a subblock of a `TensorMap` corresponding to a two-site operator.
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1289 @@
...
+
+- First we construct the physical space at each site as a $\mathbb{Z}_2$-graded vector space.
+- Then we initialize an empty `TensorMap` with the correct domain and codomain vector spaces built from the previously constructed physical space.
+- And finally we iterate over all splitting and fusion tree pairs and manually fill in the corresponding nonzero blocks of the operator.
+
+After the basis transform to the irrep basis, we can view the two-dimensional complex
+physical vector space we started with as being spanned by the trivial and sign irrep of
+$\mathbb{Z}_2$. In the language of TensorKit.jl, this can be implemented as a `Z2Space`, an
+alias for a
+$\mathbb{Z}_2$-[graded vector space](@ref GradedSpace)
```
Jutho: I would simply call this graded space here. Explicitly mentioning Z_2-graded might be confusing, because it is this way only because `Irrep[Z{2}]` is isomorphic to Z_2, but it is `Irrep[Z{2}]` that is providing the grading.

```suggestion
[graded vector space](@ref GradedSpace)
```

docs/src/index.md
leburgel: Fine with me!

docs/src/appendix/symmetric_tutorial.md
leburgel: It seems to be [generally accepted](https://www.merriam-webster.com/grammar/whose-used-for-inanimate-objects), and I personally don't mind anthropomorphizing here and there to make things sound less formal.

docs/src/appendix/symmetric_tutorial.md
leburgel: I also noticed this, and am very annoyed by it. I've tried to look into how I could solve this, but it seems like a very painful mathjax-css rabit hole. I'll give it another go if I have time, but not sure if I can reasonably solve this...

docs/src/appendix/symmetric_tutorial.md
leburgel: Ah I see. I tend to think of fusion rules from the perspective of anyon models. In that context, I'm pretty sure "non-Abelian" fusion rules exactly mean that there are objects with non-unique fusion, and this is standard terminology. From that point of view, I would say the fusion rules of Vec[G] are always Abelian, which is indeed a very confusing thing to say if G is non-Abelian.

The problem only problem I have with using the terminology used in the `FusionStyle` trait is that it is quite non-standard, precisely because it's made to cover general cases. So while more correct, it would also be less familiar and therefore less intuitive I think. Similarly, I think people might be more familiar with the notion of Abelian and non-Abelian anyons than they are with considering the category Vec[G]. In this entire section and in all the references linked, I'm explicitly talking about fusion rules of irreps, so I don't actually feel like this is giving any unfair implications, since for irreps it's well defined what non-Abelian fusion rules are.

I the end, I would be fine with switching to "*unique* fusion rules" and "*non-unique* fusion rules" instead of "*Abelian* fusion rules" and "*non-Abelian* fusion rules". But I would actually prefer not to because I think the latter is more familiar, even though I agree it's less correct.

docs/src/appendix/symmetric_tutorial.md
leburgel: I think I added this note simply because I noticed I was mixing everything, and this was the quickest fix. But I agree this might be needlessly confusing, I'll replace all careless "block" uses by "subblocks" or "reduced tensor elements". I like "tensor" more than "matrix" personally because you plug in an actual multidimensional array, but then still failed to use it consistently.

docs/src/appendix/symmetric_tutorial.md
leburgel: Good point!

docs/src/appendix/symmetric_tutorial.md
leburgel: I think I got rid of all lazy unspecified "block" mentions, should be more consistent now.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+using TensorKit
+using WignerSymbols
+using SUNRepresentations
+using Test # for showcase testing
+```
+
+
+## Level 0: The 1-dimensional Ising model
+
+As the most basic example, we will consider the
```
Jutho:
```suggestion
As the most basic example, we consider the
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+```math
+\begin{equation}
+\label{eq:z2gen}
+U = \prod_i X_i.
+\end{equation}
+```
+
+We will circle back to the implications of this symmetry later.
+
+As a warmup we will implement the Hamiltonian \eqref{eq:isingham} in the standard way by
```
Jutho:
```suggestion
As a warmup, we implement the Hamiltonian \eqref{eq:isingham} in the standard way by
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+encoding the matrix elements of the single-site operators $X$ and $Z$ into an array of
+complex numbers, and then combine them in a suitable way to get the Hamiltonian terms.
+Instead of using plain Julia arrays, we will use a representation in terms of `TensorMap`s
```
Jutho:
```suggestion
Instead of using plain Julia arrays, we use a representation in terms of `TensorMap`s
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+over complex vector spaces. These will essentially just be wrappers around base arrays at
```
Jutho:
```suggestion
over complex vector spaces. These are essentially just wrappers around base arrays at
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+this point, but their construction requires some consideration of the notion of *spaces*,
+which generalize the notion of `size` for arrays. Each of the operators $X$ and $Z$ acts on
+a local 2-dimensional complex vector space. In the context of TensorKit.jl such a space can
```
Jutho:
```suggestion
a local 2-dimensional complex vector space. In the context of TensorKit.jl, such a space can
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+$\ket{+} = \frac{\ket{\uparrow} + \ket{\downarrow}}{\sqrt{2}}$ and
+$\ket{-} = \frac{\ket{\uparrow} - \ket{\downarrow}}{\sqrt{2}}$. In the current context,
+this basis is referred to as the *irrep basis* of $\mathbb{Z}_2$, since each basis state
+corresponds to a one-dimensional irreducible representation of $\mathbb{Z}_2$. Indeed, the
+local symmetry transformation $X$ acts trivially on the state $\ket{+}$, corresponding to
+the *trivial irrep*, and yields a minus sign when acting on $\ket{-}$, corresponding to the
+*sign irrep*.
+
+Next, let's make the statement that 'the matrix blocks of the local tensors are labeled by
+$\mathbb{Z}_2$ irreps' more concrete. To this end, consider the action of $ZZ$ in the irrep
```
Jutho:
```suggestion
Next, let's make the statement that "the matrix blocks of the local tensors are labeled by
$\mathbb{Z}_2$ irreps" more concrete. To this end, consider the action of $ZZ$ in the irrep
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+irreps $a$ and $b$. These fusion rules are called *Abelian* if the tensor product of any two
+irreps corresponds to exactly one irrep. We will return to the implications of irreps with *non-Abelian* fusion rules [later](@ref ss_non_abelian). For the case of the $\mathbb{Z}_2$
+irreps, the fusion rules are Abelian, and are given by addition modulo 2,
+```math
+0 \otimes 0 \cong 0, \quad 0 \otimes 1 \cong 1, \quad 1 \otimes 0 \cong 1, \quad 1 \otimes 1 \cong 0.
+```
+To see how these fusion rules arise, we can consider the action of the symmetry
+transformation $XX$ on the possible two-site basis states, each of which corresponds to a
+tensor product of representations. We can see that $XX$ acts trivially on both $\ket{+}
+\otimes \ket{+}$ and $\ket{-} \otimes \ket{-}$ meaning these transform under the trivial
```
Jutho:
```suggestion
\otimes \ket{+}$ and $\ket{-} \otimes \ket{-}$, meaning these transform under the trivial
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+fields:
+
+- `uncoupled::NTuple{N,I}`: a list of `N` uncoupled charges of type `I<:Sector`
+- `coupled::I`: a single coupled charge of type `I<:Sector`
+- `isdual::NTuple{N,Bool}`: a list of booleans indicating whether the corresponding uncoupled charge is dual
+- `innerlines::NTuple{M,I}`: a list of inner lines of type `I<:Sector` of length `M = N - 2`
+- `vertices::NTuple{L,T}`: list of fusion vertex labels of type `T` and length `L = N - 1`
+
+For our current application only `uncoupled` and `coupled` are relevant, since
+$\mathbb{Z}_2$ irreps are self-dual and have Abelian fusion rules. We will come back to
```
Jutho: Not sure if it would be helpful to add something like this:

```suggestion
$\mathbb{Z}_2$ irreps are self-dual and have Abelian fusion rules, so that irreps on the inner lines of a fusion tree are completely determined by the uncoupled irreps. We will come back to
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+the *fusion tree* and encodes how the uncoupled charges $f_1$ and $f_2$ fuse to the coupled
+charge $c$. Both the splitting and fusion tree can be represented as a [`FusionTree`](@ref)
+instance. You will find such a `FusionTree` has the following properties encoded into its
+fields:
+
+- `uncoupled::NTuple{N,I}`: a list of `N` uncoupled charges of type `I<:Sector`
+- `coupled::I`: a single coupled charge of type `I<:Sector`
+- `isdual::NTuple{N,Bool}`: a list of booleans indicating whether the corresponding uncoupled charge is dual
+- `innerlines::NTuple{M,I}`: a list of inner lines of type `I<:Sector` of length `M = N - 2`
+- `vertices::NTuple{L,T}`: list of fusion vertex labels of type `T` and length `L = N - 1`
```
Jutho, line 262: Maybe a comment is necessary somewhere that we also use "charge" as an alternative to irrep . Or maybe the terminology above needs to be changed (I assume this comes from the `FusionTree` docs / comments).

Jutho: The terminology "charge" is also used below in `flip_charge`, so probably a small comment here would be sufficient but useful.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+    end
+end
+subblocks(X)
+```
+
+Given these local operators, we can use them to construct the full manifestly
+$\mathbb{Z}_2$-symmetric Hamiltonian.
+
+!!! note
+    An important observation is that when explicitly imposing the $\mathbb{Z}_2$ symmetry we
```
Jutho:
```suggestion
    An important observation is that, when explicitly imposing the $\mathbb{Z}_2$ symmetry, we
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+    $\mathbb{Z}_2$, since a single $Z$ is not invariant under conjugation with the symmetry
+    operator $X$. One might wonder whether it is still possible to construct a two-site
+    Hamiltonian term by combining local objects. This is possible if one introduces an auxiliary
+    index on the local tensors that carries a non-trivial charge. The intuition behind this will
+    become more clear in the next example.
+
+
+## Level 2: The $\mathrm{U}(1)$ Bose-Hubbard model
+
+For our next example we will consider the
```
Jutho:
```suggestion
For our next example, we consider the
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+```
+
+This Hamiltonian is invariant under conjugation by the global particle number operator, $U H
+U^\dagger = H$, where
+```math
+U = \sum_i N_i
+```
+This invariance corresponds to a $\mathrm{U}(1)$ particle number symmetry, which can again
+be manifestly imposed when constructing the Hamiltonian terms as `TensorMap`s. From the
+representation theory of $\mathrm{U}(1)$ we know that it's irreps are all one-dimensional
```
Jutho:
```suggestion
representation theory of $\mathrm{U}(1)$, we know that its irreps are all one-dimensional
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+!!! note
+    Although we have made a suggestive distinction between the 'left' and 'right' versions of
+    the operators $a_L^\pm$ and $a_R^\pm$, one can actually be obtained from the other by
+    permuting the physical and auxiliary indices of the corresponding `TensorMap`s. This
+    permutation has no effect on the actual subblocks of the tensors due to the bosonic
+    [`BraidingStyle`](@ref) of $\mathrm{U}(1)$ irreps, so the left and right operators can
+    in essence be seen as the 'same' tensors. This is no longer the case when considering
+    fermionic systems, where permuting indices can in fact change the subblocks as we will
+    see next. As a consequence, it is much less clear how to construct two-site symmetric
+    operators in terms of local symmetric objects.
```
Jutho: I like such a note box to refer to `BraidingStyle`, and so probably a similar box referring to `FusionStyle` could be useful at the beginning of the Z2 section. There, you could also elaborate on the nomenclature of "abelian fusion rules", versus what we call `UniqueFusion`.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+
+!!! note
+    From the construction of the Hamiltonian operators
+    [in terms of creation and annihilation operators](bosonham) we clearly see that they are
+    invariant under a transformation $a^\pm \to e^{\pm i\theta} a^\pm$. More generally, any
+    invertible transformation on the auxiliary space leaves the resulting contraction unchanged.
+    This ambiguity in the definition clearly shows that one should really always think in terms
+    of the fully symmetric procucts of $a^+$ and $a^-$ rather than in terms of these operators
+    themselves. In particular, one can always decompose such a symmetric product into the
+    [form above](bosonham) by means of an SVD.
```
Jutho: This note I find a bit confusing. What do you mean by "any invertible transformation on the auxiliary space leaves the resulting contraction unchanged". This requires that you transform the two operators in the correct way, namely with some `X` and then some `inv(X)`.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+
+
+## Level 3: Fermions and the Kitaev model
+
+While we have already covered quite a lot of ground towards understanding symmetric tensors
+in terms of fusion trees and corresponding subblocks, the symmetries considered so far have
+been quite 'simple' in the sense that sectors corresponding to irreps of $\mathbb{Z}_2$ and
+$\mathrm{U}(1)$ have [*Abelian fusion rules*](@ref FusionStyle) and
+[*bosonic exchange statistics*](@ref BraidingStyle).
+This means that the fusion of two irreps always gives a single irrep as the fusion product,
```
Jutho: Is this too suggestive?
```suggestion
This means that the fusion of two irreps always gives a unique irrep as the fusion product,
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+
+```@example symmetric_tutorial
+c⁺c⁻ = zeros(ComplexF64, V ⊗ V ← V ⊗ V)
+odd = FermionParity(1)
+for (s, f) in fusiontrees(c⁺c⁻)
+    if s.uncoupled[1] == odd && f.uncoupled[2] == odd && f.coupled == odd
+        c⁺c⁻[s, f] .= 1
+    end
+end
+c⁺c⁻
```
Jutho: Here and below, whatever we decide regarding the printing of `TensorMap`s (I agree with the general sentiment that the current printing output is not optimal), you probably intended to show the subblocks?
```suggestion
subblocks(c⁺c⁻)
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+```@example symmetric_tutorial
+N = zeros(ComplexF64, V ← V)
+for (s, f) in fusiontrees(N)
+    N[s, f] .= f.coupled == odd ? 1 : 0
+end
+N
+```
+
+Again, you can easily verify that all the reduced tensor elements are correct by inspecting
+the `subblocks` of each tensor.
```
Jutho: Given this sentence, I became unsure whether showing the subblocks above is what you had in mind, so I refrained from proposing the same suggestion for the other operators.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+
+!!! note
+    Working with fermionic systems is inherently tricky, as can already be seen from something
+    as simple as computing matrix elements of fermionic operators. Similarly, while constructing
+    symmetric tensors that correspond to the symmetric Hamiltonian terms was still quite
+    straightforward, it is far less clear in this case how to construct these terms as
+    contractions of local symmetric tensors representing individual creation and annihilation
+    operators in this case. While such a decomposition can always be in principle obtained using
```
Jutho: You have "in this case" twice in this sentence.
```suggestion
    operators. While such a decomposition can always be in principle obtained using
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+multilinear maps over tensor product spaces, it is natural to consider the tensor product of
+representation spaces in more detail.
+
+[From the representation theory of groups](https://en.wikipedia.org/wiki/Tensor_product_of_representations#Clebsch%E2%80%93Gordan_theory),
+it is known that the product of two irreps can in turn be decomposed into a direct sum of
+irreps, $V^{(l_1)} \otimes V^{(l_2)} \cong \bigoplus_{k} V^{(k)}$. The precise nature of
+this decomposition, also refered to as the *Clebsch-Gordan problem*, is given by the
+so-called *Clebsch-Gordan coefficients*, which we will denote as $C^{k}_{l_1,l_2}$. This set
+of coefficients, which can be interpreted as a $\text{dim}\left( V^{(l_1)} \right) \times
+\text{dim}\left( V^{(l_2)} \right) \times \text{dim}\left( V^{(k)} \right)$ array, that
```
Jutho:
```suggestion
\text{dim}\left( V^{(l_2)} \right) \times \text{dim}\left( V^{(k)} \right)$ array,
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+encodes how a basis state $\ket{k,n} \in V^{(k)}$ corresponding to some term in the direct
+sum can be decomposed into a linear combination of basis vectors $\ket{l_1,m_1} \otimes
+\ket{l_2,m_2}$ of the tensor product space:
+```math
+\begin{equation}
+\label{eq:cg_decomposition}
+\ket{k,n} = \sum_{m_1, m_2} \left( C^{k}_{l_1,l_2} \right)^{n}_{m_1, m_2} \ket{l_1,m_1} \otimes \ket{l_2,m_2}.
+\end{equation}
+```
```
Jutho, line 861: In this general discussion, it is kind of assumed that every irrep only appears once in the fusion output (`SimpleFusion`), or otherwise it is implied that `k` plays a different role than `l_1` and `l_2`, with the latter two really being irrep labels, while `k` a way to enumerate all irreps appearing in the fusion product. Maybe a `!!!note` box can be a solution?

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+As a small demonstration of this fact, we can make a simple $\mathrm{SU}(2)$-symmetric
+tensor with trivial subblock values and verify that its implied symmetry structure exactly
+corresponds to the expected Clebsch-Gordan coefficient. First, we [recall](su2_irreps) that
+the irreps of $\mathrm{SU}(2)$ can be labeled by a halfinteger *spin* that takes values $l =
+0, \frac{1}{2}, 1, \frac{3}{2}, ...$, and where the dimension of the spin-$l$ representation
+is equal to $2l + 1$. The fusion rules of $\mathrm{SU}(2)$ are given by
+```math
+\begin{equation}
+\label{eq:su2_fusion_rules}
+l_1 \otimes l_2 \cong \bigoplus_{k=|l_1-l_2|}^{l_1+l_2}s.
```
Jutho:
```suggestion
l_1 \otimes l_2 \cong \bigoplus_{k=|l_1-l_2|}^{l_1+l_2}k.
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+about the correspondence between the basis states used to define the original matrix
+elements and those implied by the Clebsch-Gordan coefficients. Finally, for some symmetries
+supported in TensorKit.jl, there are simply no Clebsch-Gordan coefficients. Therefore, it is
+often easier and sometimes simply necessary to directly construct the symmetric tensor and
+then fill in its reduced tensor elements based on some representation theory. Wel will cover
+some examples of this below.
+
+Having introduced and demonstrated the Clebsch-Gordan decomposition, the corresponding
+coefficients and their role in symmetric tensors for the example of $\mathrm{SU}(2)$ using
+the WignerSymbols.jl package, we'll continue our discussion using only TensorKit.jl
```
Jutho:
```suggestion
the WignerSymbols.jl package, we now continue our discussion using only TensorKit.jl
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+and using their orthogonality relations, we can construct a diagonal operator on each
+coupled irrep space $V^{(k)}$. Each of these diagonal operators is proportional to the
+identity, where the proportionality factor is precisely the reduced tensor element
+associated to the corresponding irrep fusion tree.
+```@raw html
+<center><img src="../img/symmetric_tutorial/none2symm.svg" alt="none2symm" class="color-invertible" style="zoom: 170%"/></center>
+```
+
+This procedure works for any group symmetry, and all we need are matrix elements of the
+operator in the irrep basis and the Clebsch-Gordan coefficients. In the following we
```
Jutho:
```suggestion
operator in the irrep basis and the Clebsch-Gordan coefficients. In the following, we
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+demonstrate this explicit procedure for the particular example of $G = \mathrm{SU}(2)$.
+However, it should be noted that for general groups the Clebsch-Gordan coefficients may not
+be as easy to compute (in general, no closed formulas exist). In addition, the procedure for
```
Jutho:
```suggestion
However, it should be noted that, for other non-Abelian groups, the Clebsch-Gordan coefficients may not
be as easy to compute (generically, no closed formulas exist). In addition, the procedure for
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+Sx = 1 / sqrt(2) * ComplexF64[0 1 0; 1 0 1; 0 1 0]
+Sy = 1 / sqrt(2) * ComplexF64[0 1im 0; -1im 0 1im; 0 -1im 0]
+Sz = ComplexF64[-1 0 0; 0 0 0; 0 0 1]
+
+@tensor SS_arr[-1 -2; -3 -4] := Sx[-1; -3] * Sx[-2; -4] + Sy[-1; -3] * Sy[-2; -4] + Sz[-1; -3] * Sz[-2; -4]
+nothing #hide
+```
+
+The next step is to project out the reduced tensor elements by taking the overlap with the
+appropriate Clebsch-Gordan coefficients. In our current case of a spin-1 physical space we
```
Jutho:
```suggestion
appropriate Clebsch-Gordan coefficients. In our current case of a spin-1 physical space, we
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+    k = f.coupled
+    SS[s, f] .= get_reduced_element(k)
+end
+subblocks(SS)
+```
+
+We demonstrated this entire procedure of extracting the reduced tensor elements of a
+symmetric tensor map for each fusion tree by projecting out the corresponding fusion tensors
+as an explicit illustration of how symmetric tensor maps work under the hood. In practice
+however, there's no need to perform this procedure explicitly. Given a dense array
```
Jutho:
```suggestion
however, there is no need to perform this procedure explicitly. Given a dense array
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+    While the example demonstrated here seems fairly straightforward, there's some inherent
+    challenges to directly initializing a symmetric tensor map from a full dense array. A first
+    important point to reiterate here is that in order for this procedure to work, we had to
+    initialize `SS_arr` by assuming an internal basis convention for the $\mathrm{SU}(2)$
+    representation space $V^{(1)}$ that is consistent with the convention used by
+    `fusiontensor`. While that choice here, corresponding to an ascending magnetic quantum
+    number $m = -1, 0, 1$, seems quite natural, for many symmetries there is no transparent
+    natural choice. In those cases, the only way to use this approach is to explicitly check the
+    basis convention used by [`TensorKitSectors.fusiontensor`](@ref) for that specific symmetry.
+    On top of thes, there's some additional complications when considering graded spaces which
```
Jutho:
```suggestion
    On top of this, there are some additional complications when considering graded spaces which
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+the action of an operator in the irrep basis can be a bit cumbersome for more complicated
+groups. However, using some basic representation theory we can bypass this step altogether
+for the Heisenberg model. First, we rewrite the exchange interaction in the following way:
+```math
+\begin{equation}
+\label{eq:casimir_decomp}
+\vec{S}_i \cdot \vec{S}_j = \frac{1}{2} \left( \left( \vec{S}_i + \vec{S}_j \right)^2 - \vec{S}_i^2 - \vec{S}_j^2 \right)
+\end{equation}
+```
+Here, $\vec{S}_i$ and $\vec{S}_j$ are spin operators on the physcial irrep, while total spin
```
Jutho:
```suggestion
Here, $\vec{S}_i$ and $\vec{S}_j$ are spin operators on the physical irrep, while total spin
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+```
+
+!!! note
+    In the previous section we have stressed the role of Clebsch-Gordan coefficients in
+    the structure of symmetric tensors, and how they can be used to map between the
+    representation of an operator in the irrep basis and its symmetric tensor representation.
+    However, for categorical symmetries such as the Fibonacci anyons, there are no
+    Clebsch-Gordan coefficients. Therefore, the 'matrix elements of the operator in the irrep
+    basis' are not well-defined, meaning that a Fibonacci-symmetric tensor cannot actually be
+    converted to a plain array in a meaningful way.
```
Jutho: Meaningful is debatable I guess, as it can be done using module categories (triple line structure). 

```suggestion
    converted to a plain array in a straightforward way.
```

Jutho (dismissed):
Great tutorial. I left some minor comments and suggestions.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+- `uncoupled::NTuple{N,I}`: a list of `N` uncoupled charges of type `I<:Sector`
+- `coupled::I`: a single coupled charge of type `I<:Sector`
+- `isdual::NTuple{N,Bool}`: a list of booleans indicating whether the corresponding uncoupled charge is dual
+- `innerlines::NTuple{M,I}`: a list of inner lines of type `I<:Sector` of length `M = N - 2`
+- `vertices::NTuple{L,T}`: list of fusion vertex labels of type `T` and length `L = N - 1`
+
+For our current application only `uncoupled` and `coupled` are relevant, since
+$\mathbb{Z}_2$ irreps are self-dual and have Abelian fusion rules, so that irreps on the inner lines of a fusion tree are completely determined by the uncoupled irreps. We will come back to
+these other properties when discussion more involved applications. Given some `TensorMap`,
+the method `TensorKit.fusiontrees(t::TensorMap)` returns an iterator over all pairs of
```
borisdevos: This method is exported, so no need to specify the module.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+the correct domain and codomain vector spaces, and fill in the nonzero subblocks associated
+to [the fusion trees shown above](U1_fusiontrees). To do this we need access to the integer
+label of the $\mathrm{U}(1)$ irreps in the fusion and splitting trees, which can be accessed
+through the `charge` field of the `U1Irrep` type.
+
+```@example symmetric_tutorial
+a⁺a⁻ = zeros(ComplexF64, V ⊗ V ← V ⊗ V)
+for (s, f) in fusiontrees(a⁺a⁻)
+    if s.uncoupled[1] == only(f.uncoupled[1] ⊗ U1Irrep(1)) && s.uncoupled[2] == only(f.uncoupled[2] ⊗ U1Irrep(-1))
+        a⁺a⁻[s, f] .= sqrt(s.uncoupled[1].charge * f.uncoupled[2].charge)
```
borisdevos, line 485: Maybe use the `charge` function here, and all other instances below?

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
+# [A symmetric tensor deep dive: constructing your first tensor map](@id s_symmetric_tutorial)
+
+In this tutorial, we will demonstrate how to construct specific [`TensorMap`](@ref)s which
+are relevant to some common physical systems, with an increasing degree of complexity. We
+will assume the reader is somewhat familiar with [the notion of a 'tensor map'](@ref
+ss_whatistensor) and has a rough idea of [what it means for a tensor map to be
+'symmetric'](@ref ss_symmetries). In going through these examples we aim to provide a
+relatively gently introduction to the meaning of [symmetry sectors](@ref ss_sectors) and
```
borisdevos:
```suggestion
relatively gentle introduction to the meaning of [symmetry sectors](@ref ss_sectors) and
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+[vector spaces](@ref ss_rep) within the context of TensorKit.jl, [how to initialize a
+`TensorMap` over a given vector space](@ref ss_tensor_construction) and finally how to
+manually set the data of a [symmetric `TensorMap`](@ref ss_tutorial_symmetries). We will
+keep our discussion as intuitive and simple as possible, only adding as many technical
+details as strictly necessary to understand each example. When considering a different
+physical system of interest, you should then be able to adatpt these recipes and the
```
borisdevos:
```suggestion
physical system of interest, you should then be able to adapt these recipes and the
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+Q = \exp \left( i \pi \sum_i N_i \right) = (-1)^{\sum_i N_i}.
+```
+This fermion parity symmetry, which we will denote as $f\mathbb{Z}_2$, is a
+$\mathbb{Z}_2$-like symmetry in the sense that it has a trivial representation, which we
+call *even* and again denote by '0', and a sign representation which we call *odd* and
+denote by '1'. The fusion rules of these irreps are the same as for $\mathbb{Z}_2$. Similar
+to the previous case, the local symmetry operator $Q_i$ is already diagonal, so the
+occupation number basis coincides with the irrep basis and we don't need an additional basis
+transform. The important difference with a regular $\mathbb{Z}_2$ symmetry is that the
+irreps of $f\mathbb{Z}_2$ have fermionic braiding statistics, in the sense that exhanging
```
borisdevos:
```suggestion
irreps of $f\mathbb{Z}_2$ have fermionic braiding statistics, in the sense that exchanging
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+    fermionic symmetries, but it is not something we will do here.
+
+
+## [Level 4: Non-Abelian symmetries and the quantum Heisenberg model](@id ss_non_abelian)
+
+We will now move on to systems which have more complicated *non-Abelian* symmetries. For a
+non-Abelian symmetry group $G$, the fact that its elements do not all commute has a profound
+impact on its representation theory. In particular, the irreps of such a group can be higher
+dimensional, and the fusion of two irreps can give rise to multiple different irreps. On the
+one hand this means that fusion trees of these irreps are no longer completely determined by
```
borisdevos:
```suggestion
one hand, this means that fusion trees of these irreps are no longer completely determined by
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+```@example symmetric_tutorial
+ta = convert(Array, t)
+```
+The conversion gives us a $2 \times 2 \times 3$ array, which exactly corresponds to the size
+of the $C^{1}_{\frac{1}{2},\frac{1}{2}}$ Clebsch-Gordan array. In order to explicitly
+compare whether the entries match we need to know the ordering of basis states assumed by
+TensorKit.jl when converting the tensor to its matrix elements in the irrep basis. For
+$\mathrm{SU}(2)$ the irrep basis is ordered in ascending magnetic quantum number $m$, which
+gives us a map $m = i - (l+1)$ for mapping an array index to a corresponding magnetic
+quantum number for the spin-$l$ irrep.
```
borisdevos, line 946: May I suggest having this part stick out more in a box or so? In the most recent ETN school there was a session about "unifying" tensor network packages, and a point that got brought up is a lack of describing (and consistency in) the conventions of symmetric tensors. It might be good to have this pop out? Though in general this point should be brought up in more visible parts of the docs.

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1291 @@
...
+operator in the irrep basis and the Clebsch-Gordan coefficients. In the following, we
+demonstrate this explicit procedure for the particular example of $G = \mathrm{SU}(2)$.
+However, it should be noted that, for other non-Abelian groups, the Clebsch-Gordan coefficients may not
+be as easy to compute (generically, no closed formulas exist). In addition, the procedure for
+manually projecting out the reduced tensor elements requires being particularly careful
+about the correspondence between the basis states used to define the original matrix
+elements and those implied by the Clebsch-Gordan coefficients. Finally, for some symmetries
+supported in TensorKit.jl, there are simply no Clebsch-Gordan coefficients. Therefore, it is
+often easier and sometimes simply necessary to directly construct the symmetric tensor and
+then fill in its reduced tensor elements based on some representation theory. Wel will cover
```
borisdevos:
```suggestion
then fill in its reduced tensor elements based on some representation theory. We will cover
```

docs/src/appendix/symmetric_tutorial.md
leburgel: I tried to figure this out, but with no success. It should be possible to configure this in Mathjax somehow (see e.g. [here](https://stackoverflow.com/questions/65020117/mathjax-2-7-7-eqref-color)). However, I find it very unclear how to actually configure the Mathjax enginge for Documenter.jl in detail, and some trial and error attempts lead me nowhere.

docs/src/appendix/symmetric_tutorial.md
leburgel: As long as `TensorKitSectors.charge` is not reexported by TensorKit.jl I would prefer not to do this. Accessing the struct field directly is definitely bad practice black magic, but at least the field name can be found in a public docstring and is included in the TensorKit.jl documentation. So for now, I actually prefer the `.charge` approach here.

docs/src/appendix/symmetric_tutorial.md
leburgel: I rewrote the first paragraph of the fusion trees section to explicitly introduce the names "charge" and "sector" as generalizations of group irreps.

docs/src/appendix/symmetric_tutorial.md
leburgel: I added a `FusionStyle` box, mentioning `UniqueFusion` and `MultipleFusion` and how they relate to the nomenclature commonly used in anyon models that I adopted here.

docs/src/appendix/symmetric_tutorial.md
leburgel: Good point, that wasn't clear at all. I rewrote it to clearly describe what kind of transformation I mean.

docs/src/appendix/symmetric_tutorial.md
leburgel: There's an explicit warning box on (something very similar to) this [here](https://github.com/QuantumKitHub/TensorKit.jl/blob/2c54329ced147fa995a4851510059cfa5a12f60d/docs/src/appendix/symmetric_tutorial.md?plain=1#L1072-L1087). Is this box kind of what you meant, or are there specific things missing that you think would be good to add?

docs/src/appendix/symmetric_tutorial.md
leburgel: Good point. I do say something about fusion multiplicities further down when we explicitly call `fusiontensor`, but it's indeed appropriate to already mention this here and then refer to this later. I'll add a note box.

docs/src/appendix/symmetric_tutorial.md
borisdevos: I did think your box covered part of what I asked, and so it covers that information within the tutorial, but I commented on these lines above simply because they just literally give the information, and I at least like to use packages whose doc(string)s give me information literally and noticeably. But maybe it's something that should be added at the level of the docstrings of these spaces where conventions matter.

docs/src/index.md
lkdvos:
```suggestion
Pages = ["appendix/symmetric_tutorial.md", "appendix/categories.md"]
```
This still has to be updated

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1310 @@
...
+
+# construct physical Hilbert space
+V = ℂ^2
+
+# construct the physical operators as TensorMaps
+X = TensorMap(x_mat, V ← V)
+Z = TensorMap(z_mat, V ← V)
+
+# combine single-site operators into two-site operator
+ZZ = Z ⊗ Z;
```
lkdvos:
```suggestion
ZZ = Z ⊗ Z

nothing # hide
```

I think otherwise this still prints out the result

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1310 @@
...
+```@example symmetric_tutorial
+using LinearAlgebra
+using TensorKit
+using WignerSymbols
+using SUNRepresentations
+using Test # for showcase testing
+```
+
+
+## Level 0: The 1-dimensional Ising model
```
lkdvos: Do we have to add "1-dimensional" here? Since we are always just looking at the building blocks, I don't think that actually makes a difference, so we might as well write "transverse-field Ising model" instead

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1310 @@
...
+    refer to as "Abelian" fusion rules in this tutorial corresponds to
+    `UniqueFusion <: FusionStyle`. We will also consider [examples](@ref ss_non_abelian) of
+    two different kinds of non-Abelian" fusion rules, corresponding to
+    `MultipleFusion <: FusionStyle` styles.
+
+For the case of the $\mathbb{Z}_2$ irreps, the fusion rules are Abelian, and are given by
+addition modulo 2,
+```math
+0 \otimes 0 \cong 0, \quad 0 \otimes 1 \cong 1, \quad 1 \otimes 0 \cong 1, \quad 1 \otimes 1 \cong 0.
+```
```
lkdvos, line 181:
```suggestion
```@example
for a in values(Z2Irrep), b in values(Z2Irrep)
    println(a, " ⊗ ", b, ": ", a ⊗ b)
end
```
```
We could consider adding these programatically as well

File: docs/src/appendix/symmetric_tutorial.md (hunk 27 above)
Comment (lkdvos) [reply docs/src/appendix/symmetric_tutorial.md]: I think this is actually also the case when the `FusionStyle` is not `UniqueFusion`, even for `Bosonic` braidings (e.g. the `Rsymbol` of SU2 can be -1).

File: docs/src/appendix/symmetric_tutorial.md (hunk 50 above)
Comment (leburgel) [reply docs/src/appendix/symmetric_tutorial.md:946]: I see, I think I get what you mean. I think I explained things as best I could in the context of the tutorial itself. See also the discussion we had under  [the review comment above](https://github.com/QuantumKitHub/TensorKit.jl/pull/316#discussion_r2535119880).

But I totally agree these convention choices should just be visible somewhere, for every sector type that is exported at least. The most natural place to do this is in TensorKitSectors.jl I think, either in actual docstrings (of `fusiontensor` for every sector type?), or in the documentation page in a dedicated section.

File: docs/src/appendix/symmetric_tutorial.md (hunk 53 above)
Comment (leburgel) [reply docs/src/appendix/symmetric_tutorial.md]: Good point, changing it to "transverse-field Ising model".

File: docs/src/appendix/symmetric_tutorial.md (hunk 52 above)
Comment (leburgel) [reply docs/src/appendix/symmetric_tutorial.md]: I actually don't mind it printing here (with the new short summary), but forgot to remove the semicolon.

File: docs/src/appendix/symmetric_tutorial.md (hunk 54 above)
Comment (leburgel) [reply docs/src/appendix/symmetric_tutorial.md:181]: I like the idea, but the location isn't really ideal here since we haven't actually touched any code for sectors at this point. I'll add it a bit further down, once we've introduced the `Z2Irrep` sector type.

File: docs/src/appendix/symmetric_tutorial.md (hunk 27 above)
Comment (leburgel) [reply docs/src/appendix/symmetric_tutorial.md]: I see, `Bosonic` just means the twists are trivial, right. I think properly discussing twists is a bit beyond the scope of the tutorial, it would take up quite a bit of space (and time for me to properly structure it), and would probably still be confusing and distract a bit from the main message.

I'll update the note to be technically correct, but I prefer not to give any more twist details for now.

File: docs/src/appendix/symmetric_tutorial.md [hunk 55]
```diff
@@ -0,0 +1,1334 @@
...
+
+!!! note
+    Although we have made a suggestive distinction between the 'left' and 'right' versions of
+    the operators $a_L^\pm$ and $a_R^\pm$, one can actually be obtained from the other by
+    permuting the physical and auxiliary indices of the corresponding `TensorMap`s. This
+    permutation has no effect on the actual subblocks of the tensors due to the Abelian
+    [`FusionStyle`](@ref) and bosonic [`BraidingStyle`](@ref) of $\mathrm{U}(1)$ irreps, so
+    the left and right operators can  in essence be seen as the 'same' tensors. This is no
+    longer the case when considering non-Abelian or symmetries. For these cases, permuting
```
Jutho:
```suggestion
    longer the case when considering non-Abelian symmetries, or symmetries associated with fermions or anyons. For these cases, permuting
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1334 @@
...
+end
+```
+After the basis transform to the irrep basis, we can view the two-dimensional complex
+physical vector space we started with as being spanned by the trivial and sign irrep of
+$\mathbb{Z}_2$. In the language of TensorKit.jl, this can be implemented as a `Z2Space`, an
+alias for a [graded vector space](@ref GradedSpace) `Vect[Z2Irrep]`. To construct such a
+graded space we have to specify which irreps it contains, and indicate the degenaracy of
+each irrep. Here, our physical vector space contains the trivial irrep `Z2Irrep(0)` with
+degeneracy 1 and the sign irrep `Z2Irrep(1)` with degeneracy 1. This means can define this
+space in the following way, and check its dimension:
```
Jutho: Will degeneracy be a clear concept in this context for new readers? Maybe something like:

```suggestion
alias for a [graded vector space](@ref GradedSpace) `Vect[Z2Irrep]`. To construct such a
graded space $$V$$, we have to specify the number of times each irrep appears in the direct sum decomposition, 
a concept known as the degeneracy of irrep `a` in `V` and sometimes denoted as $$N^a_V$$.
Only irreps with non-zero degeneracy need to be specified. Here, our physical vector space contains
the trivial irrep `Z2Irrep(0)` with degeneracy 1 and the sign irrep `Z2Irrep(1)` with degeneracy 1.
This means can define this space in the following way, and check its dimension:
```

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1334 @@
...
+
+```@example symmetric_tutorial
+V = Z2Space(0 => 1, 1 => 1)
+dim(V)
```
Jutho: Maybe add the method to obtain the degeneracy here as well
```suggestion
dim(V)
dim(V, Z2Irrep(0)) == 1
dim(V, Z2Irrep(1))
```

docs/src/appendix/symmetric_tutorial.md
leburgel: Introducing 'degeneracy' a bit more gently is definitely a good idea, but I don't want to introduce notation that I then don't use again, so I'll add a slightly rephrased version.

leburgel: Just explaining in words without introducing new notation looked even more confusing, so I went more in the direction of your suggestion and beyond after all :)

docs/src/appendix/symmetric_tutorial.md
```diff
@@ -0,0 +1,1359 @@
...
+using TensorKit
+using WignerSymbols
+using SUNRepresentations
+using Test # for showcase testing
+```
+
+```@contents
+Pages = ["symmetric_tutorial.md"]
+Depth = 2:3
+```
```
Jutho: Final mosquito sifting: The note box, the package loading code box and the TOC appearing directly after each other without any header or transition text looks a bit weird.

leburgel: I added headers for the toc and the setup, and switched their order. This should look better I think.

Jutho: Excellent

PR #320, opened 2025-12-01
Start on GPU extensions (again)

kshyatt:
https://github.com/QuantumKitHub/TensorKit.jl/pull/299 will be closed, was easier to just create a new branch and `cherry-pick`.

### DIFF
--- a/ext/TensorKitAMDGPUExt/TensorKitAMDGPUExt.jl
+++ b/ext/TensorKitAMDGPUExt/TensorKitAMDGPUExt.jl
@@ -1,0 +1,108 @@
+module TensorKitAMDGPUExt
+
+using AMDGPU, AMDGPU.rocBLAS, LinearAlgebra
+using AMDGPU: @allowscalar
+import AMDGPU: rand as rocrand, rand! as rocrand!, randn as rocrandn, randn! as rocrandn!
+
+using TensorKit
+using TensorKit.Factorizations
+using TensorKit.Strided
+using TensorKit.Factorizations: AbstractAlgorithm
+using TensorKit: SectorDict, tensormaptype, scalar, similarstoragetype, AdjointTensorMap, scalartype
+
+using TensorKit.MatrixAlgebraKit
+
+using Random
+
+include("roctensormap.jl")
+
+const ROCDiagonalTensorMap{T, S} = DiagonalTensorMap{T, S, ROCVector{T, AMDGPU.Mem.HIPBuffer}}
+
+"""
+    ROCDiagonalTensorMap{T}(undef, domain::S) where {T,S<:IndexSpace}
+    # expert mode: select storage type `A`
+    DiagonalTensorMap{T,S,A}(undef, domain::S) where {T,S<:IndexSpace,A<:DenseVector{T}}
+
+Construct a `DiagonalTensorMap` with uninitialized data.
+"""
+function ROCDiagonalTensorMap{T}(::UndefInitializer, V::TensorMapSpace) where {T}
+    (numin(V) == numout(V) == 1 && domain(V) == codomain(V)) ||
+        throw(ArgumentError("DiagonalTensorMap requires a space with equal domain and codomain and 2 indices"))
+    return ROCDiagonalTensorMap{T}(undef, domain(V))
+end
+function ROCDiagonalTensorMap{T}(::UndefInitializer, V::ProductSpace) where {T}
+    length(V) == 1 ||
+        throw(ArgumentError("DiagonalTensorMap requires `numin(d) == numout(d) == 1`"))
+    return ROCDiagonalTensorMap{T}(undef, only(V))
+end
+function ROCDiagonalTensorMap{T}(::UndefInitializer, V::S) where {T, S <: IndexSpace}
+    return ROCDiagonalTensorMap{T, S}(undef, V)
+end
+ROCDiagonalTensorMap(::UndefInitializer, V::IndexSpace) = ROCDiagonalTensorMap{Float64}(undef, V)
+
+function ROCDiagonalTensorMap(data::ROCVector{T}, V::S) where {T, S}
+    return ROCDiagonalTensorMap{T, S}(data, V)
+end
+
+function ROCDiagonalTensorMap(data::Vector{T}, V::S) where {T, S}
+    return ROCDiagonalTensorMap{T, S}(ROCVector{T}(data), V)
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(svd_full!), t::ROCDiagonalTensorMap, alg::DiagonalAlgorithm)
+    V_cod = fuse(codomain(t))
+    V_dom = fuse(domain(t))
+    U = similar(t, codomain(t) ← V_cod)
+    S = ROCDiagonalTensorMap{real(scalartype(t))}(undef, V_cod ← V_dom)
+    Vᴴ = similar(t, V_dom ← domain(t))
+    return U, S, Vᴴ
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(svd_vals!), t::ROCTensorMap, alg::AbstractAlgorithm)
+    V_cod = infimum(fuse(codomain(t)), fuse(domain(t)))
+    return ROCDiagonalTensorMap{real(scalartype(t))}(undef, V_cod)
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(svd_compact!), t::ROCTensorMap, ::AbstractAlgorithm)
+    V_cod = V_dom = infimum(fuse(codomain(t)), fuse(domain(t)))
+    U = similar(t, codomain(t) ← V_cod)
+    S = ROCDiagonalTensorMap{real(scalartype(t))}(undef, V_cod)
+    Vᴴ = similar(t, V_dom ← domain(t))
+    return U, S, Vᴴ
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(eigh_full!), t::ROCTensorMap, ::AbstractAlgorithm)
+    V_D = fuse(domain(t))
+    T = real(scalartype(t))
+    D = ROCDiagonalTensorMap{T}(undef, V_D)
+    V = similar(t, codomain(t) ← V_D)
+    return D, V
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(eig_full!), t::ROCTensorMap, ::AbstractAlgorithm)
+    V_D = fuse(domain(t))
+    Tc = complex(scalartype(t))
+    D = ROCDiagonalTensorMap{Tc}(undef, V_D)
+    V = similar(t, Tc, codomain(t) ← V_D)
+    return D, V
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(eigh_vals!), t::ROCTensorMap, alg::AbstractAlgorithm)
+    V_D = fuse(domain(t))
+    T = real(scalartype(t))
+    return D = ROCDiagonalTensorMap{Tc}(undef, V_D)
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(eig_vals!), t::ROCTensorMap, alg::AbstractAlgorithm)
+    V_D = fuse(domain(t))
+    Tc = complex(scalartype(t))
+    return D = ROCDiagonalTensorMap{Tc}(undef, V_D)
+end
+
+
+# TODO
+# add VectorInterface extensions for proper AMDGPU promotion
+function TensorKit.VectorInterface.promote_add(TA::Type{<:AMDGPU.StridedROCMatrix{Tx}}, TB::Type{<:AMDGPU.StridedROCMatrix{Ty}}, α::Tα = TensorKit.VectorInterface.One(), β::Tβ = TensorKit.VectorInterface.One()) where {Tx, Ty, Tα, Tβ}
+    return Base.promote_op(add, Tx, Ty, Tα, Tβ)
+end
+
+end
--- a/ext/TensorKitAMDGPUExt/roctensormap.jl
+++ b/ext/TensorKitAMDGPUExt/roctensormap.jl
@@ -1,0 +1,257 @@
+const ROCTensorMap{T, S, N₁, N₂} = TensorMap{T, S, N₁, N₂, ROCVector{T, AMDGPU.Mem.HIPBuffer}}
+const ROCTensor{T, S, N} = ROCTensorMap{T, S, N, 0}
+
+const AdjointROCTensorMap{T, S, N₁, N₂} = AdjointTensorMap{T, S, N₁, N₂, ROCTensorMap{T, S, N₁, N₂}}
+
+function TensorKit.tensormaptype(S::Type{<:IndexSpace}, N₁, N₂, TorA::Type{<:StridedROCArray})
+    if TorA <: ROCArray
+        return TensorMap{eltype(TorA), S, N₁, N₂, ROCVector{eltype(TorA), AMDGPU.Mem.HIPBuffer}}
+    else
+        throw(ArgumentError("argument $TorA should specify a scalar type (`<:Number`) or a storage type `<:ROCVector{<:Number}`"))
+    end
+end
+
+function ROCTensorMap{T}(::UndefInitializer, V::TensorMapSpace{S, N₁, N₂}) where {T, S, N₁, N₂}
+    return ROCTensorMap{T, S, N₁, N₂}(undef, V)
+end
+
+function ROCTensorMap{T}(
+        ::UndefInitializer, codomain::TensorSpace{S},
+        domain::TensorSpace{S}
+    ) where {T, S}
+    return ROCTensorMap{T}(undef, codomain ← domain)
+end
+function ROCTensor{T}(::UndefInitializer, V::TensorSpace{S}) where {T, S}
+    return ROCTensorMap{T}(undef, V ← one(V))
+end
+# constructor starting from block data
+"""
+    ROCTensorMap(data::AbstractDict{<:Sector,<:ROCMatrix}, codomain::ProductSpace{S,N₁},
+                domain::ProductSpace{S,N₂}) where {S<:ElementarySpace,N₁,N₂}
+    ROCTensorMap(data, codomain ← domain)
+    ROCTensorMap(data, domain → codomain)
+
+Construct a `ROCTensorMap` by explicitly specifying its block data.
+
+## Arguments
+- `data::AbstractDict{<:Sector,<:ROCMatrix}`: dictionary containing the block data for
+  each coupled sector `c` as a matrix of size `(blockdim(codomain, c), blockdim(domain, c))`.
+- `codomain::ProductSpace{S,N₁}`: the codomain as a `ProductSpace` of `N₁` spaces of type
+  `S<:ElementarySpace`.
+- `domain::ProductSpace{S,N₂}`: the domain as a `ProductSpace` of `N₂` spaces of type
+  `S<:ElementarySpace`.
+
+Alternatively, the domain and codomain can be specified by passing a [`HomSpace`](@ref)
+using the syntax `codomain ← domain` or `domain → codomain`.
+"""
+function ROCTensorMap(
+        data::AbstractDict{<:Sector, <:ROCArray},
+        V::TensorMapSpace{S, N₁, N₂}
+    ) where {S, N₁, N₂}
+    T = eltype(valtype(data))
+    t = ROCTensorMap{T}(undef, V)
+    for (c, b) in blocks(t)
+        haskey(data, c) || throw(SectorMismatch("no data for block sector $c"))
+        datac = data[c]
+        size(datac) == size(b) ||
+            throw(DimensionMismatch("wrong size of block for sector $c"))
+        copy!(b, datac)
+    end
+    for (c, b) in data
+        c ∈ blocksectors(t) || isempty(b) ||
+            throw(SectorMismatch("data for block sector $c not expected"))
+    end
+    return t
+end
+function ROCTensorMap{T}(
+        data::DenseVector{T}, codomain::TensorSpace{S},
+        domain::TensorSpace{S}
+    ) where {T, S}
+    return ROCTensorMap(data, codomain ← domain)
+end
+function ROCTensorMap(
+        data::AbstractDict{<:Sector, <:ROCMatrix}, codom::TensorSpace{S},
+        dom::TensorSpace{S}
+    ) where {S}
+    return ROCTensorMap(data, codom ← dom)
+end
+
+function ROCTensorMap(ts::TensorMap{T, S, N₁, N₂, A}) where {T, S, N₁, N₂, A}
+    return ROCTensorMap{T, S, N₁, N₂}(ROCArray(ts.data), ts.space)
+end
+
+for (fname, felt) in ((:zeros, :zero), (:ones, :one))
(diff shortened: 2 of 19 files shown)


ext/TensorKitAMDGPUExt/TensorKitAMDGPUExt.jl
```diff
@@ -0,0 +1,109 @@
+module TensorKitAMDGPUExt
+
+using AMDGPU, AMDGPU.rocBLAS, LinearAlgebra
+using AMDGPU: @allowscalar
+import AMDGPU: rand as rocrand, rand! as rocrand!, randn as rocrandn, randn! as rocrandn!
+
+using TensorKit
+import TensorKit.VectorInterface: scalartype as vi_scalartype
```
Jutho: Is there another `scalartype` that might conflict with the one from `VectorInterface`?

ext/TensorKitAMDGPUExt/TensorKitAMDGPUExt.jl
```diff
@@ -0,0 +1,109 @@
...
+function ROCDiagonalTensorMap{T}(::UndefInitializer, V::TensorMapSpace) where {T}
+    (numin(V) == numout(V) == 1 && domain(V) == codomain(V)) ||
+        throw(ArgumentError("DiagonalTensorMap requires a space with equal domain and codomain and 2 indices"))
+    return ROCDiagonalTensorMap{T}(undef, domain(V))
+end
+function ROCDiagonalTensorMap{T}(::UndefInitializer, V::ProductSpace) where {T}
+    length(V) == 1 ||
+        throw(ArgumentError("DiagonalTensorMap requires `numin(d) == numout(d) == 1`"))
+    return ROCDiagonalTensorMap{T}(undef, only(V))
+end
```
Jutho, line 37: We don't have these generalizations of the constructor for the normal (i.e. non-gpu) `DiagonalTensorMap` right? Is this useful?

src/tensors/linalg.jl
```diff
@@ -317,8 +317,8 @@ function LinearAlgebra.cond(t::AbstractTensorMap, p::Real = 2)
             return zero(real(float(scalartype(t))))
         end
         S = LinearAlgebra.svdvals(t)
-        maxS = maximum(first, values(S))
-        minS = minimum(last, values(S))
+        maxS = mapreduce(maximum, max, values(S))
+        minS = mapreduce(minimum, min, values(S))
```
Jutho, line 321: Do `first` and `last` not work for the gpu arrays?

src/tensors/tensor.jl
```diff
@@ -139,7 +139,7 @@ function TensorMap(
         datac = data[c]
         size(datac) == size(b) ||
             throw(DimensionMismatch("wrong size of block for sector $c"))
-        copy!(b, datac)
+        copyto!(b, datac)
```
Jutho: Is `copyto!` better supported on the gpu? We tend to like `copy!` because it is more strict in checking size compatibility between its two arguments (but should otherwise behave the same). If `copy!` doesn't work well on the gpu arrays, that really seems like a (CUDA/AMDGPU).jl problem.

src/tensors/tensor.jl
```diff
@@ -357,7 +357,7 @@ function TensorMap(
     ) where {S}
     return TensorMap(data, codom ← dom; kwargs...)
 end
-function Tensor(data::AbstractArray, codom::TensorSpace, ; kwargs...)
+function Tensor(data::AbstractArray, codom::TensorSpace; kwargs...)
```
Jutho, line 360: Nice catch 😄

ext/TensorKitAMDGPUExt/TensorKitAMDGPUExt.jl
```diff
@@ -0,0 +1,109 @@
...
+function ROCDiagonalTensorMap(data::Vector{T}, V::S) where {T, S}
+    return ROCDiagonalTensorMap{T, S}(ROCVector{T}(data), V)
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(svd_full!), t::ROCDiagonalTensorMap, alg::DiagonalAlgorithm)
+    V_cod = fuse(codomain(t))
+    V_dom = fuse(domain(t))
+    U = similar(t, codomain(t) ← V_cod)
+    S = ROCDiagonalTensorMap{real(scalartype(t))}(undef, V_cod ← V_dom)
+    Vᴴ = similar(t, V_dom ← domain(t))
```
Jutho, line 56: This doesn't sound right. For `svd_full!`, we aren't using a `DiagonalTensorMap` for `S`, exactly because we make `U` and `Vᴴ` "square" and `S` rectangular.

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,278 @@
...
+        end
+    end
+end
+
+# converters
+# ----------
+function Base.convert(::Type{ROCTensorMap}, d::Dict{Symbol, Any})
+    try
+        codomain = eval(Meta.parse(d[:codomain]))
+        domain = eval(Meta.parse(d[:domain]))
```
lkdvos:
```suggestion
        codomain = eval(TensorKit, Meta.parse(d[:codomain]))
        domain = eval(TensorKit, Meta.parse(d[:domain]))
```

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,278 @@
...
+        $randfun(codomain::TensorSpace) = $randfun(codomain ← one(codomain))
+        function $randfun(::Type{T}, codomain::TensorSpace) where {T}
+            return $randfun(T, codomain ← one(codomain))
+        end
+        function $randfun(
+                rng::Random.AbstractRNG, ::Type{T},
+                codomain::TensorSpace
+            ) where {T}
+            return $randfun(rng, T, codomain ← one(domain))
+        end
```
Jutho: Can these be obtained by adding `domain::TensorSpace{S} = one(codomain)` in the previous methods, analogously to how you did it for `zeros` and `ones` above?

Definite also line 139 is wrong, cause it says `one(domain)` whereas `domain` is not a valid name in this scope.

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,278 @@
...
+end
+
+# Scalar implementation
+#-----------------------
+function TensorKit.scalar(t::ROCTensorMap)
+
+    # TODO: should scalar only work if N₁ == N₂ == 0?
+    return @allowscalar dim(codomain(t)) == dim(domain(t)) == 1 ?
+        first(blocks(t))[2][1, 1] : throw(DimensionMismatch())
+end
```
Jutho, line 174: This function is being somewhat rewritten in #291 so it might need to be revised if that PR is merged.

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,278 @@
...
+
+TensorKit.scalartype(A::StridedROCArray{T}) where {T} = T
+vi_scalartype(::Type{<:ROCTensorMap{T}}) where {T} = T
+vi_scalartype(::Type{<:ROCArray{T}}) where {T} = T
```
Jutho: I have checked and we have `TensorKit.scalartype === VectorInterface.scalartype`, so there is no need to distinguish these two.

src/tensors/linalg.jl
kshyatt: Unfortunately, no, as they require scalar indexing

ext/TensorKitAMDGPUExt/TensorKitAMDGPUExt.jl
kshyatt: This is for the `DiagonalAlgorithm` (where the input `A` is itself `Diagonal`), so in that case, we do require that `S` is `Diagonal`.

src/tensors/linalg.jl
lkdvos: Would it make sense to have allowscalar here instead? I don't know anything about these performances, but it seems hard to imagine that computing the max/min is faster than just taking the element

kshyatt: Unsure, we could test it. Generically usage of `@allowscalar` is extremely discouraged

kshyatt: A possibly better solution would be to rewrite this entire for-loop to use broadcasting

lkdvos: If we don't mind touching the internals of the DiagonalTensorMap we could also just take the minimum and maximum of the entire vector in one call as well

src/tensors/tensor.jl
kshyatt: `copy!` now working well, I think I had a broken cuTENSOR before

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+
+const AdjointROCTensorMap{T, S, N₁, N₂} = AdjointTensorMap{T, S, N₁, N₂, ROCTensorMap{T, S, N₁, N₂}}
+
+function TensorKit.tensormaptype(S::Type{<:IndexSpace}, N₁, N₂, TorA::Type{<:StridedROCArray})
+    if TorA <: ROCArray
+        return TensorMap{eltype(TorA), S, N₁, N₂, ROCVector{eltype(TorA), AMDGPU.Mem.HIPBuffer}}
+    else
+        throw(ArgumentError("argument $TorA should specify a scalar type (`<:Number`) or a storage type `<:ROCVector{<:Number}`"))
+    end
+end
```
lkdvos, line 12: Could you maybe explain where this definition comes from? I would have guessed that if we ask for the `tensormaptype` with a given `A`, this function really should return the `TensorMap{..., A}`, which would be the default implementation but is overwritten here. I'm guessing I'm missing something, but it might be good to add a comment in that case to explain why this is necessary.

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
+const ROCTensorMap{T, S, N₁, N₂} = TensorMap{T, S, N₁, N₂, ROCVector{T, AMDGPU.Mem.HIPBuffer}}
```
lkdvos, line 1: I don't know much about the AMD ecosystem, but would there ever be a point to be less restrictive about this, and define this as follows, possibly with a const to determine what the allowed array types are?


```suggestion
const ROCTensorMap{T, S, N₁, N₂, A <: DenseRocVector{T}} = TensorMap{T, S, N₁, N₂, A}
```

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+    if TorA <: ROCArray
+        return TensorMap{eltype(TorA), S, N₁, N₂, ROCVector{eltype(TorA), AMDGPU.Mem.HIPBuffer}}
+    else
+        throw(ArgumentError("argument $TorA should specify a scalar type (`<:Number`) or a storage type `<:ROCVector{<:Number}`"))
+    end
+end
+
+function ROCTensorMap{T}(::UndefInitializer, V::TensorMapSpace{S, N₁, N₂}) where {T, S, N₁, N₂}
+    return ROCTensorMap{T, S, N₁, N₂}(undef, V)
+end
```
lkdvos, line 16: Thinking out loud here:
Would it also work if we rewrite the undef constructors logic in `tensors.jl` in a pattern like the following, using an additional intermediate step that selects the storage type, rather than doing that all in one go?

```julia
const _TensorMap{T, A <: DenseVector{T}, S, N1, N2} = TensorMap{T, S, N1, N2, A}
const DefaultTensorData{T} = Vector{T}

TensorMap{T}(::UndefInitializer, V::TensorMapSpace) where {T} =
    _TensorMap{T, DefaultTensorData{T}}(undef, V)
_TensorMap{T, A}(::UndefInitializer, V::TensorMapSpace) where {T, A} =
    TensorMap{T, spacetype(V), numout(V), numin(V), A}(undef, V)
```

and similar constructions for the other ones? This might avoid having to duplicate these constructors for AMD and CUDA, and could also pave the way for experiments with using `TensorMap{T, S, N1, N2, Memory{T}}`, which I'm guessing has very similar kinds of demands (and could be used to test the constructors etc without having to rely on buildkite?)

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+  each coupled sector `c` as a matrix of size `(blockdim(codomain, c), blockdim(domain, c))`.
+- `codomain::ProductSpace{S,N₁}`: the codomain as a `ProductSpace` of `N₁` spaces of type
+  `S<:ElementarySpace`.
+- `domain::ProductSpace{S,N₂}`: the domain as a `ProductSpace` of `N₂` spaces of type
+  `S<:ElementarySpace`.
+
+Alternatively, the domain and codomain can be specified by passing a [`HomSpace`](@ref)
+using the syntax `codomain ← domain` or `domain → codomain`.
+"""
+function ROCTensorMap(
```
lkdvos, line 47: Double comment here:

1. I think we probably also want `TensorMap(::SectorDict{I,<:ROCMatrix}, ...)` to return a `RocTensorMap`, which I don't think is currently happening.
2. We might also want `RocTensorMap(::SectorDict{I,<:AbstractMatrix})` to automatically promote to `ROCVector`?

Given this, and my previous comment, would it make sense to alter this constructor to first attempt to figure out a type `A`, and then call `TensorMap{T,...,A}(data)` which then enforces the data type?

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+        data::AbstractDict{<:Sector, <:ROCArray},
```
lkdvos, line 48:
```suggestion
        data::AbstractDict{<:Sector, <:ROCMatrix},
```

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+function ROCTensorMap(
+        data::AbstractDict{<:Sector, <:ROCMatrix}, codom::TensorSpace{S},
+        dom::TensorSpace{S}
+    ) where {S}
+    return ROCTensorMap(data, codom ← dom)
+end
+
+function ROCTensorMap(ts::TensorMap{T, S, N₁, N₂, A}) where {T, S, N₁, N₂, A}
+    return ROCTensorMap{T, S, N₁, N₂}(ROCArray(ts.data), ts.space)
+end
```
lkdvos, line 81: This might also go in the main package and can be generic?
```julia
(_TensorMap{T, A})(t::TensorMap) where {T, A} = _TensorMap(A(t.data), space(t))
```

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+            Base.eval(Main, Meta.parse(c)) => ROCArray(b)
+                for (c, b) in d[:data]
+        )
+        return ROCTensorMap(data, codomain, domain)
+    end
+end
+
+function Base.convert(::Type{ROCTensorMap}, t::AbstractTensorMap)
+    return copy!(ROCTensorMap{scalartype(t)}(undef, space(t)), t)
+end
```
lkdvos, line 165: This also feels like we could be more careful with defining the correct generic converters in the main package

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+
+    # TODO: should scalar only work if N₁ == N₂ == 0?
+    return @allowscalar dim(codomain(t)) == dim(domain(t)) == 1 ?
+        first(blocks(t))[2][1, 1] : throw(DimensionMismatch())
+end
+
+TensorKit.scalartype(A::StridedROCArray{T}) where {T} = T
+TensorKit.scalartype(::Type{<:ROCTensorMap{T}}) where {T} = T
+TensorKit.scalartype(::Type{<:ROCArray{T}}) where {T} = T
+
```
lkdvos, line 179:
```suggestion
```
I think this would be handled by the other definitions already?

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+        TT::Type{ROCTensorMap{T, S, N₁, N₂}},
+        t::AbstractTensorMap{<:Any, S, N₁, N₂}
+    ) where {T, S, N₁, N₂}
+    if typeof(t) === TT
+        return t
+    else
+        tnew = TT(undef, space(t))
+        return copy!(tnew, t)
+    end
+end
```
lkdvos, line 194: Is this not covered by the generic https://github.com/QuantumKitHub/TensorKit.jl/blob/a93edc663d41db67a77199c739b5526c37d008e2/src/tensors/tensor.jl#L525-L534

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+
```
lkdvos, line 195: Is this not covered by the generic https://github.com/QuantumKitHub/TensorKit.jl/blob/a93edc663d41db67a77199c739b5526c37d008e2/src/tensors/tensor.jl#L525-L534

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+function Base.copy!(tdst::ROCTensorMap{T, S, N₁, N₂}, tsrc::ROCTensorMap{T, S, N₁, N₂}) where {T, S, N₁, N₂}
+    space(tdst) == space(tsrc) || throw(SpaceMismatch("$(space(tdst)) ≠ $(space(tsrc))"))
+    for ((c, bdst), (_, bsrc)) in zip(blocks(tdst), blocks(tsrc))
+        copy!(bdst, bsrc)
+    end
+    return tdst
```
lkdvos, line 201: I'm guessing this is here to avoid going through the `StridedView` path which doesn't work on GPU? It might be reasonable to just remove the `StridedView` method, since this historically was there before everything was in a single large vector in `TensorMap`, and I would say this is a bit too opinionated for a generic `AbstractTensorMap` implementation.

Additionally, does this not coincide with the `Base.copy!(::TensorMap, ::TensorMap)` definition?

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+end
+
+function Base.copy!(tdst::ROCTensorMap, tsrc::TensorKit.AdjointTensorMap)
```
lkdvos, line 204: same comment about possibly just having this implementation as the `copy!(::AbstractTensorMap, ::AbstractTensorMap)` implementation

ext/TensorKitAMDGPUExt/roctensormap.jl
```diff
@@ -0,0 +1,257 @@
...
+        isherm || return false
+        isposdef(Hermitian(b)) || return false
+    end
+    return true
+end
+
+# Conversion to ROCArray:
+#----------------------
+# probably not optimized for speed, only for checking purposes
+function Base.convert(::Type{ROCArray}, t::AbstractTensorMap)
```
lkdvos, line 240: Can we rewrite the base version to take `Type{A} where {A <: Array}`, rather than hardcoding `Array` there?

ext/TensorKitAMDGPUExt/TensorKitAMDGPUExt.jl
```diff
@@ -0,0 +1,108 @@
...
+    return U, S, Vᴴ
+end
+
+function TensorKit.Factorizations.MAK.initialize_output(::typeof(svd_vals!), t::ROCTensorMap, alg::AbstractAlgorithm)
+    V_cod = infimum(fuse(codomain(t)), fuse(domain(t)))
+    return ROCDiagonalTensorMap{real(scalartype(t))}(undef, V_cod)
```
lkdvos, line 62: We could consider introducing a `similar_diagonal(t, T)` function/hook in the base implementations, such that we could overload just that function and avoid having to copy the boilerplate here

lkdvos: [edit] I hadn't seen you already more or less did this for the base implementation, in which case these functions might not be necessary anymore?

lkdvos (commented):
I went through the constructors for now, mostly with an eye of trying to minimize the amount of code duplication. Definitely feel free to discard some of these comments if that doesn't make too much sense, but I think there are some viable candidate ideas in there that would improve the overall robustness of `TensorMap` constructors/converters to begin with, and would simultaneously reduce the amount of boilerplate in the CUDA and AMD extensions

ext/TensorKitAMDGPUExt/roctensormap.jl
Jutho: I agree, but I think we want to alter the `TensorMap(data::AbstractDict{<:Sector, <:AbstractMatrix}, ...)` for that, in such a way that we indeed first determine a suitable `A<:DenseMatrix` associated with the types in the `values(data)` list.

lkdvos: with the caveat that `A <: DenseVector` 😉

Jutho: Potato potato

lkdvos: true, although it can be slightly tricky to find a generic implementation of a function that takes in `M <: AbstractMatrix` and spits out `A <: AbstractVector` that corresponds to this, which is more or less why I brought it up

Jutho: Absolutely, I have been struggling with that before. This was just my funny Friday evening comment.

ext/TensorKitAMDGPUExt/roctensormap.jl
kshyatt: Can try!

ext/TensorKitAMDGPUExt/roctensormap.jl
kshyatt: As you can likely tell, I tried to have as light a footprint in `src` as possible (in part to avoid breaking things that already work), but I think it makes sense to try to make things in there more flexible

Jutho: I think there are many places in the main TensorKit src where we are currently explicitly allocating tensors with a `Vector` storage (basically any occurrence of the `TensorMap{T}` constructor or `DiagonalTensorMap{T}` constructor), and where we will need a better allocation strategy. So it would be great to flesh these out. We can perhaps schedule a meeting next week to then discuss a strategy of how to proceed with all of those.

kshyatt: Yes, I think that would make sense. It could be as simple as passing a parameter `::Type{A} where {A<:DenseVector}` for the storage type and the default being `Vector`.

Jutho: Note that even our latest PR #324 (which is intended to help simplify doing minimum over singular values etc, because also `svd_vals` now returns them in a new type that has them stored in a single vector) will have those issues of default Vector allocation.

Jutho:
Superseded by #325 and more.

PR #321, opened 2025-12-01
Bump v0.16 + Remove old deprecations

lkdvos:
This PR removes the `deprecate.jl` file, which mostly contained factorization code.
The goal is to free up these names again, which could then be reused in further updates without breaking changes, if we wish to do so.

There were some older deprecated functions in that file as well, that had been there forever (Some even 2+ years), which I think is fine to also delete.

Finally, I removed the PackageExtensionCompat dependency, since that was used for supporting package extensions in Julia <v1.9, which we no longer support anyways.

(I also bumped the version and took the liberty of adding myself to the author list, if you don't mind)

This would probably best be merged after #291 and #316, and can then be used to register v0.16, for which I'll try to write some release notes/changelog and add this to the PR.

### DIFF
--- a/docs/src/Changelog.md
+++ b/docs/src/Changelog.md
@@ -20,7 +20,7 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 ## [Unreleased](https://github.com/QuantumKitHub/TensorKit.jl/compare/v0.16.0...HEAD)
 
-## [0.16.0](https://github.com/QuantumKitHub/TensorKit.jl/releases/tag/v0.16.0) - 2025-12-03
+## [0.16.0](https://github.com/QuantumKitHub/TensorKit.jl/releases/tag/v0.16.0) - 2025-12-05
 
 ### Added
 
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -10,7 +10,7 @@ authors:
 title: "TensorKit.jl"
 version: "0.16.0"
 doi: "10.5281/zenodo.8421339"
-date-released: "2025-10-03"
+date-released: "2025-12-05"
 url: "https://github.com/QuantumKitHub/TensorKit.jl"
 preferred-citation:
   type: article


docs/src/Changelog.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Added
+
+- `rrule` for `transpose` operation ([#319](https://github.com/QuantumKitHub/TensorKit.jl/pull/319))
+- New functions for multifusion support: `unitspace`, `zerospace`, `leftunitspace`, `rightunitspace`, `isunitspace` ([#291](https://github.com/QuantumKitHub/TensorKit.jl/pull/291))
+- Support for projections and orthogonal complements ([#312](https://github.com/QuantumKitHub/TensorKit.jl/pull/312))
+
+### Changed
+
+- Tensors now no longer print their data by default, only their spaces. Use `blocks(t)` or `subblocks(t)` to inspect data ([#304](https://github.com/QuantumKitHub/TensorKit.jl/pull/304))
```
Jutho, line 33: I forgot about the status of that. I do think people are not too happy with no output being shown; would it be so bad to shows blocks(t) by default?

Jutho: Or is this having a lot of repercussions again on MPSKit etc?

lkdvos: We definitely used to show the output of `subblocks(t)`, which we could add back here (even though I strongly prefer `blocks(t)`, I think that is what people wanted?).
I think MPSKit wouldn't be affected since that would have to use the `compact = true` setting anyways, which shouldn't print this.
If we were to do this, I would however want to limit the output of that by default.
While I agree that for small tensors this can be useful, this is a nuisance for larger tensors...
Do you think this constitutes breaking changes though?

lkdvos:
I tried to further stabilize the tests, but it does seem like playing whack-a-mole with the random number generator again...
I would be fine with merging this as-is, although we might consider updating the readme changelog and coming up with some more release notes?

Jutho:
Looks good to me, but there is a failing AD test on LTS, so that would preferably be fixed. I am to tired at this point to investigate myself but can do so tomorrow.

PR #322, opened 2025-12-03
further tweak tensor and block show

Jutho:
@lkdvos , this tries to get some sensible default for showing some blocks while respecting the displaysize when `:limit = true`. Since standard repl printing sets `:limit = true`, this is the default printing.

However, with the current implementation, this also means that the blocks will be truncated (both the matrices itself and the number of blocks shown) in both the output of the tensor `t` and in the output of `blocks(t)`.

There might be a case to be made that, if a user calls `blocks(t)` explicitly as a REPL prompt, the user wants to see all blocks, and so we should "ignore" the default `:limit = true` setting of REPL printing, at least for truncating the number of blocks, but we could still pass on `:limit = true` to the individual block printing. This would then be analogous to how an array of size n x n x 10 with large n is shown. All 10 slices along the 3rd dimension are shown, but every slice is truncated such that single slice fits the display.


src/tensors/blockiterator.jl
```diff
@@ -45,13 +45,40 @@ function foreachblock(f, t::AbstractTensorMap; scheduler = nothing)
...
-function show_blocks(io, mime::MIME"text/plain", iter)
-    first = true
-    for (c, b) in iter
-        first || print(io, "\n\n")
-        print(io, " * ", c, " => ")
-        show(io, mime, b)
-        first = false
+function show_blocks(io, mime::MIME"text/plain", iter; maytruncate::Bool = true)
+    if maytruncate && get(io, :limit, false)
+        numlinesleft, numcols = get(io, :displaysize, displaysize(io))
```
lkdvos: I'm not sure if this actually matters, but this could improve type-stability/invalidations? I'm mostly just guessing though, feel free to ignore this.

```suggestion
        numlinesleft, numcols = get(io, :displaysize, displaysize(io))::Tuple{Int, Int}
```

src/tensors/blockiterator.jl
```diff
@@ -73,7 +100,9 @@ end
 function Base.show(io::IO, mime::MIME"text/plain", b::BlockIterator)
     summary(io, b)
     println(io, ":")
-    show_blocks(io, mime, b)
+    (numlines, numcols) = get(io, :displaysize, displaysize(io))
```
lkdvos:
```suggestion
    (numlines, numcols) = get(io, :displaysize, displaysize(io))::Tuple{Int, Int}
```

lkdvos (approved):
Do you think it would make sense to have some dummy tests that just checks that the output of showing a tensor is what it is now, such that we get notified if we (accidentally/intentionally) change this in the future?

Otherwise this looks ready for me.

PR #325, opened 2025-12-08
Start on CUDA extension

kshyatt:
This splits off *basic* support for CUDA tensor maps from https://github.com/QuantumKitHub/TensorKit.jl/pull/320, and doesn't cover Diagonals or factorizations at all. I made some modifications to things in `src` to make them a little more generic also. Some things aren't working yet:

- converting from an `AdjointTensorMap` of a `CuTensorMap` to a `CuTensorMap` of a different scalar type
- converting to a `CuArray`
- the `hasfusiontensor(I)` tests
- `diag`/`diagm`
- index flipping


ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,225 @@
...
+#-----------------------
+function TensorKit.scalar(t::CuTensorMap)
+    # TODO: should scalar only work if N₁ == N₂ == 0?
+    return @allowscalar dim(codomain(t)) == dim(domain(t)) == 1 ?
+        first(blocks(t))[2][1, 1] : throw(DimensionMismatch())
+end
+
+TensorKit.scalartype(A::StridedCuArray{T}) where {T} = T
+TensorKit.scalartype(::Type{<:CuTensorMap{T}}) where {T} = T
+TensorKit.scalartype(::Type{<:CuArray{T}}) where {T} = T
```
lkdvos:
```suggestion
```

This shouldn't be necessary I think

src/tensors/linalg.jl
```diff
@@ -299,7 +299,7 @@ function LinearAlgebra.rank(
     r = 0 * dim(first(allunits(sectortype(t))))
     dim(t) == 0 && return r
     S = LinearAlgebra.svdvals(t)
```
Jutho: With `MAK.svd_vals` and the new `SectorVector`, you should be able to immediately replace the things below by `maximum(S.data)`. Though probably we should just add `maximum(S::SectorVector) = maximum(s.data)` and the same for `minimum`.

src/tensors/tensor.jl
```diff
@@ -198,6 +198,15 @@ TensorMap(data::AbstractArray, codom::TensorSpace, dom::TensorSpace; kwargs...)
 Tensor(data::AbstractArray, codom::TensorSpace; kwargs...) =
     TensorMap(data, codom ← one(codom); kwargs...)
 
+function _project_symmetric_and_check(::Type{T}, ::Type{A}, data::AbstractArray, V::TensorMapSpace; tol = sqrt(eps(real(float(eltype(data)))))) where {T, A <: DenseVector{T}}
```
Jutho: I would be ok with this not having a leading `_`. Many internal methods do not start with `_`; I think I only use that for very low level things, that are probably unsafe (though then it is often named as such).

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,226 @@
...
```
Jutho: Are these not automatic?

kshyatt: These might be artifacts from this PR having evolved over a period of months, I'll try removing them and see what happens...

src/tensors/linalg.jl
```diff
@@ -281,7 +281,7 @@ function _norm(blockiter, p::Real, init::Real)
         end
     elseif p > 0
         nᵖ = mapreduce(+, blockiter; init = init) do (c, b)
-            return isempty(b) ? init : oftype(init, dim(c) * LinearAlgebra.normp(b, p)^p)
+            return isempty(b) ? init : oftype(init, dim(c) * LinearAlgebra.norm(b, p)^p)
```
lkdvos: Looking at these implementations, if we just use `norm` everywhere anyways, isthere any performance benefit to manually splitting this out?
How about just replacing all of the `isfinite(p)` cases with:

```julia
np = sum(blockiter; init) do (c, b)
    return oftype(init, dim(c) * norm(b, p)^p)
end
return np^(inv(oftype(np, p)))
```

src/tensors/diagonal.jl
```diff
@@ -273,7 +273,7 @@ function LinearAlgebra.mul!(
         dC::DiagonalTensorMap, dA::DiagonalTensorMap, dB::DiagonalTensorMap, α::Number, β::Number
     )
     dC.domain == dA.domain == dB.domain || throw(SpaceMismatch())
-    mul!(Diagonal(dC.data), Diagonal(dA.data), Diagonal(dB.data), α, β)
+    @. dC.data = (α * dA.data * dB.data) + β * dC.data
```
lkdvos: Am I correct that this is to account for a missing GPU implementation? Would you mind putting that in a comment here?

ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
```diff
@@ -0,0 +1,27 @@
...
+
+using Random
+
+include("cutensormap.jl")
+
+# TODO
+# add VectorInterface extensions for proper CUDA promotion
+function TensorKit.VectorInterface.promote_add(TA::Type{<:CUDA.StridedCuMatrix{Tx}}, TB::Type{<:CUDA.StridedCuMatrix{Ty}}, α::Tα = TensorKit.VectorInterface.One(), β::Tβ = TensorKit.VectorInterface.One()) where {Tx, Ty, Tα, Tβ}
+    return Base.promote_op(add, Tx, Ty, Tα, Tβ)
+end
```
lkdvos: I'm a bit confused by this change, we might have to revisit the VectorInterface implementations (and I should probably clarify these with some docstrings), but I think this method should really just depend on the scalar types, and output a resulting scalar type. Therefore, as long as `scalartype` is implemented correctly, I would expect that this method does not have to be overloaded?

pasting the implementation below for convenience:

```julia
function promote_add(x, y, α::Number = One(), β::Number = One())
    return promote_add(scalartype(x), scalartype(y), typeof(α), typeof(β))
end
function promote_add(
        ::Type{Ty}, ::Type{Tx},
        ::Type{Tα} = One, ::Type{Tβ} = One
    ) where {Ty, Tx, Tα <: Number, Tβ <: Number}
    return Base.promote_op(add, Tx, Ty, Tα, Tβ)
end
```

(i.e. the former extracts scalar types, the latter tries to apply `add` on the scalar types `<:Number`)

src/tensors/linalg.jl
```diff
@@ -299,7 +299,7 @@ function LinearAlgebra.rank(
     r = 0 * dim(first(allunits(sectortype(t))))
     dim(t) == 0 && return r
     S = LinearAlgebra.svdvals(t)
-    tol = max(atol, rtol * maximum(first, values(S)))
+    tol = max(atol, rtol * maximum(S.data))
```
lkdvos:
```suggestion
    tol = max(atol, rtol * maximum(parent(S)))
```
minor style change to avoid depending on fields

Jutho:
Haven't looked at the tests in detail yet, but certainly looks very good already.

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,222 @@
...
+                V::TensorMapSpace
+            ) where {T}
+            t = CuTensorMap{T}(undef, V)
+            $randfun!(rng, t)
+            return t
+        end
+    end
+end
+
+for randfun in (:rand, :randn, :randisometry)
```
lkdvos: Can we avoid these definitions by changing the `undef` call in the `rand` constructors?
i.e here: https://github.com/QuantumKitHub/TensorKit.jl/blob/d60855e4acfc53209455f873a8105300d9aa95ff/src/tensors/tensor.jl#L388

```julia
t = tensormaptype(spacetype(V), numout(V), numin(V), TorA)(undef, V)
```

I don't think we have been restricting the `T` in these functions, so we can just rename all of them `TorA` and handle both cases in one go?

As a separate comment/question: I see you chose to have `rand(T, A, ...)` here, I think I would be fine with merging these arguments `rand(TorA, ...)` and using dispatch to distinguish between `<:Number` subtypes and other types. I think this is in line with the rest of the package, and would be automatically fulfilled by the suggestion above.

kshyatt:
Re the tests, they are mostly copied from the CPU `tensors` tests. I have added a few to test CUDA specific code paths and to test conversion to/from CPU based `TensorMap`. Some tests do not work yet (the basic linear algebra ones, `sylvester`) because some of the functionality isn't yet supported on CUDA.

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,222 @@
...
+            $randfun!(rng, t)
+            return t
+        end
+    end
+end
+
+function Base.convert(::Type{CuTensorMap}, t::AbstractTensorMap{T, S}) where {T, S}
+    d_data = CuArray(t.data)
+    return TensorMap{T}(d_data, t.space)
+end
```
lkdvos: Can we move this to the main package and implement it in general for `TensorMapWithStorage`, or is this a weird special case? these dispatch things with parametric types where not all types are filled in are definitely confusing.

As a sidenote, this method is really only valid for `::TensorMap` inputs since we are accessing the `.data` field directly

```julia
function Base.convert(::Type{TensorMapWithStorage{T, A}}, t::TensorMap) where {T, A}
    d_data = convert(A, t.data)
    return TensorMapWithStorage{T}(d_data, space(t))
end
```

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,222 @@
...
+
+# Scalar implementation
+#-----------------------
+function TensorKit.scalar(t::CuTensorMap)
+    # TODO: should scalar only work if N₁ == N₂ == 0?
+    return @allowscalar dim(codomain(t)) == dim(domain(t)) == 1 ?
+        first(blocks(t))[2][1, 1] : throw(DimensionMismatch())
+end
+
+function Base.convert(
```
lkdvos, line 104: Here I'm a bit confused, since `CuTensorMap{T, S, N1, N2} = TensorMap{t, S, N1, N2, CuVector{T}}` I would expect that this is already just caught by the default converter?

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,222 @@
...
+    return TensorMap{T}(d_data, t.space)
+end
+
+# Scalar implementation
+#-----------------------
+function TensorKit.scalar(t::CuTensorMap)
+    # TODO: should scalar only work if N₁ == N₂ == 0?
+    return @allowscalar dim(codomain(t)) == dim(domain(t)) == 1 ?
+        first(blocks(t))[2][1, 1] : throw(DimensionMismatch())
+end
```
lkdvos: This still needs to be updated for the multitensorkit changes.
I took the liberty here of making it slightly more efficient by leveraging the fact that it doesn't have to be implemented for `AbstractTensorMap` and accessing the data field directly


```suggestion
function TensorKit.scalar(t::CuTensorMap{T, S, 0, 0}) where {T, S}
    inds = findall(!iszero, t.data)
    return isempty(inds) ? zero(scalartype(t)) : @allowscalar @inbounds t.data[only(inds)
end
```

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,222 @@
...
+
+function Base.promote_rule(
+        ::Type{<:TT₁},
+        ::Type{<:TT₂}
+    ) where {
+        S, N₁, N₂, TTT₁, TTT₂,
+        TT₁ <: CuTensorMap{TTT₁, S, N₁, N₂},
+        TT₂ <: CuTensorMap{TTT₂, S, N₁, N₂},
+    }
+    T = TensorKit.VectorInterface.promote_add(TTT₁, TTT₂)
```
lkdvos, line 137: Just realized why you had the `VectorInterface` overload, this is just a wrong usage from our (my) part in the base package of `promote_add`.

Probably the base implementation could become something like the following, using the hook I added in the last PR for `tensorcontract_type` as well:

```julia
function Base.promote_rule(
        ::Type{<:TT₁}, ::Type{<:TT₂}
    ) where {S, N₁, N₂, TT₁ <: TensorMap{<:Any, S, N₁, N₂}, TT₂ <: TensorMap{<:Any, S, N₁, N₂}}
    T = VectorInterface.promote_add(scalartype(TT₁), scalartype(TT₂))
    A = promote_storagetype(similarstoragetype(TT₁, T), similarstoragetype(TT₂, T))
    return tensormaptype(S, N₁, N₂, A)
end
```

ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
lkdvos: [edit] figured out why this is here, probably my bad we needed this in the first place 😉

src/tensors/linalg.jl
```diff
@@ -317,8 +317,8 @@ function LinearAlgebra.cond(t::AbstractTensorMap, p::Real = 2)
             return zero(real(float(scalartype(t))))
         end
         S = LinearAlgebra.svdvals(t)
-        maxS = maximum(first, values(S))
-        minS = minimum(last, values(S))
+        maxS = maximum(S.data)
+        minS = minimum(S.data)
```
lkdvos: Same comment about `parent` but apparently I'm not allowed to make suggestions for these lines

src/tensors/tensor.jl
```diff
@@ -411,12 +412,12 @@ function Base.convert(::Type{Dict}, t::AbstractTensorMap)
     d[:domain] = repr(domain(t))
     data = Dict{String, Any}()
     for (c, b) in blocks(t)
-        data[repr(c)] = Array(b)
+        data[repr(c)] = b
```
lkdvos: This isn't entirely what we want I think, we really do want to store this using easy data types, and like this we would be storing reshaped views rather than arrays. Given that this is meant for storing data on disk, I'm not sure we ever want to not collect data to the CPU, so maybe it's fine to use `collect` here instead?

src/tensors/tensor.jl
```diff
@@ -411,12 +412,12 @@ function Base.convert(::Type{Dict}, t::AbstractTensorMap)
...
     end
     d[:data] = data
     return d
 end
-function Base.convert(::Type{TensorMap}, d::Dict{Symbol, Any})
+function Base.convert(::Type{<:TensorMap}, d::Dict{Symbol, Any})
```
lkdvos: Similarly here, I think either we have to add some type parameters to determine the output storage type, or just not bother with having this on GPU (for now) and manually send the data to GPU after loading it in

src/tensors/tensor.jl
lkdvos: Since I realize this is completely not obvious, probably it's a good idea to also add some comment about this method to begin with...

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+            end
+        end
+        @timedtestset "Tensor Dict conversion" begin
+            W = V1 ⊗ V2 ⊗ V3 ← V4 ⊗ V5
+            for T in (Int, Float32, ComplexF64)
+                t = @constinferred CUDA.rand(T, W)
+                d = convert(Dict, t)
+                @test t == convert(TensorMap, d)
+            end
+        end
```
lkdvos, line 125: As mentioned above, it might be fair to not support this

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+                t5 = @constinferred insertleftunit(t, 4; dual = true)
+                @test numin(t5) == numin(t) + 1 && numout(t5) == numout(t)
+                for (c, b) in blocks(t)
+                    @test b == block(t5, c)
+                end
+                @test @constinferred(removeunit(t5, 4)) == t
+            end
+        end=#
+        if hasfusiontensor(I)
+            @timedtestset "Basic linear algebra: test via conversion" begin
```
lkdvos: How do you feel about `"test via CPU"` instead? I.e. instead of checking that the linear algebra commutes with `convert(Array, x)`, we check if it commutes with `collect`. This means that we can still test methods that don't have `hasfusiontensor`, and this should be faster, while being as good of a test.
Additionally, this avoids the issue about how to deal with array conversions from GPU for now

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+                        @test convert(Array, t + t) ≈ 2 * convert(Array, t)
+                    end
+                end
+            end
+            @timedtestset "Real and imaginary parts" begin
+                W = V1 ⊗ V2
+                for T in (Float64, ComplexF64, ComplexF32)
+                    t = @constinferred CUDA.randn(T, W, W)
+
+                    tr = @constinferred real(t)
```
lkdvos, line 226: Might be worth it to verify that the storagetype remained GPU as well

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+                    @test scalartype(tr) <: Real
+                    CUDA.@allowscalar begin
+                        @test real(convert(Array, t)) == convert(Array, tr)
+                    end
```
lkdvos:
```suggestion
                    @test real(collect(t)) == collect(tr)
```

test/cuda/tensors.jl
lkdvos: I said `collect` here, assuming that this is what we want to overload to get data from GPU to CPU, happy to discuss other options as well

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+
+                    ti = @constinferred imag(t)
+                    @test scalartype(ti) <: Real
```
lkdvos, line 232: same comment here

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+                    CUDA.@allowscalar begin
+                        @test imag(convert(Array, t)) == convert(Array, ti)
+                    end
+
+                    tc = @inferred complex(t)
```
lkdvos, line 236: same here

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+                    end
+
+                    t3 = CUDA.@allowscalar repartition(t, k)
+                    a3 = CUDA.@allowscalar convert(Array, t3)
+                    @test a3 ≈ permutedims(
+                        a, (ntuple(identity, k)..., reverse(ntuple(i -> i + k, 5 - k))...)
+                    )
+                end
+            end
+        end
```
lkdvos: Here again, this is quite an expensive part of the testsuite, which we can just replace by collecting to CPU instead

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+
+                    @tensor HrA12array[a, s1, s2, c] := ad(rhoL)[a, a'] *
+                        conj(ad(A1)[a', t1, b]) *
+                        ad(A2)[b, t2, c'] *
+                        ad(rhoR)[c', c] *
+                        ad(H)[s1, s2, t1, t2]
+                end
+                @test HrA12array ≈ ad(HrA12)
+            end
+        end
```
lkdvos: Same here about testing vs CPU implementation

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+                @tensor tb[a, b] := flip(t1, (1, 3))[x, y, a, z] * flip(t2, (2, 4))[y, b, z, x]
+                @test flip(ta, (1, 2)) ≈ tb
+            end
+        end
+        @timedtestset "Multiplication of isometries: test properties" begin
+            W2 = V4 ⊗ V5
+            W1 = W2 ⊗ (oneunit(V1) ⊕ oneunit(V1))
+            for T in (Float64, ComplexF64)
+                t1 = randisometry(T, CuMatrix{T}, W1, W2)
+                t2 = randisometry(T, CuMatrix{T}, W2 ← W2)
```
lkdvos:
```suggestion
                t1 = randisometry(CuMatrix{T}, W1, W2)
                t2 = randisometry(CuMatrix{T}, W2 ← W2)
```

test/cuda/tensors.jl
```diff
@@ -0,0 +1,648 @@
...
+                @test one(t1) / t1 ≈ pinv(t1)
+                @test_throws SpaceMismatch inv(t)
+                @test_throws SpaceMismatch t2 \ t
+                @test_throws SpaceMismatch t / t1
+                tp = pinv(t) * t
+                @test tp ≈ tp * tp
+            end
+        end
+        if BraidingStyle(I) isa Bosonic && hasfusiontensor(I)
+            @timedtestset "Multiplication and inverse: test via conversion" begin
```
lkdvos: Same here

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,222 @@
...
+    return CuTensorMap{T, S, N₁, N₂}
+end
+
+# CuTensorMap exponentation:
+function TensorKit.exp!(t::CuTensorMap)
+    domain(t) == codomain(t) ||
+        error("Exponential of a tensor only exist when domain == codomain.")
+    for (c, b) in blocks(t)
+        copy!(b, parent(Base.exp(Hermitian(b))))
```
lkdvos, line 147: Does this automatically error if the input isn't hermitian? Can we also use `exp!`?

src/tensors/diagonal.jl
kshyatt: Can do

kshyatt: I can also see if we still actually need this, I added a lot of `Diagonal` support to `GPUArrays` the last few months

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Yeah can do, this is another older artifact in here

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Yeah, I think there was some annoying dispatch issue with the memory type of the backing `CuVector`, but I can remove this and see if things play nicely

src/tensors/linalg.jl
kshyatt: What, what about the free speech rights of Lukas????

src/tensors/tensor.jl
kshyatt: Yeah there are a bunch of places where you can see me making a "best guess" based on rather cryptic stuff I didn't quite see the full picture of... 😅

test/cuda/tensors.jl
kshyatt: `collect` should be ok for now

ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
Jutho: I am still confused: why is this needed? How is it different from the general definition?

src/tensors/diagonal.jl
kshyatt: Either way this is a bit out of scope for this PR which doesn't deal with Diagonals (yet)

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: CUDA doesn't have `exp!` defined (because there's a copy no matter what)

ext/TensorKitCUDAExt/cutensormap.jl
lkdvos: @Jutho this is where that promote_add overload came from, but I think this was really a bit of an accident that we were using that function wrongly in the promotion rule but this accidentally gave the right result for regular Vector storage

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,161 @@
...
+
+const AdjointCuTensorMap{T, S, N₁, N₂} = AdjointTensorMap{T, S, N₁, N₂, CuTensorMap{T, S, N₁, N₂}}
+
+function CuTensorMap(t::TensorMap{T, S, N₁, N₂, A}) where {T, S, N₁, N₂, A}
+    return CuTensorMap{T, S, N₁, N₂}(CuArray{T}(t.data), space(t))
+end
+
+function Base.collect(t::CuTensorMap{T}) where {T}
+    return convert(TensorKit.TensorMapWithStorage{T, Vector{T}}, t)
+end
```
lkdvos: Does this work as a generic definition too?

```julia
Base.collect(t::TensorMap) = convert(TensorMapWithStorage{scalartype(t), similarstoragetype(scalartype(t))}, t)
```

(also used `similarstoragetype(::Type{Number})` here to not hard-code the default storage type, as we might consider switching this to `Memory{T}` at some point)

test/cuda/tensors.jl
kshyatt: It's currently tested just to make sure you can in theory do this, but it compares with a CPU based `TensorMap` at the end.

kshyatt:
OK I think I've responded to everything 🧐

src/tensors/tensor.jl
```diff
@@ -315,11 +316,14 @@ end
...
-        $randf([rng=default_rng()], [T=Float64], codomain::ProductSpace{S,N₁},
+        $randf([rng=default_rng()], [TorA=Float64], codomain::ProductSpace{S,N₁},
                      domain::ProductSpace{S,N₂}) where {S,N₁,N₂,T} -> t
-        $randf([rng=default_rng()], [T=Float64], codomain ← domain) -> t
+        $randf([rng=default_rng()], [TorA=Float64], codomain ← domain) -> t
         
     Generate a tensor `t` with entries generated by `$randf`.
+    The type `TorA` can be used to control the element type and
+    data type generated. For example, if `TorA` is a `SparseVector{ComplexF32}`,
+    then the final output `TensorMap` will have that as its storage type.
```
Jutho: That might not be the best example, as we have `A<:DenseVector` in the `TensorMap` type. We might have to reconsider this at some point, but many implementations do implicitly assume this.

kshyatt: Changed to a GPU example instead

.buildkite/pipeline.yml
```diff
@@ -15,12 +15,12 @@ steps:
...
       cuda: "*"
     if: build.message !~ /\[skip tests\]/
-    timeout_in_minutes: 30
+    timeout_in_minutes: 60
     matrix:
       setup:
         julia:
           - "1.10"
-          - "1.11"
+          - "1.12"
```
Jutho, line 23: I guess there is no `lts` and `1` on buildkite?

kshyatt: Unfortunately, no 😞

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,157 @@
...
+            return $randfun(Random.default_rng(), T, V)
+        end
+
+        # implementation
+        function $randfun(
+                rng::Random.AbstractRNG, ::Type{T},
+                V::TensorMapSpace
+            ) where {T}
+            t = CuTensorMap{T}(undef, V)
+            $randfun!(rng, t)
```
Jutho, line 84: I am probably missing something, but where is `curand(n)!(rng, t)` itself defined?

kshyatt: No, you're right, that one was missed. I actually doubt we need these as they are only defined here (and cannot be exported), and we have a better (imo) solution using `rand(TorA)` as discussed above. We can always restore this if needed.

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,157 @@
...
+    end
+end
+
+function LinearAlgebra.isposdef(t::CuTensorMap)
+    domain(t) == codomain(t) ||
+        throw(SpaceMismatch("`isposdef` requires domain and codomain to be the same"))
+    InnerProductStyle(spacetype(t)) === EuclideanInnerProduct() || return false
+    for (c, b) in blocks(t)
+        # do our own hermitian check
+        isherm = TensorKit.MatrixAlgebraKit.ishermitian(b; atol = eps(real(eltype(b))), rtol = eps(real(eltype(b))))
```
Jutho: Since you already have `using TensorKit.MatrixAlgebraKit`, I don't know if the `qualification` here is necessary. It could also be nice to have `using TensorKit: MatrixAlgebraKit` and then only qualify names with `MatrixAlgebraKit` (or even `MAK`).

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,157 @@
...
+    T = TensorKit.VectorInterface.promote_add(TTT₁, TTT₂)
+    return CuTensorMap{T, S, N₁, N₂}
+end
+
+# CuTensorMap exponentation:
+function TensorKit.exp!(t::CuTensorMap)
+    domain(t) == codomain(t) ||
+        error("Exponential of a tensor only exist when domain == codomain.")
+    for (c, b) in blocks(t)
+        copy!(b, parent(Base.exp(Hermitian(b))))
```
Jutho, line 147: These `Hermitian` wrappers are dangerous no? They don't check for hermiticity, but simply assume that the matrix is hermitian and fully specified by its upper part. So this will silently produce incorrect results (also below in the other functions)

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -0,0 +1,157 @@
...
+# functions that don't map ℝ to (a subset of) ℝ
+for f in (:sqrt, :log, :asin, :acos, :acosh, :atanh, :acoth)
+    sf = string(f)
+    @eval function Base.$f(t::CuTensorMap)
+        domain(t) == codomain(t) ||
+            throw(SpaceMismatch("`$($sf)` of a tensor only exist when domain == codomain"))
+        T = complex(float(scalartype(t)))
+        tf = similar(t, T)
+        for (c, b) in blocks(t)
+            copy!(block(tf, c), parent($f(Hermitian(b))))
```
Jutho, line 162: Here as well.

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Yes, it's a bit dicey exactly because there is no check, but on the other hand the check itself can be very expensive. It's all downstream of the fact that CUDA only just got a general eigensolver so there's no support for these functions on generic `CuMatrix`...

src/tensors/diagonal.jl
```diff
@@ -78,12 +78,10 @@ function DiagonalTensorMap(t::AbstractTensorMap{T, S, 1, 1}) where {T, S}
...
-Base.similar(d::DiagonalTensorMap) = similar_diagonal(d)
-Base.similar(d::DiagonalTensorMap, ::Type{T}) where {T} = similar_diagonal(d, T)
-
-similar_diagonal(d::DiagonalTensorMap) = DiagonalTensorMap(similar(d.data), d.domain)
-similar_diagonal(d::DiagonalTensorMap, ::Type{T}) where {T <: Number} =
-    DiagonalTensorMap(similar(d.data, T), d.domain)
+Base.similar(d::DiagonalTensorMap) = DiagonalTensorMap(similar(d.data), d.domain)
+function Base.similar(d::DiagonalTensorMap, ::Type{T}) where {T <: Number}
+    return DiagonalTensorMap(similar(d.data, T), d.domain)
+end
```
Jutho: Is there a specific reason for this change? Is `similar_diagonal` then still defined for `DiagonalTensorMap` ?

src/tensors/linalg.jl
```diff
@@ -299,7 +290,7 @@ function LinearAlgebra.rank(
     r = 0 * dim(first(allunits(sectortype(t))))
     dim(t) == 0 && return r
     S = LinearAlgebra.svdvals(t)
```
Jutho: Not sure if this needs to be qualified with `MatrixAlgebraKit` or `MAK`, but I would prefer using "our" method.
```suggestion
    S = svd_vals(t)
```

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Or I can keep it in place but actually test it...

src/tensors/tensor.jl
```diff
@@ -399,12 +404,20 @@ for randf in (:rand, :randn, :randexp, :randisometry)
     end
 end
 
+# Collecting arbitrary TensorMaps
+#-----------------------------
+Base.collect(t::TensorMap) = convert(TensorMapWithStorage{scalartype(t), similarstoragetype(scalartype(t))}, t)
```
Jutho: I a not sure about this one. It has been my impression that `Base.collect` very specifically expects to collect all the data in some arbitrary collection into an `Array` object, so I don't think we can reuse this in this way. Where do you want to use this?

kshyatt: FWIW `collect` is used in exactly this way in `CUDA.jl`: https://github.com/JuliaGPU/CUDA.jl/blob/5d9474ae73fab66989235f7ff4fd447d5ee06f8e/test/base/array.jl#L9

src/tensors/linalg.jl
kshyatt: Fixed :)

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Added a check

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Added a check

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Added the inplace definition and tests

src/tensors/tensor.jl
kshyatt: It's used in the GPUArrays ecosystem as "collect to CPU"

src/tensors/diagonal.jl
kshyatt: I think it's another artifact from Lukas' diagonal PR getting merged before this

ext/TensorKitCUDAExt/cutensormap.jl
Jutho: Not necessary, but if you are still making changes, maybe the error message could be: `... is currently only supported ...`

I do hope we have have full support at some point, hopefully not to far away (Is this also not supported on standard matrices? The default way of computing matrix exponentials via Pade approximation should work on GPUs I believe)

kshyatt: 100% agreed, at least CUDA now has a general eigensolver (ROCm doesn't). Can easily update the error messages as I need to fix a few things re SVD values

src/tensors/tensor.jl
Jutho: Yes, but there the return value is an Array. This is from the manual

```julia-repl
help?> collect
search: collect coalesce convert @coalesce reflect!

  collect(iterator)

  Return an Array of all items in a collection or iterator. For dictionaries, returns a Vector of
  key=>value Pairs. If the argument is array-like or is an iterator with the HasShape trait, the
  result will have the same shape and number of dimensions as the argument.

  Used by comprehensions to turn a generator expression into an Array. Thus, on generators, the
  square-brackets notation may be used instead of calling collect, see second example.

  The element type of the returned array is based on the types of the values collected. However, if
  the iterator is empty then the element type of the returned (empty) array is determined by type
  inference.
```

So the result being an `Array` seems a pretty essential feature of `collect`.

kshyatt: In that case we could change this to `TensorMap{T, Vector{T}}` or similar? Or create a new `to_cpu` function, or use `Adapt.jl`? It would be nice to have a concise way of expressing this which is why we went with `collect` initially

Jutho: I think we definitely want Adapt compatibility. For the tests, a simple `to_cpu` function would probably be useful. Ultimately, I think this is again a shortcoming in Base, with no standard interface to move data structures between different address spaces :-). But that is exactly what Adapt aims to solve, right?

kshyatt: Should we split the difference and do `to_cpu` for now (for tests) and open an issue to add broad Adapt support in a future PR? We'll want that (in a somewhat generic way I'd think) for `ROCm` too

src/tensors/linalg.jl
kshyatt: This introduces a bunch of annoying issues with iteration over `DiagonalTensorMap` so I might go back to the LA method for now and open an issue to address this separately

Jutho: I thought both `svdvals` and `svd_vals` now return a `SectorVector`?

src/tensors/tensor.jl
kshyatt: Done!

src/tensors/linalg.jl
kshyatt: Indeed, maybe I was missing the `MatrixAlgebraKit` tag

test/cuda/tensors.jl
```diff
@@ -0,0 +1,609 @@
...
+using Test, TestExtras
+using TensorKit, Combinatorics
+ad = adapt(Array)
+const CUDAExt = Base.get_extension(TensorKit, :TensorKitCUDAExt)
+@assert !isnothing(CUDAExt)
+const CuTensorMap = getglobal(CUDAExt, :CuTensorMap)
+const curand = getglobal(CUDAExt, :curand)
+const curandn = getglobal(CUDAExt, :curandn)
+const curand! = getglobal(CUDAExt, :curand!)
+const curandn! = getglobal(CUDAExt, :curandn!)
```
lkdvos:
```suggestion
using CUDA: rand as curand, rand! as curand!, randn as curandn, randn! as curandn!
```

I think this doesn't really change anything, but I find it slightly easier to follow how these are defined, and then the tests also don't depend on the implementation details of the extension

test/cuda/tensors.jl
```diff
@@ -0,0 +1,609 @@
...
+                    @test norm(t3′) ≈ norm(t′)
+                    @test dot(t′, t) ≈ dot(t3′, t3)
+                end
+            end
+        end
+        if BraidingStyle(I) isa Bosonic && hasfusiontensor(I)
+            @timedtestset "Permutations: test via CPU" begin
+                W = V1 ⊗ V2 ⊗ V3 ⊗ V4 ⊗ V5
+                t = CUDA.rand(ComplexF64, W)
+                a = convert(Array, TensorKit.to_cpu(t))
```
lkdvos:
```suggestion
```

test/cuda/tensors.jl
```diff
@@ -0,0 +1,609 @@
...
+                    t3 = @constinferred repartition(t, $k)
+                    t3 = repartition(t, k)
+                    @test norm(t3) ≈ norm(t)
+                    t3′ = @constinferred repartition!(similar(t3), t′)
+                    @test norm(t3′) ≈ norm(t′)
+                    @test dot(t′, t) ≈ dot(t3′, t3)
+                end
+            end
+        end
+        if BraidingStyle(I) isa Bosonic && hasfusiontensor(I)
```
lkdvos:
```suggestion
        if BraidingStyle(I) isa SymmetricBraiding
```
Since we don't have to convert to array, we can test more things!

test/cuda/tensors.jl
```diff
@@ -0,0 +1,609 @@
...
+                @test t1 \ one(t1) ≈ inv(t1)
+                @test one(t1) / t1 ≈ pinv(t1)
+                @test_throws SpaceMismatch inv(t)
+                @test_throws SpaceMismatch t2 \ t
+                @test_throws SpaceMismatch t / t1
+                tp = pinv(t) * t
+                @test tp ≈ tp * tp
+            end
+        end
+        if BraidingStyle(I) isa Bosonic && hasfusiontensor(I)
```
lkdvos:
```suggestion
```
Here we can also test more things!

ext/TensorKitCUDAExt/cutensormap.jl
lkdvos: Sorry to be annoying about this, was this now confirmed to be required? (Also happy to revisit this later)

src/tensors/tensor.jl
```diff
@@ -315,11 +316,15 @@ end
...
+            $randf([rng=default_rng()], [TorA=Float64], codomain::ProductSpace{S,N₁},
                      domain::ProductSpace{S,N₂}) where {S,N₁,N₂,T} -> t
-        $randf([rng=default_rng()], [T=Float64], codomain ← domain) -> t
+            $randf([rng=default_rng()], [TorA=Float64], codomain ← domain) -> t
         
     Generate a tensor `t` with entries generated by `$randf`.
+        The type `TorA` can be used to control the element type and
+        data type generated. For example, if `TorA` is a `CuVector{ComplexF32}`
+        or `ROCVector{Float64}`, then the final output `TensorMap` will have
+        that as its storage type.
```
lkdvos: This looks like there might be some slight indenting issues?

Project.toml
```diff
@@ -48,21 +54,29 @@ TestExtras = "0.2,0.3"
...
 TestExtras = "5ed8adda-3752-4e41-b88a-e8b09835ee3a"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+cuTENSOR = "011b41b2-24ef-40a8-b3eb-fa098493e9e1"
 
 [targets]
-test = ["ArgParse", "Aqua", "Combinatorics", "LinearAlgebra", "TensorOperations", "Test", "TestExtras", "SafeTestsets", "ChainRulesCore", "ChainRulesTestUtils", "FiniteDifferences", "Zygote"]
+test = ["ArgParse", "Adapt", "Aqua", "Combinatorics", "CUDA", "cuTENSOR", "GPUArrays", "LinearAlgebra", "SafeTestsets", "TensorOperations", "Test", "TestExtras", "ChainRulesCore", "ChainRulesTestUtils", "FiniteDifferences", "Zygote"]
+
+[sources]
+MatrixAlgebraKit = {url="https://github.com/QuantumKitHub/MatrixAlgebraKit.jl", rev="main"}
```
lkdvos: TODO, remove once MAK tag is done (just a reminder)

lkdvos (dismissed):
Except for my remaining question about the convert, and the comment about the tagged MatrixAlgebraKit version, I'm happy to merge this.

As always, thanks for the enormous amount of work (and patience), much appreciated!

Project.toml
kshyatt: Done

src/tensors/tensor.jl
kshyatt: Was me fighting with the formatter

lkdvos (approved):
I think I'm also giving up on the formatter.
Manual calls to runic seem fine, it really is `git runic` that screws up the `"""` string indentation.
Luckily it only triggers on changed lines so we can just ignore this for now.
Otherwise this PR is good to go for me, if no-one disagrees I'll just merge this evening to keep things moving forwards.

PR #327, opened 2025-12-11
rework tensor contructors to allow storagetype specification

lkdvos:
This is a proposal for reworking the dispatch strategies of the `TensorMap` constructors to avoid having to special case GPU-backed tensors.

The main idea is that instead of using `TensorMap{T}` as an entry point, I want to use `(TensorMap{T, S, N1, N2, A} where {S, N1, N2})` as an entry point, such that it becomes easier to work with type aliases that specify the storage type.
I think I am imagining that at least for now we will have a very limited (finite) set of these, so it might be reasonable to imagine that these each have their own type alias, for which the generic functionality is defined in this PR.

Furthermore, I tried to leave a hook (`_tensormap_storagetype`) to map array types to `A`, to correctly infer the storage type for various inputs. This is definitely not an exhaustive way of doing this and we might have to rethink this strategy, but I do think we can at least make it work for a number of often-occurring cases.

Obviously comments welcome, names are tentative, and it would be nice to have some input from @kshyatt to see if this actually alleviates the need for the GPU constructors, as I did not test this. (I locally tested some of these to try `_TensorMap{T, Memory{T}}`, which did actually work 🎉 

While going over this, I did think of a different way of going from a `Dict{<:Sector, <:AbstractMatrix}` to a data vector, for which the main idea consists of leveraging `mapreduce(x -> reshape(x, :), vcat, data)` to simply let dispatch figure out what kind of vector comes out (up to making sure the order is correct of course), which might just work out of the gate for a large variety of input matrix types (and maybe also abstract array types?), although it would probably be a little less efficient. This is definitely also something worth exploring.

### DIFF
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -586,7 +586,7 @@ Base.similar(::Type{TT}, cod::TensorSpace, dom::TensorSpace) where {TT <: Abstra
 # The implementation is again written for similar_diagonal(t, TorA, V::ElementarySpace) -> DiagonalTensorMap
 # and all other methods are just filling in default arguments
 @doc """
-    similar_diagonal(t::AbstractTensorMap, [AorT=storagetype(t)], [V::ElementarySpace])
+    similar_diagonal(t::AbstractTensorMap, [AorT=scalartype(t)], [V::ElementarySpace])
 
 Creates an uninitialized mutable diagonal tensor with the given scalar or storagetype `AorT` and
 structure `V ← V`, based on the source tensormap. The second argument is optional and defaults
@@ -602,21 +602,12 @@ See also [`Base.similar`](@ref).
 
 # 3 arguments
 function similar_diagonal(t::AbstractTensorMap, ::Type{TorA}, V::ElementarySpace) where {TorA}
-    if TorA <: Number
-        T = TorA
-        A = similarstoragetype(t, T)
-    elseif TorA <: DenseVector
-        A = TorA
-        T = scalartype(A)
-    else
-        throw(ArgumentError("Type $TorA not supported for similar"))
-    end
-
-    return DiagonalTensorMap{T, spacetype(V), A}(undef, V)
+    A = similarstoragetype(TorA <: Number ? similarstoragetype(t, TorA) : TorA)
+    return DiagonalTensorMap{scalartype(A), spacetype(V), A}(undef, V)
 end
 
-similar_diagonal(t::AbstractTensorMap) = similar_diagonal(t, similarstoragetype(t), _diagspace(t))
-similar_diagonal(t::AbstractTensorMap, V::ElementarySpace) = similar_diagonal(t, similarstoragetype(t), V)
+similar_diagonal(t::AbstractTensorMap) = similar_diagonal(t, scalartype(t), _diagspace(t))
+similar_diagonal(t::AbstractTensorMap, V::ElementarySpace) = similar_diagonal(t, scalartype(t), V)
 similar_diagonal(t::AbstractTensorMap, T::Type) = similar_diagonal(t, T, _diagspace(t))
 
 function _diagspace(t)


src/tensors/tensor.jl
```diff
@@ -83,33 +95,35 @@ dim(t::TensorMap) = length(t.data)
...
-    ) where {T, S}
-    return TensorMap(data, codomain ← domain)
-end
+# here, we force the `T` and/or `A` argument to distinguish it from the more general constructor below
+TensorMap{T}(data::DenseVector{T}, V::TensorMapSpace) where {T} =
+    _TensorMap{T, typeof(data)}(data, V)
+TensorMap{T}(data::DenseVector{T}, codomain::TensorSpace, domain::TensorSpace) where {T} =
+    TensorMap{T}(data, codomain ← domain)
+
+_TensorMap{T, A}(data::DenseVector{T}, V::TensorMapSpace) where {T, A} =
```
kshyatt: Could `data` here have the type `A` and we move the `<:DenseVector{T}` to the `where`?

src/tensors/tensor.jl
```diff
@@ -317,49 +335,45 @@ cases.
...
-    if sectortype(S) === Trivial # refer to same method, but now with vector argument
-        return TensorMap(reshape(data, length(data)), V)
+    if sectortype(V) === Trivial # refer to same method, but now with vector argument
+        return _TensorMap{T, A}(reshape(data, length(data)), V)
     end
 
-    t = TensorMap{T}(undef, codom, dom)
+    t = _TensorMap{T, A}(undef, V)
     arraydata = reshape(collect(data), arraysize)
     t = project_symmetric!(t, arraydata)
```
kshyatt: I still think `project_symmetric!` may not play nicely with GPU arrays because of the reliance on `StridedView`, but perhaps we can deal with that separately

src/tensors/tensor.jl
lkdvos: I was somehow imagining this to also have a convert, i.e. calling `_TensorMap{T, CuVector{T}}(data::Vector{T}, ...)` might automatically convert to GPU? It's a bit iffy to know where to draw the line really, I guess you are probably right that it might be better to not have too much of this kind of functionality in our constructors and ask people to be explicit about it

src/tensors/tensor.jl
lkdvos: I also think this hits the issue again that we don't have `fusiontensor` on GPU, so it's probably fastest right now to just do this on the CPU? We never really need this in any high-performance parts of the code either I think

src/tensors/tensor.jl
lkdvos: So I decided to keep this in for now, since the default behavior of constructors with typed fields is to indeed call `convert` on the respective fields, which I think fits with this idea. Happy to rediscuss though

src/tensors/tensor.jl
lkdvos: will leave this issue as unresolved for now, and try not to deal with that in this PR.

lkdvos:
I think this is ready for review and shouldn't be breaking.

src/tensors/tensor.jl
```diff
@@ -71,87 +71,109 @@ dim(t::TensorMap) = length(t.data)
 
 # General TensorMap constructors
 #--------------------------------
-# undef constructors
+# hook for mapping input types to storage types -- to be implemented in extensions
+_tensormap_storagetype(::Type{A}) where {A <: AbstractArray} = _tensormap_storagetype(eltype(A))
```
Jutho: Could it be useful to have instead something like 
```julia
_tensormap_storagetype(::Type{A}) where {A <: AbstractArray} = _tensormap_storagetype(parent(A))
```

Not all `AbstractArray` types listen to `parent`, but the ones we care about typically do, and this would already help with overloading this method for `CuArray` storage and similar, where all the `view` and `adjoint` wrappers would immediately be taken care of.

lkdvos: I definitely like that, I also think that's just a better default than always returning `Vector{T}`

lkdvos: So I wanted to implement this, but we would have to rethink slightly how it works since `parent` doesn't work in the type domain...

lkdvos: It turns out I actually don't really have a great idea about how to do this, it is used in the dict of blocks case where I really just want to call it on `valtype`, and I could try and obtain a block and check that it is the correct type but that also doesn't look so elegant...

src/tensors/tensor.jl
```diff
@@ -70,89 +75,215 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
...
+# - dispatch starts with TensorMap{T}(::DenseVector{T}, ...)
+# - select A and map to (TensorMap{T, S, N₁, N₂, A} where {S, N₁, N₂})(::DenseVector{T}, ...)
+# - select S, N₁, N₂ and map to TensorMap{T, S, N₁, N₂, A}(::DenseVector{T}, ...)
+
+"""
+    TensorMap{T}(data::DenseVector{T}, codomain::ProductSpace{S, N₁}, domain::ProductSpace{S, N₂}) where {T, S, N₁, N₂}
+    TensorMap{T}(data::DenseVector{T}, codomain ← domain)
+    TensorMap{T}(data::DenseVector{T}, domain → codomain)
+
+Constrct a `TensorMap` from the given raw data.
```
kshyatt: Construct (typo)

src/tensors/tensor.jl
```diff
@@ -70,89 +75,215 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
...
+    _TensorMap{T, typeof(data)}(data, V)
+TensorMap{T}(data::DenseVector{T}, codomain::TensorSpace, domain::TensorSpace) where {T} =
+    TensorMap{T}(data, codomain ← domain)
+
+"""
+    (TensorMap{T, S, N₁, N₂, A} where {S, N₁, N₂})(data::DenseVector{T}, codomain, domain) where {T, A}
+    (TensorMap{T, S, N₁, N₂, A} where {S, N₁, N₂})(data::DenseVector{T}, codomain ← domain) where {T, A}
+    (TensorMap{T, S, N₁, N₂, A} where {S, N₁, N₂})(data::DenseVector{T}, domain → codomain) where {T, A}
+
+Constrct a `TensorMap` from the given raw data.
```
kshyatt: Here also

src/tensors/tensor.jl
```diff
@@ -70,89 +75,215 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
...
+    
+Here, `data` can be specified in three ways:
+1) `data` can be a `DenseVector` of length `dim(codomain ← domain)`; in that case it represents
+   the actual independent entries of the tensor map. An instance will be created that directly
+   references `data`.
+2) `data` can be an `AbstractMatrix` of size `(dim(codomain), dim(domain))`
+3) `data` can be an `AbstractArray` of rank `N₁ + N₂` with a size matching that of the domain
+   and codomain spaces, i.e. `size(data) == (dims(codomain)..., dims(domain)...)`
+
+In case 2 and 3, the `TensorMap` constructor will reconstruct the tensor data such that the
```
kshyatt: In cases (typo)

src/tensors/tensor.jl
Jutho: We can probably use the `_tensormap_storagetype` below to rewrite this `if ... elseif ... else` logic. This way, if we change the former, these two remain consistent.

src/tensors/tensor.jl
```diff
@@ -70,89 +75,215 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
...
+"""
+    (TensorMap{T, S, N₁, N₂, A} where {S, N₁, N₂})(data::DenseVector{T}, codomain, domain) where {T, A}
+    (TensorMap{T, S, N₁, N₂, A} where {S, N₁, N₂})(data::DenseVector{T}, codomain ← domain) where {T, A}
+    (TensorMap{T, S, N₁, N₂, A} where {S, N₁, N₂})(data::DenseVector{T}, domain → codomain) where {T, A}
+
+Construct a `TensorMap` from the given raw data.
+"""
+function _TensorMap{T, A}(data::DenseVector{T}, V::TensorMapSpace) where {T, A}
+    length(data) == dim(V) || throw(DimensionMismatch("invalid length of data"))
+    return TensorMap{T, spacetype(V), numout(V), numin(V), A}(data, V)
```
Jutho: I don't think this is captured anywhere if `typeof(data) != A`.

src/tensors/tensor.jl
```diff
@@ -70,89 +75,215 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
 dim(t::TensorMap) = length(t.data)
 
 # General TensorMap constructors
-#--------------------------------
+# ==============================
+
+# INTERNAL: utility type alias that makes constructors also work for type aliases that specify
+# different storage types. (i.e. CuTensorMap = _TensorMap{T, CuVector{T}, ...})
+const _TensorMap{T, A <: DenseVector{T}, S, N₁, N₂} = TensorMap{T, S, N₁, N₂, A}
+const _Tensor{T, A <: DenseVector{T}, S, N} = Tensor{T, S, N, A}
```
Jutho: How about

```suggestion
const TensorMapWithStorage{T, A <: DenseVector{T}, S, N₁, N₂} = TensorMap{T, S, N₁, N₂, A}
const TensorWithStorage{T, A <: DenseVector{T}, S, N} = Tensor{T, S, N, A}
```

to clarify (somewhat) what this is for (aside from the comment)?

src/tensors/tensor.jl
```diff
@@ -70,89 +78,216 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
...
+    ndims(data) == 1 && length(data) == dim(V) &&
+        return tensormaptype(spacetype(V), numout(V), numin(V), A)(data, V)
+
+    # special case trivial: refer to same method, but now with vector argument
+    sectortype(V) === Trivial &&
+        return tensormaptype(spacetype(V), numout(V), numin(V), A)(reshape(data, length(data)), V)
+
+    # do projection
+    t = TensorMapWithStorage{T, A}(undef, V)
+    t = project_symmetric!(t, data)
```
kshyatt, line 214: just a note that this will not work with GPU-backed arrays until we write a `project_symmetric` which supports them. For those arrays, we'll currently need to do this projection in CPU memory and then copy over, but this overall shouldn't impact performance too much.

lkdvos: I guess for now we should then specialize `project_symmetric!` in the extensions to go through CPU and go from there?

kshyatt: We might also have an annoying problem with converting a `CuTensorMap` to a `CuArray`. I might even suggest refactoring the entire " `project_symmetric!` and check " routine into a separate function? Then `CuTensorMap` (or `ROCTensorMap`) can handle copying back and forth to/from CPU. But that could be done in a separate PR

kshyatt: I think if that is the best approach it should be done in the CUDA support PR and not delay this any further as it's really pretty incidental

src/tensors/tensor.jl
```diff
@@ -70,89 +78,216 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
...
+
+"""
+    TensorMapWithStorage{T, A}(data::DenseVector{T}, codomain, domain) where {T, A}
+    TensorMapWithStorage{T, A}(data::DenseVector{T}, codomain ← domain) where {T, A}
+    TensorMapWithStorage{T, A}(data::DenseVector{T}, domain → codomain) where {T, A}
+
+Construct a `TensorMap` from the given raw data.
+This constructor takes ownership of the provided vector, and will not make an independent copy.
+"""
+function TensorMapWithStorage{T, A}(data::A, V::TensorMapSpace) where {T, A}
```
Jutho, line 135: Do we want an `A<:DenseVector{T}` restriction here (and below)?

lkdvos: I think that is already enforced by the type definition, i.e. you cannot even construct `TensorMapWithStorage{T, A}` where that isn't fulfilled, so I didn't really see the need to specify further?

src/tensors/tensor.jl
```diff
@@ -70,89 +67,215 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
...
-    return TensorMap{T, S, N₁, N₂, Vector{T}}(undef, V)
-end
-function TensorMap{T}(
-        ::UndefInitializer, codomain::TensorSpace{S}, domain::TensorSpace{S}
-    ) where {T, S}
-    return TensorMap{T}(undef, codomain ← domain)
-end
-function Tensor{T}(::UndefInitializer, V::TensorSpace{S}) where {T, S}
-    return TensorMap{T}(undef, V ← one(V))
+TensorMap{T}(::UndefInitializer, V::TensorMapSpace) where {T} =
```
Jutho, line 89: So this will allow something like
```julia
TensorMap{Vector{Float64}}(undef, V)
```

Is that intentional?

src/tensors/abstracttensor.jl
```diff
@@ -45,11 +45,57 @@ end
...
 """ storagetype
 
-similarstoragetype(TT::Type{<:AbstractTensorMap}) = similarstoragetype(TT, scalartype(TT))
+# storage type determination and promotion - hooks for specializing
+# the default implementation tries to leverarge inference and `similar`
+@doc """
+    similarstoragetype(t, [T = scalartype(t)]) -> Type{<:DenseVector{T}}
+    similarstoragetype(TT, [T = scalartype(t)]) -> Type{<:DenseVector{T}}
+    similarstoragetype(A, [T = scalartype(t)]) -> Type{<:DenseVector{T}}
+    similarstoragetype(D, [T = scalartype(t)]) -> Type{<:DenseVector{T}}
```
Jutho:
```suggestion
    similarstoragetype(t, [T = scalartype(t)]) -> Type{<:DenseVector{T}}
    similarstoragetype(TT, [T = scalartype(TT)]) -> Type{<:DenseVector{T}}
    similarstoragetype(A, [T = scalartype(A)]) -> Type{<:DenseVector{T}}
    similarstoragetype(D, [T = scalartype(D)]) -> Type{<:DenseVector{T}}
```

src/tensors/tensor.jl
lkdvos:
```julia
julia> using TensorKit

julia> V = ComplexSpace(2)
ℂ^2

julia> TensorMap{Vector{Float64}}(undef, V)
ERROR: TypeError: in AbstractTensorMap, in T, expected T<:Number, got Type{Vector{Float64}}
```

That's not actually true, you can't instantiate these types because of the type restrictions in the struct definition, so it will already error before you get to call that method.

src/tensors/abstracttensor.jl
```diff
@@ -494,54 +539,39 @@ specialize this method.
...
-function Base.similar(
-        t::AbstractTensorMap, codomain::TensorSpace{S}, domain::TensorSpace{S}
-    ) where {S}
-    return similar(t, similarstoragetype(t), codomain ← domain)
-end
-function Base.similar(t::AbstractTensorMap, ::Type{T}, codomain::TensorSpace) where {T}
-    return similar(t, T, codomain ← one(codomain))
-end
+Base.similar(t::AbstractTensorMap, codomain::TensorSpace, domain::TensorSpace) =
+    similar(t, similarstoragetype(t, scalartype(t)), codomain ← domain)
```
Jutho, line 557: Is there a reason this one has `similarstoragetype(t, scalartype(t))` as second argument (which would be equivalent to the shorter `similarstoragetype(t)`), but then the definitions below simply use `scalartype(t)` as second argument?

src/tensors/abstracttensor.jl
```diff
@@ -494,54 +539,39 @@ specialize this method.
...
-function Base.similar(
-        ::Type{TT}, cod::TensorSpace{S}, dom::TensorSpace{S}
-    ) where {TT <: AbstractTensorMap, S}
-    return TensorMap{scalartype(TT)}(undef, cod, dom)
+function Base.similar(::Type{TT}, V::TensorMapSpace) where {TT <: AbstractTensorMap}
+    TT′ = tensormaptype(spacetype(V), numout(V), numin(V), similarstoragetype(TT, scalartype(TT)))
+    return TT′(undef, V)
 end
+Base.similar(::Type{TT}, cod::TensorSpace, dom::TensorSpace) where {TT <: AbstractTensorMap} =
+    similar(TT, cod ← dom)
```
Jutho, line 582: Are there no type-domain `similar` methods with a `TorA` argument?

src/tensors/abstracttensor.jl
lkdvos: Yes, which might actually be a good reason to revert my last several commits:

For the constructors, we want to take the type of the provided raw vector, so there I need to have `similarstoragetype(data::DenseVector) = typeof(data)`. On the other hand, for the `similar` calls I need to take the type of what would be the result of `similar`, which doesn't always align. This is what was failing - the `PtrArray` data from the manual allocator in TensorOperations has `PtrArray{T,1} <: DenseVector`, but `similar(::PtrArray{T,1})::Vector{T}`.
In order to still force these two methods together, I therefore distinguish by the one and two-argument versions.

src/tensors/tensor.jl
Jutho: Ok, this seems like the second time I ask this question today? Or at least a very similar one.

Jutho: I am definitely confused by how this works from a compiler perspective. This means that `TensorMap{T}` methods are constrained by the type restrictions in the `TensorMap{T, ...}` type definition, even if that specific method definition would not end up calling and thereby creating an actual `TensorMap{T, ...}` instance.

I always assumed that additional constructor definitions were treated just the same as additional method definitions of a given function, and that they did not implicitly carry over type constraints.

src/tensors/tensor.jl
```diff
@@ -70,89 +67,215 @@ storagetype(::Type{<:TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A
...
+"""
+TensorMap{T}(data::DenseVector{T}, V::TensorMapSpace) where {T} =
+    tensormaptype(spacetype(V), numout(V), numin(V), typeof(data))(data, V)
+TensorMap{T}(data::DenseVector{T}, codomain::TensorSpace, domain::TensorSpace) where {T} =
+    TensorMap{T}(data, codomain ← domain)
+
+"""
+    TensorMapWithStorage{T, A}(data::DenseVector{T}, codomain, domain) where {T, A}
+    TensorMapWithStorage{T, A}(data::DenseVector{T}, codomain ← domain) where {T, A}
+    TensorMapWithStorage{T, A}(data::DenseVector{T}, domain → codomain) where {T, A}
```
Jutho: Even if the `A<:DenseVector{T}` restriction is not necessary in the function definition because it is enforced by the constraints in the type definition, it would be useful to have it here:
```suggestion
    TensorMapWithStorage{T, A}(data::A, codomain, domain) where {T, A<:DenseVector{T}}
    TensorMapWithStorage{T, A}(data::A, codomain ← domain) where {T, A<:DenseVector{T}}
    TensorMapWithStorage{T, A}(data::A, domain → codomain) where {T, A<:DenseVector{T}}
```

src/tensors/tensor.jl
lkdvos: I think the way I would think about this is that really you are first constructing the type itself, which is also a julia object, for which the type restrictions of the type definition are checked, and then you are just using that type as a callable and interpreting that as a constructor. In that regards, Julia really doesn't have constructors and this is just a convention that it should return a type that conforms to the callable.

src/auxiliary/auxiliary.jl
```diff
@@ -86,3 +86,6 @@ end
 else
     _allequal(f, xs) = allequal(f, xs)
 end
+
+Base.@assume_effects :foldable parenttype(::Type{T}) where {T} =
+    Core.Compiler.return_type(parent, Tuple{T})
```
Jutho: I think this is no longer used?

Jutho:
Ok, if all tests run and you take a look at my final comments, I think this is good for me!

src/tensors/abstracttensor.jl
lkdvos: There were none before, which is in line with `Base.similar` which also doesn't have a `similar` in the type domain with 3 arguments

PR #330, opened 2025-12-16
Improve `Diagonal` constructors with `similar_diagonal`

lkdvos:
This PR adds the (unexported for now?) `similar_diagonal` function which acts like `similar` but instead returns a `DiagonalTensorMap`.
The main point is to facilitate GPU support by correctly taking into account the storage types of the provided input tensor.

### DIFF
--- a/src/factorizations/truncation.jl
+++ b/src/factorizations/truncation.jl
@@ -132,7 +132,7 @@ for f! in (:eig_trunc!, :eigh_trunc!)
         ind = MAK.findtruncated(diagview(D), strategy)
         V_truncated = truncate_space(space(D, 1), ind)
 
-        D = similar_diagonal(D, V_truncated)
+        D̃ = similar_diagonal(D, V_truncated)
         truncate_diagonal!(D̃, D, ind)
 
         Ṽ = similar(V, codomain(V) ← V_truncated)


src/factorizations/diagonal.jl
```diff
@@ -72,7 +72,7 @@ function MAK.initialize_output(
     V_cod = fuse(codomain(t))
     V_dom = fuse(domain(t))
     U = similar(t, codomain(t) ← V_cod)
-    S = DiagonalTensorMap{real(scalartype(t))}(undef, V_cod ← V_dom)
+    S = similar_diagonal(t, real(scalartype(t)), V_cod ← V_dom)
```
kshyatt: 🎉 🎉 🎉

kshyatt (approved):
OK, I think this will really simplify things for the GPU support. Thanks very much for doing this!

src/factorizations/diagonal.jl
```diff
@@ -72,7 +72,7 @@ function MAK.initialize_output(
     V_cod = fuse(codomain(t))
     V_dom = fuse(domain(t))
     U = similar(t, codomain(t) ← V_cod)
-    S = DiagonalTensorMap{real(scalartype(t))}(undef, V_cod ← V_dom)
```
Jutho, line 75: I am confused by this whole method. In `svd_full!` we cannot use a `DiagonalTensorMap`, right? `V_cod` and `V_dom` can be different, which is not supported by `DiagonalTensorMap`.

Jutho: Oh wait, I hadn't seen the `DiagonalAlgorithm` part, never mind.

src/tensors/abstracttensor.jl
```diff
@@ -543,6 +545,70 @@ function Base.similar(
...
+# 3 arguments
+function similar_diagonal(t::AbstractTensorMap, ::Type{T}, V::TensorMapSpace) where {T}
+    numout(V) == numin(V) == 1 && domain(V) == codomain(V) ||
+        throw(ArgumentError("requested space is not square"))
+    return similar_diagonal(t, T, codomain(V))
+end
+function similar_diagonal(t::AbstractTensorMap, ::Type{T}, V::ProductSpace) where {T}
+    length(V) == 1 || throw(ArgumentError())
+    return similar_diagonal(t, T, only(V.spaces))
+end
```
Jutho: Personally, I wouldn't mind if all of these method definitions are gone, and it is just very clear from the onset that you can only construct `DiagonalTensorMap` for `V ← V`, and so you should only specify a single `V::ElementarySpace`, i.e. only the calls below are valid. We currently have the same approach for `DiagonalTensorMap`. If there is really a motivating argument to have these generalizations, we can consider them later. But if we have them now, it is harder to get rid of them and more code to maintain/process.

Most (but currently not all) of the call sites already use the single `V` method anyway.

src/tensors/abstracttensor.jl
```diff
@@ -543,6 +545,70 @@ function Base.similar(
...
+similar_diagonal(t::AbstractTensorMap, ::Type{T}) where {T} =
+    similar_diagonal(t, T, space(t))
+similar_diagonal(t::AbstractTensorMap, P::TensorMapSpace) =
+    similar_diagonal(t, similarstoragetype(t), P)
+similar_diagonal(t::AbstractTensorMap, P::TensorSpace) =
+    similar_diagonal(t, similarstoragetype(t), P)
+
+# 1 argument
+similar_diagonal(t::AbstractTensorMap) =
+    similar_diagonal(t, similarstoragetype(t), space(t))
```
Jutho:
```suggestion
similar_diagonal(t::AbstractTensorMap) = similar_diagonal(t, similarstoragetype(t), _diagspace(t))
similar_diagonal(t::AbstractTensorMap, V::ElementarySpace) = similar_diagonal(t, similarstoragetype(t), V)
similar_diagonal(t::AbstractTensorMap, T::Type) = similar_diagonal(t, T, _diagspace(t))

function _diagspace(t)
    cod, dom = codomain(t), domain(t)
    length(cod) == 1 && cod == dom ||
        throw(ArgumentError("space does not support a diagonal TensorMap"))
    return only(cod)
end
```

src/tensors/sectorvector.jl
```diff
@@ -16,6 +16,11 @@ function SectorVector{T}(::UndefInitializer, V::ElementarySpace) where {T}
     structure = diagonalblockstructure(V ← V)
     return SectorVector(data, structure)
 end
+function SectorVector{T, I, A}(::UndefInitializer, V::ElementarySpace) where {T, I, A <: AbstractVector{T}}
+    data = A(undef, reduceddim(V))
```
Jutho, line 20: I know we discussed whether `A <: DenseVector` was desirable in the `SectorVector` type definition. For this constructor, I think that if not `A<:DenseVector`, it is unlikely that `A(undef, reduceddim(V))` is going to be a valid constructor call, but I am fine leaving this as is.

lkdvos: Yeah, this might indeed need further specialization, but I also agree that it probably doesn't hurt

src/tensors/abstracttensor.jl
```diff
@@ -543,6 +545,51 @@ function Base.similar(
...
+end
+
+similar_diagonal(t::AbstractTensorMap) = similar_diagonal(t, similarstoragetype(t), _diagspace(t))
+similar_diagonal(t::AbstractTensorMap, V::ElementarySpace) = similar_diagonal(t, similarstoragetype(t), V)
+similar_diagonal(t::AbstractTensorMap, T::Type) = similar_diagonal(t, T, _diagspace(t))
+
+function _diagspace(t)
+    cod, dom = codomain(t), domain(t)
+    length(cod) == 1 && cod == dom ||
+        throw(ArgumentError("space does not support a diagonal TensorMap"))
```
Jutho: My apologies for changes to my own code suggestions
```suggestion
        throw(ArgumentError("space does not support a DiagonalTensorMap"))
```

Jutho (approved):

PR #333, opened 2025-12-19
`svd_vals(::DiagonalTensorMap)` should return a `SectorVector`

lkdvos:
Noticed some test failures in #325 due to the diagonal algorithms not returning a `SectorVector`, so I fixed that, added some tests and then found some other inconsistencies: `svd_vals` was sorting the entire `SectorVector`, effectively mixing up the blocks.

Both of these are fixed here.

### DIFF
--- a/docs/src/Changelog.md
+++ b/docs/src/Changelog.md
@@ -24,7 +24,6 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 - Extended support for selecting storage types in the `TensorMap` constructors ([#327](https://github.com/QuantumKitHub/TensorKit.jl/pull/327))
 - `similar_diagonal` to handle storage types when constructing diagonals ([#330](https://github.com/QuantumKitHub/TensorKit.jl/pull/330))
-- `LinearAlgebra.svd` overloads
 
 ### Fixed
 


src/factorizations/diagonal.jl
```diff
@@ -95,15 +96,18 @@ end
 
 # f_vals
 # ------
-for f! in (:eig_vals!, :eigh_vals!, :svd_vals!)
+# this is invalid for svd_vals since we cannot sort between blocks
+for f! in (:eig_vals!, :eigh_vals!)
```
Jutho: This might be a bit risky for `eig(h)` as well. For example, `eigh_full` returns a sorted output of the eigenvalues in the LAPACK/dense case. So we could in principle also implement this for the `DiagonalAlgorithm`, but then this will also invalidate this implementation. So maybe  a block per block approach is the safest, even if there is a little bit extra overhead.

lkdvos: I don't think that's entirely true here though, we aren't calling LAPACK since these inputs are diagonal, and at least for the eigenvalue decompositions we're somewhat protected by the fact that in order for V to be output as a DiagonalTensorMap, you can't reorder the values.
For the SVD specifically we don't have that, since we have to sort the singular values anyways so U and Vh aren't actually DiagonalTensorMaps. 

In any case, I don't think this overhead ever really matters though, so I'd be happy to remove this if you prefer

PR #335, opened 2026-01-01
`transpose` specialization for `DiagonalTensorMap`

lkdvos:
This is a partial fix for an issue mentioned in #243, specifically for `transpose`.
The main point is to preserve the `DiagonalTensorMap` type whenever possible, rather than instantiating a dense tensor.


Jutho (dismissed):

Jutho:
Oh but it is just the newly added tests that are wrong. Permutation and transposition are not compatible in this case, though I agree that this might be very confusing.

lkdvos:
Will fix, I forgot about that subtle difference :)

Jutho (approved):

PR #336, opened 2026-01-02
Initial support for CUDA + factorizations

kshyatt:
OK, now only truncation is not working, which is expected because it is also a WIP in MatrixAlgebraKit!

### DIFF
--- a/ext/TensorKitCUDAExt/cutensormap.jl
+++ b/ext/TensorKitCUDAExt/cutensormap.jl
@@ -119,15 +119,11 @@ function LinearAlgebra.isposdef(t::CuTensorMap)
     InnerProductStyle(spacetype(t)) === EuclideanInnerProduct() || return false
     for (c, b) in blocks(t)
         # do our own hermitian check
-        isherm = MatrixAlgebraKit.ishermitian(b; atol = eps(real(eltype(b))), rtol = eps(real(eltype(b))))
+        isherm = MatrixAlgebraKit.ishermitian(b)
         isherm || return false
-        isposdef(project_hermitian!(b)) || return false
+        isposdef(Hermitian(b)) || return false
     end
     return true
-end
-
-function LinearAlgebra.isposdef(t::TensorKit.AdjointTensorMap{T, S, N₁, N₂, <:CuTensorMap}) where {T, S, N₁, N₂}
-    return isposdef(adjoint(t))
 end
 
 function Base.promote_rule(
--- a/src/tensors/linalg.jl
+++ b/src/tensors/linalg.jl
@@ -196,10 +196,6 @@ LinearAlgebra.isdiag(t::AbstractTensorMap) = all(LinearAlgebra.isdiag ∘ last, 
 
 # In-place methods
 #------------------
-# Wrapping the blocks in a StridedView enables multithreading if JULIA_NUM_THREADS > 1
-# but can cause problems with underlying array types (like CuArray) that don't yet play
-# nicely with StridedViews
-# TODO: reconsider this strategy, consider spawning different threads for different blocks
 
 # Copy, adjoint and fill:
 function Base.copy!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap)
--- a/src/factorizations/adjoint.jl
+++ b/src/factorizations/adjoint.jl
@@ -111,3 +111,7 @@ function MAK.svd_compact!(t::AdjointTensorMap, F, alg::DiagonalAlgorithm)
     F′ = svd_compact!(adjoint(t), reverse(adjoint.(F)), _adjoint(alg))
     return reverse(adjoint.(F′))
 end
+
+function LinearAlgebra.isposdef(t::AdjointTensorMap)
+    return isposdef(adjoint(t))
+end


ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -164,3 +164,11 @@ for f in (:sqrt, :log, :asin, :acos, :acosh, :atanh, :acoth)
         return tf
     end
 end
+
+function Base.copy!(tdst::CuTensorMap, tsrc::AdjointTensorMap)
```
lkdvos: I think we should really just replace the `AbstractTensorMap` implementation with this one, and bypass the `StridedView` attempt at using multithreading within the blocks. (https://github.com/QuantumKitHub/TensorKit.jl/blob/78a070f65d2fb42f9b4618dbdf91fa1a38d7b612/src/tensors/linalg.jl#L203)

I'd say it feels a little wrong that we have a fallback definition for `AbstractTensorMap` that assumes we can convert to `StridedView`, and then we have to specialize different cases to avoid taking that codepath to begin with 😁

test/cuda/factorizations.jl
```diff
@@ -157,17 +157,17 @@ for V in spacelist
...
+                        CUDA.rand(T, V4, W)',
                         DiagonalTensorMap(CUDA.rand(T, reduceddim(V1)), V1),
                     )
 
                 @assert domain(t) ≾ codomain(t)
                 w, p = @constinferred left_polar(t)
                 @test w * p ≈ t
                 @test isisometric(w)
-                @test isposdef(p)
+                @test isposdef(project_hermitian!(p))
```
lkdvos: Do we need the projection here, and if so, why? Shouldn't the result already be hermitian?

test/cuda/factorizations.jl
```diff
@@ -0,0 +1,499 @@
...
+                @test cond(t2) == 0.0
+            end
+            for T in eltypes, t in (
+                        CUDA.rand(T, W, W),
+                        CUDA.rand(T, W, W)',
+                    )
+                project_hermitian!(t)
+                vals = @constinferred LinearAlgebra.eigvals(t)
+                λmax = maximum(s -> maximum(abs, s), values(vals))
+                λmin = minimum(s -> minimum(abs, s), values(vals))
```
lkdvos:
```suggestion
                λmax = maximum(abs, parent(vals))
                λmin = minimum(abs, parent(vals))
```

lkdvos (commented):
Overall definitely looks good to me!
I'm slightly worried about the code duplication for the tests, similar to how we have that for MatrixAlgebraKit, but I also don't think there are better alternatives so I'm ok with merging like this, except for some small comments.

test/cuda/factorizations.jl
kshyatt: Currently in the extension we're checking the input is approximately hermitian then wrapping it in a `Hermitian` it if so for `isposdef`. I'll try removing the projection here (or maybe we need to project out the extraneous non posdef terms?).

lkdvos: But positive definite requires hermitian anyways right? Testing if the result is positive definite should also mean testing it is hermitian, which is why I'm confused we have to project it to be hermitian, rather than checking for that property.

ext/TensorKitCUDAExt/cutensormap.jl
Jutho: I agree.

test/cuda/factorizations.jl
```diff
@@ -0,0 +1,499 @@
...
+            (Vtr, Vℤ₃, VfU₁, VfSU₂, VIB_M)
+        else
+            (Vtr, VU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
+        end
+    else
+        (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
+    end
+catch
+    (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
+end
```
Jutho: I think I proposed/introduced this logic at some point, but it seems like this should simply be replaced by


```suggestion
spacelist = if get(ENV, "CI", "false") == "true"
    println("Detected running on CI")
    if Sys.iswindows()
        (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VIB_diag)
    elseif Sys.isapple()
        (Vtr, Vℤ₃, VfU₁, VfSU₂, VIB_M)
    else
        (Vtr, VU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
    end
else
    (Vtr, Vℤ₃, VU₁, VfU₁, VCU₁, VSU₂, VfSU₂, VIB_diag, VIB_M)
end
```

test/cuda/factorizations.jl
Jutho: I would also think that our `left_polar` implementations output a `p` factor which is exactly hermitian. If this is not the case, this is something to add to the TODO list of MAK.

kshyatt: I think there is also now an actual problem with the test for `AdjointTensorMap` (related to copying?) which I'll dig into further today

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -121,11 +121,15 @@ function LinearAlgebra.isposdef(t::CuTensorMap)
         # do our own hermitian check
         isherm = MatrixAlgebraKit.ishermitian(b; atol = eps(real(eltype(b))), rtol = eps(real(eltype(b))))
         isherm || return false
-        isposdef(Hermitian(b)) || return false
+        isposdef(project_hermitian!(b)) || return false
     end
     return true
 end
```
Jutho, line 127: Still confused by this. The CPU version doesn't have this approximate hermiticity check and then hermitian projection. It just fails directly if it is not exactly hermitian. Why does the "generic" implementation in factorizations.jl fail for CUDA? Does it not support `isposdef!`?

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -121,11 +121,15 @@ function LinearAlgebra.isposdef(t::CuTensorMap)
...
 
+function LinearAlgebra.isposdef(t::TensorKit.AdjointTensorMap{T, S, N₁, N₂, <:CuTensorMap}) where {T, S, N₁, N₂}
+    return isposdef(adjoint(t))
+end
```
Jutho: If this is necessary (depending on previous comment), it could probably become a general method, over at "src/factorizations/factorizations.jl"


```suggestion
LinearAlgebra.isposdef(t::TensorKit.AdjointTensorMap) = isposdef(adjoint(t))
```

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: `isposdef!` works BUT I think `CUSOLVER` is very brittle about what counts as being `posdef` (see https://github.com/JuliaGPU/CUDA.jl/blob/0c00b83fe9df007efcb3e738fa5f1973603f79c4/lib/cusolver/dense.jl#L905)

kshyatt: So for example if you have an eigenvalue `-1e-17` that may return "not positive definite"

lkdvos: I think that we should really try and follow the CPU conventions for this, so indeed maybe this should just really check for exact hermitian, and go from there.
If this gives issues for the GPU projections, we might want to alter these implementations to put a `project_hermitian` at the end of these, rather than try and alter the checks?

src/tensors/linalg.jl
```diff
@@ -197,13 +197,15 @@ LinearAlgebra.isdiag(t::AbstractTensorMap) = all(LinearAlgebra.isdiag ∘ last,
 # In-place methods
 #------------------
 # Wrapping the blocks in a StridedView enables multithreading if JULIA_NUM_THREADS > 1
+# but can cause problems with underlying array types (like CuArray) that don't yet play
+# nicely with StridedViews
 # TODO: reconsider this strategy, consider spawning different threads for different blocks
```
lkdvos: We can probably just delete this comment + TODO, since we reconsidered the strategy and decided against wrapping in StridedView?

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: OK, so *don't* project_hermitian here, and make the check above exact?

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Or `src/factorizations/adjoint.jl`?

ext/TensorKitCUDAExt/cutensormap.jl
Jutho: The exact `ishermitian` check is already part of `isposdef`; I think if the MAK implementation of `PolarViaSVD` is fixed (see Zulip), this standard `isposdef` check should work without need for a CUDA specialisation.

PR #342, opened 2026-01-05
change blocktype of TensorMap to `StridedView`

lkdvos:
Will leave this here as a WIP for now, since some of the matrix functions still require `<:Base.StridedMatrix` (e.g. `exp!`, ...) to work.
This will probably be solved once we have more support in MatrixAlgebraKit, but since this is not urgent I'll just leave this as-is for now.

### DIFF
--- a/src/tensors/tensor.jl
+++ b/src/tensors/tensor.jl
@@ -458,28 +458,35 @@ blocks(t::TensorMap) = BlockIterator(t, fusionblockstructure(t).blockstructure)
 function blocktype(::Type{TT}) where {TT <: TensorMap}
     A = storagetype(TT)
     T = eltype(A)
-    return Base.ReshapedArray{T, 2, SubArray{T, 1, A, Tuple{UnitRange{Int}}, true}, Tuple{}}
+    @static if isdefined(Core, :Memory) # StridedViews normalizes parent types!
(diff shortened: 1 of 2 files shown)


src/tensors/tensor.jl
```diff
@@ -455,29 +455,38 @@ block(t::TensorMap, c::Sector) = blocks(t)[c]
...
 blocks(t::TensorMap) = BlockIterator(t, fusionblockstructure(t).blockstructure)
 
-function blocktype(::Type{TensorMap{T, S, N₁, N₂, A}}) where {T, S, N₁, N₂, A <: Vector{T}}
-    return Base.ReshapedArray{T, 2, SubArray{T, 1, A, Tuple{UnitRange{Int}}, true}, Tuple{}}
+function blocktype(::Type{TT}) where {TT <: TensorMap}
+    A = storagetype(TT)
+    T = eltype(A)
+    @static if isdefined(Core, :Memory) # StridedViews normalizes parent types!
+        if A <: Vector{T}
+            A = GenericMemory{T}
```
kshyatt, line 463: Shouldn't this just be a `Memory{T}` (or have the [`kind` `Symbol`](https://docs.julialang.org/en/v1/base/arrays/#Core.GenericMemory))

PR #344, opened 2026-01-05
Adapt extension

lkdvos:


test/cuda/tensors.jl
```diff
@@ -115,6 +115,18 @@ for V in spacelist
                 @test domain(t2) == one(W)
             end
         end
+        @timedtestset "Adapt" begin
+            W = V1 ⊗ V2 ⊗ V3 ⊗ V4 ⊗ V5
+            t = rand(Float64, W)
+            t_gpu = @constinferred adapt(CuArray, t)
+            @test storagetype(t_gpu) <: CuArray
```
kshyatt: Can we make this a bit tighter and require the element type too?

kshyatt: I know it's covered below also

PR #345, opened 2026-01-06
Some more documentation progress

Jutho:
This includes a major restructuring (breaking down long pages into separate pages) and a complete rewrite of the `Sector` documentation. This could become the manual page for TensorKitSectors.jl in time, but for now I would still include it here.

By lack of a good Markdown formatter, I switched to breaking lines at natural places (end of sentence, subsentences, …), like one might do with LaTeX documents. Maybe that strategy can lead to clean `git diff`s in future (potentially smaller) doc changes, but I am open to other suggestions.

This would very much benefit from some proofreading by e.g. @borisdevos , @leburgel , @lkdvos .

### DIFF
--- a/docs/src/man/tensors.md
+++ b/docs/src/man/tensors.md
@@ -10,7 +10,7 @@ As this is probably the most important part of the manual, we will also focus mo
 The only aspect of the implementation that we will address is the storage of the tensor data, as this is important to know how to create and initialize a tensor, but will in fact also shed light on how some of the methods work.
 
 As mentioned, all tensors in TensorKit.jl are interpreted as linear maps (morphisms) from a domain (a `ProductSpace{S, N₂}`) to a codomain (another `ProductSpace{S, N₁}`), with the same `S <: ElementarySpace` that labels the type of spaces associated with the individual tensor indices.
-The overall type for all such tensor maps is `AbstractTensorMap{S, N₁, N₂}`.
+The overall type for all such tensor maps is `AbstractTensorMap{T, S, N₁, N₂}`.
 Note that we place information about the codomain before that of the domain.
 Indeed, we have already encountered the constructor for the concrete parametric type `TensorMap` in the form `TensorMap(..., codomain, domain)`.
 This convention is opposite to the mathematical notation, e.g. ``\mathrm{Hom}(W, V)`` or ``f : W → V``, but originates from the fact that a normal matrix is also denoted as having size `m × n` or is constructed in Julia as `Array(..., (m, n))`, where the first integer `m` refers to the codomain being `m`- dimensional, and the seond integer `n` to the domain being `n`-dimensional.
@@ -379,9 +379,9 @@ Furthermore, note that they render the tensor block diagonal, but that now every
 
 ## [Tensor properties](@id ss_tensor_properties)
 
-Given a `t::AbstractTensorMap{S, N₁, N₂}`, there are various methods to query its properties.
+Given a `t::AbstractTensorMap{T, S, N₁, N₂}`, there are various methods to query its properties.
 The most important are clearly `codomain(t)` and `domain(t)`.
-For `t::AbstractTensor{S, N}`, i.e. `t::AbstractTensorMap{S, N, 0}`, we can use `space(t)` as synonym for `codomain(t)`.
+For `t::AbstractTensor{S, N}`, i.e. `t::AbstractTensorMap{T, S, N, 0}`, we can use `space(t)` as synonym for `codomain(t)`.
 However, for a general `AbstractTensorMap` this has no meaning.
 However, we can query `space(t, i)`, the space associated with the `i`th index.
 For `i ∈ 1:N₁`, this corresponds to `codomain(t, i) = codomain(t)[i]`.
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -346,10 +346,10 @@ function planartrace!(
     return planartrace!(C, TensorMap(A), p, q, α, β, backend, allocator)
 end
 
-# function planarcontract!(C::AbstractTensorMap{S,N₁,N₂},
+# function planarcontract!(C::AbstractTensorMap{<:Any,S,N₁,N₂},
 #                          A::BraidingTensor{S},
 #                          (oindA, cindA)::Index2Tuple{0,4},
-#                          B::AbstractTensorMap{S},
+#                          B::AbstractTensorMap{<:Any,S},
 #                          (cindB, oindB)::Index2Tuple{4,<:Any},
 #                          (p1, p2)::Index2Tuple{N₁,N₂},
 #                          α::Number, β::Number,
@@ -414,8 +414,8 @@ end
 #     end
 #     return C
 # end
-# function planarcontract!(C::AbstractTensorMap{S,N₁,N₂},
-#                          A::AbstractTensorMap{S},
+# function planarcontract!(C::AbstractTensorMap{<:Any,S,N₁,N₂},
+#                          A::AbstractTensorMap{<:Any,S},
 #                          (oindA, cindA)::Index2Tuple{0,4},
 #                          B::BraidingTensor{S},
 #                          (cindB, oindB)::Index2Tuple{4,<:Any},
@@ -482,10 +482,10 @@ end
 #     end
 #     return C
 # end
-# function planarcontract!(C::AbstractTensorMap{S,N₁,N₂},
+# function planarcontract!(C::AbstractTensorMap{<:Any,S,N₁,N₂},
 #                          A::BraidingTensor{S},
 #                          (oindA, cindA)::Index2Tuple{1,3},
-#                          B::AbstractTensorMap{S},
+#                          B::AbstractTensorMap{<:Any,S},
 #                          (cindB, oindB)::Index2Tuple{1,<:Any},
 #                          (p1, p2)::Index2Tuple{N₁,N₂},
 #                          α::Number, β::Number,
@@ -544,8 +544,8 @@ end
 #     end
 #     return C
 # end
-# function planarcontract!(C::AbstractTensorMap{S,N₁,N₂},
-#                          A::AbstractTensorMap{S},
+# function planarcontract!(C::AbstractTensorMap{<:Any,S,N₁,N₂},
+#                          A::AbstractTensorMap{<:Any,S},
 #                          (oindA, cindA)::Index2Tuple{<:Any,3},
 #                          B::BraidingTensor{S},
 #                          (cindB, oindB)::Index2Tuple{3,1},
--- a/src/tensors/indexmanipulations.jl
+++ b/src/tensors/indexmanipulations.jl
@@ -219,14 +219,15 @@ function LinearAlgebra.transpose(
 end
 
 """
-    repartition!(tdst::AbstractTensorMap{S}, tsrc::AbstractTensorMap{S}) where {S} -> tdst
+    repartition!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap) -> tdst
 
 Write into `tdst` the result of repartitioning the indices of `tsrc`. This is just a special
 case of a transposition that only changes the number of in- and outgoing indices.
 
 See [`repartition`](@ref) for creating a new tensor.
 """
-@propagate_inbounds function repartition!(tdst::AbstractTensorMap{S}, tsrc::AbstractTensorMap{S}) where {S}
+@propagate_inbounds function repartition!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap)
+    check_spacetype(tdst, tsrc)
     numind(tsrc) == numind(tdst) ||
         throw(ArgumentError("tsrc and tdst should have an equal amount of indices"))
     all_inds = (codomainind(tsrc)..., reverse(domainind(tsrc))...)
@@ -236,12 +237,12 @@ See [`repartition`](@ref) for creating a new tensor.
 end
 
 """
-    repartition(tsrc::AbstractTensorMap{S}, N₁::Int, N₂::Int; copy::Bool=false) where {S}
-        -> tdst::AbstractTensorMap{S,N₁,N₂}
+    repartition(
+        tsrc::AbstractTensorMap{T, S}, N₁::Int, N₂::Int; copy::Bool=false
+    ) where {T, S} -> tdst::AbstractTensorMap{T, S, N₁, N₂}
 
 Return tensor `tdst` obtained by repartitioning the indices of `t`.
-The codomain and domain of `tdst` correspond to the first `N₁` and last `N₂` spaces of `t`,
-respectively.
+The codomain and domain of `tdst` correspond to the first `N₁` and last `N₂` spaces of `t`, respectively.
 
 If `copy=false`, `tdst` might share data with `tsrc` whenever possible. Otherwise, a copy is always made.
 


docs/src/appendix/categories.md
```diff
@@ -283,8 +281,10 @@ In particular, one could choose ``\tilde{ϵ}_{{}^{∨}V} = ϵ_V`` and thus defin
 right dual of ``{}^{∨}V``. While there might be other choices, this choice must at least be
 isomorphic, such that ``({}^{∨}V)^{∨} ≂ V``.
 
-If objects ``V`` and ``W`` have left (respectively right) duals, than for a morphism ``f ∈ \mathrm{Hom}(W,V)``, we furthermore define the left (respectively right)
-*transpose* ``{}^{∨}f ∈ \mathrm{Hom}({}^{∨}V, {}^{∨}W)`` (respectively  ``f^{∨} ∈ \mathrm{Hom}(V^{∨}, W^{∨})``) as
+If objects ``V`` and ``W`` have left (respectively right) duals, than for a morphism
```
borisdevos:
```suggestion
If objects ``V`` and ``W`` have left (respectively right) duals, then for a morphism
```

docs/src/appendix/categories.md
```diff
@@ -366,17 +366,16 @@ object ``V``, referred to as `dim(V)` in TensorKit.jl.
 
 For further information and a more detailed treatment of rigid and pivotal categories, we
 refer to [^turaev] and [^selinger]. We conclude this section by studying the example of
```
borisdevos: Just an idea, but the way the citations are printed in the docs is precisely how you define it here, so e.g. [turaev] and [selinger]. Do we want to capitalise the names?

docs/src/appendix/categories.md
```diff
@@ -366,17 +366,16 @@ object ``V``, referred to as `dim(V)` in TensorKit.jl.
...
-``\mathbf{SVect}``. Let us, in every super vector space ``V``, define a basis ``|n⟩``
-that is compatible with the grading, such ``|n|=0,1`` indicates that ``|n⟩ ∈ V_{|n|}``.
-We again define a dual basis ``{⟨m|}`` for ``V^*`` (such that ``⟨m|n⟩ = δ_{m,n}``), and
-then define the left evaluation by
-``⁠ϵ_V:V^* ⊗ V → ℂ: ⟨m| ⊗_\mathrm{g} |n⟩ → ⟨m|n⟩ = δ_{m,n}`` and the left coevaluation by
-``η_V:ℂ→ V ⊗ V^*:α → α ∑_n |n⟩ ⊗_\mathrm{g} ⟨n|``. Note that this does not require an inner
-product and satisfies the snake rules. For the right evaluation and coevaluation, there are
-two natural choices, namely
+``\mathbf{SVect}``. Let us, in every super vector space ``V``, define a basis ``|n⟩`` that
+is compatible with the grading, such ``|n|=0,1`` indicates that ``|n⟩ ∈ V_{|n|}``. We again
```
borisdevos:
```suggestion
is compatible with the grading. The value ``|n|=0,1`` indicates that ``|n⟩ ∈ V_{|n|}``. We again
```
or something along these lines. I'm actually not too sure what the message is here, but the punctuation confused me anyway.

docs/src/appendix/categories.md
```diff
@@ -425,25 +424,26 @@ representation, it means that there is no distinction between over- and under- c
...
 
 ```@raw html
 <img src="../img/diagram-braidingdual.svg" alt="braiding dual" class="color-invertible"/>
 ```
 
 **Balanced categories** ``C`` are braided categories that come with a **twist** ``θ``, a
 natural transformation from the identity functor ``1_C`` to itself, such that
-``θ_V ∘ f = f ∘ θ_W`` for all morphisms ``f ∈ \mathrm{Hom}(W,V)``, and for which main requirement is
-that
+``θ_V ∘ f = f ∘ θ_W`` for all morphisms ``f ∈ \mathrm{Hom}(W,V)``, and for which main
```
borisdevos:
```suggestion
``θ_V ∘ f = f ∘ θ_W`` for all morphisms ``f ∈ \mathrm{Hom}(W,V)``, and for which the main
```

docs/src/appendix/categories.md
```diff
@@ -513,30 +513,30 @@ a compact closed category, where the resulting twist is given by
...
+together with an involutive functor ``†:C→C^{\mathrm{op}}``, i.e. it acts as the identity on
+objects, whereas on morphisms ``f:W→V`` it defines a morphism ``f^†:V→W`` such that
 * ``\mathrm{id}_V^† = \mathrm{id}_V``
 * ``(f ∘ g)^† = f^† ∘^{\mathrm{op}} g^† = g^† ∘ f^†``
-* ``(f^†)^† = f``
-Sometimes also the symbol ``*`` is used instead of ``†``, however we have already used
-``*`` to denote dual objects and transposed morphisms in the case of a pivotal category.
+* ``(f^†)^† = f`` Sometimes also the symbol ``*`` is used instead of ``†``, however we have
+  already used ``*`` to denote dual objects and transposed morphisms in the case of a
+  pivotal category.
```
borisdevos:
```suggestion
* ``(f^†)^† = f``. Sometimes also the symbol ``*`` is used instead of ``†``. However, we have
  already used ``*`` to denote dual objects and transposed morphisms in the case of a
  pivotal category.
```

docs/src/appendix/categories.md
```diff
@@ -581,28 +581,28 @@ discuss for the category ``\mathbf{Vect}``, but which can also be realized in ot
...
 morphisms from any ``𝕜``-linear monoidal category, but assumes categories with duals to be
 pivotal and in fact spherical, and categories with a braiding to be ribbon categories. A
-dagger ribbon category where the braiding is symmetric, i.e. a dagger category which is
-also a compact closed category and where the right (co)evaluation is given via the dagger
-of the left (co)evaluation is called a **dagger compact** category. This is the playground
-of quantum mechanics of bosonic and fermionic systems. However, we also allow for non-
+dagger ribbon category where the braiding is symmetric, i.e. a dagger category which is also
+a compact closed category and where the right (co)evaluation is given via the dagger of the
+left (co)evaluation is called a **dagger compact** category. This is the playground of
+quantum mechanics of bosonic and fermionic systems. However, we also allow for non-
```
borisdevos:
```suggestion
dagger ribbon category where the braiding is symmetric, i.e. a dagger category which is also
a compact closed category and where the right (co)evaluation is given via the dagger of the
left (co)evaluation, is called a **dagger compact** category. This is the playground of
quantum mechanics of bosonic and fermionic systems. However, we also allow for non-
```

docs/src/appendix/categories.md
```diff
@@ -659,30 +659,29 @@ be decomposed as
...
+rigid, and that only has a finite number of isomorphism classes of simple objects. Note that
+the duality functor maps ``\mathrm{End}(V)`` to ``\mathrm{End}(V^*)``, such that, if ``V``
+is a simple object, so must be ``V^*``. Henceforth, we will be sloppy about the distinction
+between a pre-fusion or fusion category, only use the latter term, even when it is not fully
+justified.
+
+Before continuing, let us use some examples to sketch the relevance of the concept of fusion
+categories. As mentioned, the categories ``\mathbf{Vect}_𝕜`` and ``\mathbf{SVect}_𝕜`` have
+``I ≂ 𝕜`` as simple object. For ``\mathbf{Vect}``, this is the only simple object, i.e. any
+other vector space ``V`` over ``𝕜``, can be thought of as a direct sum over
```
borisdevos:
```suggestion
other vector space ``V`` over ``𝕜`` can be thought of as a direct sum over
```

docs/src/appendix/categories.md
```diff
@@ -779,35 +778,34 @@ be rewritten in terms of these matrix elements, and as such yields the celebrate
...
-object ``a^*`` of a simple object ``a`` is simple, and thus, it must be isomorphic to one
-of the representives ``\bar{a}`` of the different isomorphism classes of simple objects
-that we have chosen. Note that it can happen that ``\bar{a}=a``. Duality implies an
-isomorphism between ``\mathrm{Hom}(W,V)`` and ``\mathrm{Hom}(I,V⊗W^*)``, and thus, for a
-simple object ``a``, ``\mathrm{End}(a) ≂ 𝕜`` is isomorphic to ``\mathrm{Hom}(1,a⊗a^*)``,
-such that the latter is also isomorphic to ``𝕜``, or thus ``N^{a\bar{a}}_1 = 1``. Also,
-all possible duals of ``a`` must be isomorphic, and thus there is a single representive
-``\bar{a}``, meaning that ``N^{ab}_1 = δ^{b,\bar{a}}``, i.e. for all other ``b ≠ \bar{a}``,
+object ``a^*`` of a simple object ``a`` is simple, and thus, it must be isomorphic to one of
+the representives ``\bar{a}`` of the different isomorphism classes of simple objects that we
```
borisdevos:
```suggestion
the representatives ``\bar{a}`` of the different isomorphism classes of simple objects that we
```

docs/src/appendix/categories.md
```diff
@@ -779,35 +778,34 @@ be rewritten in terms of these matrix elements, and as such yields the celebrate
...
+have chosen. Note that it can happen that ``\bar{a}=a``. Duality implies an isomorphism
+between ``\mathrm{Hom}(W,V)`` and ``\mathrm{Hom}(I,V⊗W^*)``, and thus, for a simple object
+``a``, ``\mathrm{End}(a) ≂ 𝕜`` is isomorphic to ``\mathrm{Hom}(1,a⊗a^*)``, such that the
+latter is also isomorphic to ``𝕜``, or thus ``N^{a\bar{a}}_1 = 1``. Also, all possible
+duals of ``a`` must be isomorphic, and thus there is a single representive ``\bar{a}``,
+meaning that ``N^{ab}_1 = δ^{b,\bar{a}}``, i.e. for all other ``b ≠ \bar{a}``,
```
borisdevos:
```suggestion
duals of ``a`` must be isomorphic, and thus there is a single representative ``\bar{a}`` such that ``N^{ab}_1 = δ^{b,\bar{a}}``, i.e. for all other ``b ≠ \bar{a}``,
```

docs/src/appendix/categories.md
```diff
@@ -887,16 +885,14 @@ allows to conclude that
...
-    In the context of fusion categories, one often resorts to the so-called *isotopic*
-    normalization convention, where splitting tensors are normalized as
-    ``(X^{ab}_{c,μ})^† ∘ X^{ab}_{c′,\mu′} = \sqrt{\frac{d_a d_b}{d_c}} δ_{c,c′} δ_{μ,μ′} \mathrm{id}_c``.
-    This kills some of the quantum dimensions in formulas like the ones above and
-    essentially allows to rotate the graphical notation of splitting and fusion tensors (up
-    to a unitary transformation). Nonetheless, for our implementation of tensors and
-    manipulations thereof (in particular orthonormal factorizations such as the singular
-    value decomposition), we find it more convenient to work with the original normalization
-    convention.
+!!! note In the context of fusion categories, one often resorts to the so-called *isotopic*
```
borisdevos: This note block isn't working. The main text needs to start on a new line with four spaces (https://documenter.juliadocs.org/stable/showcase/#Admonitions)

borisdevos: Actually this might just be an artefact of the preview, since it's fine in the current stable version.

docs/src/lib/sectors.md
```diff
@@ -54,17 +54,17 @@ TensorKitSectors.Dihedral
 TensorKitSectors.ProductGroup
 ```
 
-The following types are used to characterise different properties of the different types
-of sectors:
+The following types are used to characterise different properties of the different types of
```
borisdevos:
```suggestion
The following types are used to characterize different properties of the different types of
```
Since American English is preferred in the docs

docs/src/lib/sectors.md
```diff
@@ -121,8 +121,15 @@ Furthermore, we also have one specific method acting on groups, represented as t
 ×
 ```
 
-Because we sometimes want to customize the string representation of our sector types,
-we also have the following method:
+Mapping between sectors and linear indices is only used for sectors `I` for which
+`Base.IteratorSize(values(I)) == HasLength()`. In that case, we map an index `i` to a sector
```
borisdevos: Is this true? I can call e.g. `values(SU2Irrep)[3]` and it gives `SU2Irrep(1)`

docs/src/lib/spaces.md
```diff
@@ -110,16 +110,16 @@ isepimorphic
...
 ```@docs
 insertleftunit(::ProductSpace, ::Val{i}) where {i}
 insertrightunit(::ProductSpace, ::Val{i}) where {i}
 removeunit(::ProductSpace, ::Val{i}) where {i}
 ```
 
-There are also specific methods for `HomSpace` instances, that are used in determining
-the resuling `HomSpace` after applying certain tensor operations.
+There are also specific methods for `HomSpace` instances, that are used in determining the
+resuling `HomSpace` after applying certain tensor operations.
```
borisdevos:
```suggestion
resulting `HomSpace` after applying certain tensor operations.
```

docs/src/lib/tensors.md
```diff
@@ -38,9 +38,9 @@ TensorMap{T}(::UndefInitializer, V::TensorMapSpace)
...
 
 The resulting object can then be filled with data using the `setindex!` method as discussed
-below, using functions such as `VectorInterface.zerovector!`, `rand!` or `fill!`, or it can 
-be used as an output argument in one of the many methods that accept output arguments, or
-in an `@tensor output[...] = ...` expression.
+below, using functions such as `VectorInterface.zerovector!`, `rand!` or `fill!`, or it can
+be used as an output argument in one of the many methods that accept output arguments, or in
+an `@tensor output[...] = ...` expression.
 
 Alternatively, a `TensorMap` can be constructed by specifying its data, codmain and domain
```
borisdevos:
```suggestion
Alternatively, a `TensorMap` can be constructed by specifying its data, codomain and domain
```

docs/src/lib/tensors.md
```diff
@@ -128,7 +130,7 @@ Base.getindex(::AbstractTensorMap, ::FusionTree, ::FusionTree)
 Base.setindex!(::AbstractTensorMap, ::Any, ::FusionTree, ::FusionTree)
 ```
 
-For a tensor `t` with `FusionType(sectortype(t)) isa UniqueFusion`, fusion trees are 
+For a tensor `t` with `FusionType(sectortype(t)) isa UniqueFusion`, fusion trees are
```
borisdevos:
```suggestion
For a tensor `t` with `FusionStyle(sectortype(t)) isa UniqueFusion`, fusion trees are
```

docs/src/lib/tensors.md
```diff
@@ -155,25 +157,25 @@ Random.randexp!
 The operations that can be performed on an `AbstractTensorMap` can be organized into the
 following categories:
 
-* *vector operations*: these do not change the `space` or index strucure of a tensor and
-  can be straightforwardly implemented on on the full data. All the methods described in
+* *vector operations*: these do not change the `space` or index strucure of a tensor and can
+  be straightforwardly implemented on on the full data. All the methods described in
```
borisdevos:
```suggestion
* *vector operations*: these do not change the `space` or index structure of a tensor and can
  be straightforwardly implemented on the full data. All the methods described in
```

docs/src/lib/tensors.md
```diff
@@ -155,25 +157,25 @@ Random.randexp!
...
-  
+          
 * *(planar) contractions* and *(planar) traces* (i.e., contractions with identity tensors).
-  Tensor contractions correspond to a combination of some index manipulations followed by
-  a composition or multiplication of the tensors in their role as linear maps.
-  Tensor contractions are however of such important and frequency that they require a
-  dedicated implementation.
+  Tensor contractions correspond to a combination of some index manipulations followed by a
+  composition or multiplication of the tensors in their role as linear maps. Tensor
+  contractions are however of such important and frequency that they require a dedicated
```
borisdevos:
```suggestion
  contractions are however of such importance and frequency that they require a dedicated
```

docs/src/lib/tensors.md
```diff
@@ -155,25 +157,25 @@ Random.randexp!
...
+  implementation.
 
 * *tensor factorisations*, which relies on their identification of tensors with linear maps
-  between tensor spaces. The factorisations are applied as ordinary matrix factorisations
-  to the matrix blocks associated with the coupled charges.
+  between tensor spaces. The factorisations are applied as ordinary matrix factorisations to
+  the matrix blocks associated with the coupled charges.
```
borisdevos:
```suggestion
  between tensor spaces. The factorizations are applied as ordinary matrix factorizations to
  the matrix blocks associated with the coupled charges.
```
There's another "factorisation" in the line above

docs/src/lib/tensors.md
```diff
@@ -220,20 +222,22 @@ contract!
 
 ## `TensorMap` factorizations
 
-The factorisation methods are powered by [MatrixAlgebraKit.jl](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl)
-and all follow the same strategy. The idea is that the `TensorMap` is interpreted as a linear
-map based on the current partition of indices between `domain` and `codomain`, and then the
-entire range of MatrixAlgebraKit functions can be called.
-Factorizing a tensor according to a different partition of the indices is possible
-by prepending the factorization step with an explicit call to [`permute`](@ref) or [`transpose`](@ref).
+The factorisation methods are powered by
```
borisdevos:
```suggestion
The factorization methods are powered by
```

docs/src/man/intro.md
```diff
@@ -56,9 +56,9 @@ to the dual space of ``W``. This simple example introduces two new concepts.
 
 1.  Typical vector spaces can appear in the domain and codomain in different related forms,
     e.g. as normal space or dual space. In fact, the most generic case is that every vector
```
borisdevos:
```suggestion
    e.g. as normal spaces or dual spaces. In fact, the most generic case is that every vector
```

docs/src/man/tutorial.md
```diff
@@ -17,12 +17,12 @@ The most important objects in TensorKit.jl are tensors, which we now create with
 A = randn(ℝ^3 ⊗ ℝ^2 ⊗ ℝ^4)
 ```
 
-Note that we entered the tensor size not as plain dimensions, by specifying the vector
-space associated with these tensor indices, in this case `ℝ^n`, which can be obtained by
-typing `\bbR+TAB`. The tensor then lives in the tensor product of the different spaces,
-which we can obtain by typing `⊗` (i.e. `\otimes+TAB`), although for simplicity also the
-usual multiplication sign `*` does the job. Note also that `A` is printed as an instance of
-a parametric type `TensorMap`, which we will discuss below and contains `Tensor`.
+Note that we entered the tensor size not as plain dimensions, by specifying the vector space
```
borisdevos:
```suggestion
Note that we entered the tensor size not as plain dimensions, but by specifying the vector space
```

docs/src/man/tutorial.md
```diff
@@ -35,8 +35,8 @@ supertype(ElementarySpace)
 ```
 
 i.e. `ℝ^n` can also be created without Unicode using the longer syntax `CartesianSpace(n)`.
-It is subtype of `ElementarySpace`, with a standard (Euclidean) inner product
-over the real numbers. Furthermore,
+It is subtype of `ElementarySpace`, with a standard (Euclidean) inner product over the real
```
borisdevos:
```suggestion
It is a subtype of `ElementarySpace` with a standard (Euclidean) inner product over the real
```

docs/src/man/tutorial.md
```diff
@@ -35,8 +35,8 @@ supertype(ElementarySpace)
...
+numbers. Furthermore,
 
 ```@repl tutorial
 W = ℝ^3 ⊗ ℝ^2 ⊗ ℝ^4
```
borisdevos, line 37: Line 60 has a typo: Given that they ~are~ behave as vectors, they also...

docs/src/man/tutorial.md
```diff
@@ -80,16 +80,16 @@ scalarBA′ = dot(B′,A)
...
 ```
 
 We can contract two tensors using Einstein summation convention, which takes the interface
-from [TensorOperations.jl](https://github.com/quantumkithub/TensorOperations.jl). TensorKit.jl
-reexports the `@tensor` macro
+from [TensorOperations.jl](https://github.com/quantumkithub/TensorOperations.jl).
+TensorKit.jl reexports the `@tensor` macro
 
 ```@repl tutorial
 @tensor D[a,b,c,d] := A[a,b,e] * B[d,c,e]
```
borisdevos: Line 105 and onwards uses the outdated `tsvd` function to illustrate a factorization, so the REPL tutorial is just showing errors. This continues to line 141 and 177-183.

docs/src/man/tutorial.md
```diff
@@ -157,7 +157,7 @@ Note that for the construction of `M₁`, in accordance with how one specifies t
 of a matrix (e.g. `randn(4,3)`), the first space is the codomain and the second the domain.
 This is somewhat opposite to the general notation for a function `f:domain→codomain`, so
 that we also support this more mathemical notation, as illustrated in the construction of
```
borisdevos:
```suggestion
that we also support this more mathematical notation, as illustrated in the construction of
```

docs/src/man/tutorial.md
```diff
@@ -201,11 +201,11 @@ In fact, `tsvd(A, ((1, 3), (2,)))` is a shorthand for `tsvd(permute(A, ((1, 3),
 where `tsvd(A::TensorMap)` will just compute the singular value decomposition according to
```
borisdevos: More mentions on old functions and broken REPL displays here and below

docs/src/man/tutorial.md
```diff
@@ -224,8 +224,8 @@ immediately specify the desired codomain and domain indices.
 ## Complex tensors
 
 For applications in e.g. quantum physics, we of course want to work with complex tensors.
-Trying to create a complex-valued tensor with `CartesianSpace` indices is of course
-somewhat contrived and prints a (one-time) warning
+Trying to create a complex-valued tensor with `CartesianSpace` indices is of course somewhat
+contrived and prints a (one-time) warning
 
 ```@repl tutorial
 A = randn(ComplexF64, ℝ^3 ⊗ ℝ^2 ⊗ ℝ^4)
```
borisdevos, line 196: Here below `tsvd` is also used

docs/src/man/tutorial.md
```diff
@@ -323,9 +323,9 @@ Here, we create a 5-dimensional space `V1`, which has a three-dimensional subspa
 associated with charge 0 (the trivial irrep of ``ℤ₂``) and a two-dimensional subspace with
 charge 1 (the non-trivial irrep). Similar for `V2`, where both subspaces are one-
 dimensional. Representing the tensor as a dense `Array`, we see that it is zero in those
```
borisdevos:
```suggestion
charge 1 (the non-trivial irrep). Similar for `V2`, where both subspaces are one-dimensional. Representing the tensor as a dense `Array`, we see that it is zero in those
```

docs/src/man/tutorial.md
```diff
@@ -323,9 +323,9 @@ Here, we create a 5-dimensional space `V1`, which has a three-dimensional subspa
...
-regions where the charges don't add to zero (modulo 2). Of course, the `Tensor(Map)` type
-in TensorKit.jl won't store these zero blocks, and only stores the non-zero information,
-which we can recognize also in the full `Array` representation.
+regions where the charges don't add to zero (modulo 2). Of course, the `Tensor(Map)` type in
+TensorKit.jl won't store these zero blocks, and only stores the non-zero information, which
+we can recognize also in the full `Array` representation.
 
 From there on, the resulting tensors support all of the same operations as the ones we
 encountered in the previous examples.
```
borisdevos: More `tsvd`s here

docs/src/man/tutorial.md
```diff
@@ -378,8 +378,8 @@ supertype(GradedSpace)
 Note that `GradedSpace` is not immediately parameterized by some group `G`, but actually by
```
borisdevos: Line 361 above mentions being able to use `\times+TAB`, but this has been replaced with `\boxtimes+TAB`.

docs/src/man/tutorial.md
```diff
@@ -412,26 +412,26 @@ norm(A) ≈ norm(convert(Array, A))
...
-fusion trees, which are made up out of the Clebsch-Gordan coefficients of the group.
-Indeed, note that the non-zero blocks are also no longer labeled by a list of sectors, but
-by pairs of fusion trees. This will be explained further in the manual. However, the
-Clebsch-Gordan coefficients of the group are only needed to actually convert a tensor to an
-`Array`. For working with tensors with `SU₂Space` indices, e.g. contracting or factorizing
-them, the Clebsch-Gordan coefficients are never needed explicitly. Instead, recoupling
-relations are used to symbolically manipulate the basis of fusion trees, and this only
-requires what is known as the topological data of the group (or its representation theory).
+fusion trees, which are made up out of the Clebsch-Gordan coefficients of the group. Indeed,
+note that the non-zero blocks are also no longer labeled by a list of sectors, but by pairs
```
borisdevos: The new printing no longer does this by default. Maybe mention/use the new `subblocks` function here?

docs/src/man/spaces.md
```diff
@@ -6,16 +6,17 @@ using TensorKit
...
 
-Technically speaking, this name does not capture the full generality that TensorKit.jl supports,
-as instances of subtypes of `VectorSpace` can encode general objects in linear monoidal categories,
-which are not necessarily vector spaces. However, in order not to make the remaining discussion
-to abstract or complicated, we will simply use the nomenclature of vector spaces. In particular,
-we define two abstract subtypes
+Technically speaking, this name does not capture the full generality that TensorKit.jl
+supports, as instances of subtypes of `VectorSpace` can encode general objects in linear
+monoidal categories, which are not necessarily vector spaces. However, in order not to make
+the remaining discussion to abstract or complicated, we will simply use the nomenclature of
```
borisdevos:
```suggestion
the remaining discussion too abstract or complicated, we will simply use the nomenclature of
```

docs/src/man/gradedspaces.md
```diff
@@ -0,0 +1,210 @@
...
+decomposition ``V = ⨁_a ℂ^{n_a} ⊗ R_{a}``. To actually represent such spaces, we now also
+introduce a corresponding type `GradedSpace`, which is a subtype of `ElementarySpace`, i.e.
+```julia
+struct GradedSpace{I<:Sector, D} <: ElementarySpace
+    dims::D
+    dual::Bool
+end
+```
+Here, `D` is a type parameter to denote the data structure used to store the degeneracy or
+multiplicity dimensions ``n_a`` of the different sectors. For conviency, `Vect[I]` will
```
borisdevos:
```suggestion
multiplicity dimensions ``n_a`` of the different sectors. For conveniency, `Vect[I]` will
```

docs/src/man/gradedspaces.md
```diff
@@ -0,0 +1,210 @@
...
+In fact, for consistency, `Vect[Trivial]` will just return `ComplexSpace`, which is not even
+a specific type of `GradedSpace`. For the specific case of group irreps as sectors, one can
+use `Rep[G]` with `G` the group, as inspired by the categorical name
+``\mathbf{Rep}_{\mathsf{G}}``. Some illustrations:
+```@repl gradedspaces
+Vect[Trivial]
+Vect[U1Irrep]
+Vect[Irrep[U₁]]
+Rep[U₁]
+Rep[ℤ₂ × SU₂]
```
borisdevos, line 51: I mentioned above how `\times+TAB` wasn't working, but I was using it wrong. I think my point on `\boxtimes+TAB` still holds there, though.

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,426 @@
...
+Given that the fusion tree can be considered to be the adjoint of a corresponding splitting
+tree ``c → (b_1 ⊗ b_2) ⊗ b_3``, we now first consider splitting trees in isolation. A
+splitting tree which goes from one coupled sector ``c`` to ``N`` uncoupled sectors ``a_1``,
+``a_2``, …, ``a_N`` needs ``N-2`` additional internal sector labels ``e_1``, …, ``e_{N-2}``,
+and, if `FusionStyle(I) isa GenericFusion`, ``N-1`` additional multiplicity labels ``μ_1``,
+…, ``μ_{N-1}``. We henceforth refer to them as vertex labels, as they are associated with
+the vertices of the splitting tree. In the case of `FusionStyle(I) isa UniqueFusion`, the
+internal sectors ``e_1``, …, ``e_{N-2}`` are completely fixed, for
+`FusionStyle(I) isa MultipleFusion` they can also take different values. In our abstract
+notation of the splitting basis ``X^{a_1a_2…a_N}_{c,α}`` used above, ``α`` can be consided a
```
borisdevos:
```suggestion
notation of the splitting basis ``X^{a_1a_2…a_N}_{c,α}`` used above, ``α`` can be considered a
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,426 @@
...
+```julia
+struct FusionTree{I<:Sector,N,M,L}
+    uncoupled::NTuple{N,I}
+    coupled::I
+    isdual::NTuple{N,Bool}
+    innerlines::NTuple{M,I} # fixed to M = N-2
+    vertices::NTuple{L,Int} # fixed to L = N-1
+end
+```
+Here, the fields are probably self-explanotary. The `isdual` field indicates whether an
```
borisdevos:
```suggestion
Here, the fields are probably self-explanatory. The `isdual` field indicates whether an
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,426 @@
...
+``Z`` isomorphism, i.e. the splitting tree in the above example would have
+`sectors = (a₁, a₂, a₃, a₄)`. The `FusionTree` type has a number of basic properties and
+capabilities, such as checking for equality with `==` and support for
+`hash(f::FusionTree, h::UInt)`, as splitting and fusion trees are used as keys in look-up
+tables (i.e. `AbstractDictionary` instances) to look up certain parts of the data of a
+tensor.
+
+`FusionTree` instances are not checked for consistency (i.e. valid fusion rules etc) upon
+creation, hence, they are assumed to be created correctly. The most natural way to create
+them is by using the `fusiontrees(uncoupled::NTuple{N,I}, coupled::I = one(I))` method,
```
borisdevos:
```suggestion
them is by using the `fusiontrees(uncoupled::NTuple{N,I}, coupled::I = unit(I))` method,
```

borisdevos: Oh actually, at some point the default value to the unit sector was removed. This needs to be changed in the docstring as well

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,426 @@
...
+```@raw html
+<img src="../img/tree-insertat.svg" alt="insertat" class="color-invertible"/>
+```
+
+*   [`split(f::FusionTree{I,N}, M::Int)`](@ref TensorKit.split) : splits a fusion tree `f`
+    into two trees `f1` and `f2`, such that `f1` has the first `M` uncoupled sectors of `f`,
+    and `f2` the remaining `N - M`. This function is type stable if `M` is a compile time
+    constant.
+
+    `split(f, M)` is the inverse of `insertat` in the sence that `insertat(f2, 1, f1)`
```
borisdevos:
```suggestion
    `split(f, M)` is the inverse of `insertat` in the sense that `insertat(f2, 1, f1)`
```

docs/src/man/fusiontrees.md
```diff
@@ -0,0 +1,426 @@
...
+checking purposes. Hereto, it is necessary that the *splitting tensor* ``X^{ab}_{c,μ}``,
+i.e. the Clebsch-Gordan coefficients of the group, are encoded via the routine
+`fusiontensor(a,b,c [,μ = nothing])`, where the last argument is only necessary in the case
+of `FusionStyle(I) == GenericFusion()`. We can then convert a `FusionTree{I,N}` into an
+`Array`, which will yield a rank `N+1` array where the first `N` dimensions correspond to
+the uncoupled sectors, and the last dimension to the coupled sector. Note that this is
+mostly useful for the case of `FusionStyle(I) isa MultipleFusion` groups, as in the case of
+abelian groups, all irreps are one-dimensional.
+
+Some examples:
```
borisdevos, line 263: This REPL example doesn't import LinearAlgebra, and also uses `frobeniusschur` instead of `frobenius_schur_phase`.

docs/src/man/symmetries.md
```diff
@@ -0,0 +1,168 @@
...
+required data of the representation theory of a group. We refer to the appendix on
+[categories](@ref s_categories), and in particular the subsection on
+[topological data of a unitary fusion category](@ref ss_topologicalfusion), for more
+details.
+
+!!! note The infrastructure for defining sectors is actually implemented in a standalone
+package, [TensorKitSectors.jl](https://github.com/QuantumKitHub/TensorKitSectors.jl), that
+is imported and reexported by TensorKit.jl.
+
+!!! note On this and the next page of the manual, we assume some famliarity with the
```
borisdevos:
```suggestion
!!! note On this and the next page of the manual, we assume some familiarity with the
```

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+defining a custom inner product in these spaces. Currently, there is no concrete subtype of
+`HasInnerProduct` that is not an `EuclideanProduct`. In particular, `CartesianSpace`,
+`ComplexSpace` and `GradedSpace` all have `InnerProductStyle(S) <: EuclideanProduct`.
+
+With tensors that have `InnerProductStyle(t) <: EuclideanProduct` there is associated an
+adjoint operation, given by `adjoint(t)` or simply `t'`, such that
+`domain(t') == codomain(t)` and `codomain(t') == domain(t)`. Note that for an instance
+`t::TensorMap{S,N₁,N₂}`, `t'` is simply stored in a wrapper called
+`AdjointTensorMap{S,N₂,N₁}`, which is another subtype of `AbstractTensorMap`. This should be
+mostly unvisible to the user, as all methods should work for this type as well. It can be
```
borisdevos:
```suggestion
mostly invisible to the user, as all methods should work for this type as well. It can be
```

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+`hcat`, i.e. we stack them such that the new tensor `catdomain(t1, t2)` has also the same
+codomain, but has a domain which is `domain(t1) ⊕ domain(t2)`. Similarly, if `t1` and `t2`
+are of type `TensorMap{T,S,1,N}` and have the same domain, the operation
+`catcodomain(t1, t2)` results in a new tensor with the same domain and a codomain given by
+`codomain(t1) ⊕ codomain(t2)`, which is the analogy of `vcat`. Note that direct sum only
+makes sense between `ElementarySpace` objects, i.e. there is no way to give a tensor product
+meaning to a direct sum of tensor product spaces.
+
+Time for some more examples:
+```@repl tensors
```
borisdevos, line 50: This REPL is failing because TensorKit isn't exported, and `t` and `V1` aren't defined

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+coevaluation ``η``, but the right evaluation ``\tilde{ϵ}``, when repartitioning the indices
+between domain and codomain.
+
+There are a number of other index related manipulations. We can apply a twist (or inverse
+twist) to one of the tensor map indices via [`twist(t, i; inv = false)`](@ref) or
+[`twist!(t, i; inv = false)`](@ref). Note that the latter method does not store the result
+in a new destination tensor, but just modifies the tensor `t` in place. Twisting several
+indices simultaneously can be obtained by using the defining property
+
+``θ_{V⊗W} = τ_{W,V} ∘ (θ_W ⊗ θ_V) ∘ τ_{V,W} = (θ_V ⊗ θ_W) ∘ τ_{W,V} ∘ τ_{V,W}.``
```
borisdevos:
```suggestion
``θ_{V⊗W} = τ_{W,V} ∘ (θ_W ⊗ θ_V) ∘ τ_{V,W} = (θ_V ⊗ θ_W) ∘ τ_{W,V} ∘ τ_{V,W},``
```

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+
+but is currently not implemented explicitly.
+
+For all sector types `I` with `BraidingStyle(I) == Bosonic()`, all twists are `1` and thus
+have no effect. Let us start with some examples, in which we illustrate that, albeit
+`permute` might act highly non-trivial on the fusion trees and on the corresponding data,
+after conversion to a regular `Array` (when possible), it just acts like `permutedims`
+```@repl tensors
```
borisdevos, line 140: Same as above

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+transpose(m)
+convert(Array, transpose(t)) ≈ permutedims(ta,(4,3,2,1))
+dot(t2, t) ≈ dot(transpose(t2), transpose(t))
+transpose(transpose(t)) ≈ t
+twist(t, 3) ≈ t
+# as twist acts trivially for
+BraidingStyle(sectortype(t))
+```
+Note that `transpose` acts like one would expect on a `TensorMap{T,S,1,1}`. On a
+`TensorMap{TS,N₁,N₂}`, because `transpose` replaces the codomain with the dual of the
```
borisdevos:
```suggestion
`TensorMap{T,S,N₁,N₂}`, because `transpose` replaces the codomain with the dual of the
```

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+domain, which has its tensor product operation reversed, this in the end amounts in a
+complete reversal of all tensor indices when representing it as a plain mutli-dimensional
```
borisdevos:
```suggestion
complete reversal of all tensor indices when representing it as a plain multi-dimensional
```

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+definition that one could think of is `conj(t) = adjoint(transpose(t))`. However note that
+`codomain(adjoint(tranpose(t))) == domain(transpose(t)) == dual(codomain(t))` and similarly
+`domain(adjoint(tranpose(t))) == dual(domain(t))`, where `dual` of a `ProductSpace` is
+composed of the dual of the `ElementarySpace` instances, in reverse order of tensor product.
+This might be very confusing, and as such we leave tensor conjugation undefined. However,
+note that we have a conjugation syntax within the context of [tensor contractions](@ref
+ss_tensor_contraction).
+
+To show the effect of `twist`, we now consider a type of sector `I` for which
+`BraidingStyle{I} != Bosonic()`. In particular, we use `FibonacciAnyon`. We cannot convert
```
borisdevos:
```suggestion
`BraidingStyle(I) != Bosonic()`. In particular, we use `FibonacciAnyon`. We cannot convert
```

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+the resulting `TensorMap` to an `Array`, so we have to rely on indirect tests to verify our
+results.
+
+```@repl tensors
```
borisdevos, line 168: Same here

docs/src/man/tensormanipulations.md
```diff
@@ -0,0 +1,638 @@
...
+hermitian (hence this does not require `EuclideanTensorMap`), and always returns complex
+values eigenvalues and eigenvectors. Like for matrices, `LinearAlgebra.eigen` is type
+unstable and checks hermiticity at run-time, then falling back to either `eig` or `eigh`.
+
+### Orthogonal factorizations
+
+Other factorizations that are provided by TensorKit.jl are orthogonal or unitary in nature,
+and thus always require a Euclidean inner product. However, they don't require equal domain
+and codomain. Let us first discuss the *singular value decomposition*, for which we define
+and export the methods [`tsvd`](@ref) and `tsvd!` (where as always, the latter destroys the
```
borisdevos: `tsvd` mentioned. In general this section is outdated and doesn't mention MAK

docs/src/man/sectors.md
```diff
@@ -1,464 +1,508 @@
...
+appear in the topological data, it means tensor data will necessarily become complex after
+simple manipulations such as permuting indices, and should therefore probably be stored as
+complex numbers from the start.
+
+Finally, additional topological data can be extracted from the minimal interface. In
+particular, the quantum dimensions ``d_a`` and Frobenius-Schur phase ``χ_a`` and indicator
+(only if ``a == \overline{a}``) are encoded in the F-symbol. They are obtained as
+[`dim(a)`](@ref), [`frobenius_schur_phase(a)`](@ref) and
+[`frobenius_schur_indicator(a)`](@ref). These functions have default definitions which
+compute the requested data from `Fsymbol(a, conj(a), a, a, one(a), one(a))`, but they can be
```
borisdevos:
```suggestion
compute the requested data from `Fsymbol(a, dual(a), a, a, unit(a), unit(a))`, but they can be
```
I think it's fine to not make the discrepancy between `unit` and `left/rightunit`

docs/src/man/sectors.md
```diff
@@ -1,464 +1,508 @@
...
+overloaded in case the value can be computed more efficiently. The same holds for related
+fusion manipulations such as the B-symbol, which is obtained as [`Bsymbol(a, b, c)`](@ref).
+Finally, the twist associated with a sector `a` is obtained as [`twist(a)`](@ref), which
+also has a default implementation in terms of the R-symbol. In addition, tThe function
```
borisdevos:
```suggestion
also has a default implementation in terms of the R-symbol. In addition, the function
```

docs/src/man/sectors.md
```diff
@@ -1,464 +1,508 @@
...
 const ℤ₃ = ℤ{3}
 const ℤ₄ = ℤ{4}
+const D₃ = Dihedral{3}
+const D₄ = Dihedral{4}
 const SU₂ = SU{2}
 ```
-Groups themselves are abstract types without any functionality (at least for now). We also
-provide a number of convenient Unicode aliases. These group names are probably self-
-explanatory, except for `CU₁` which is explained below.
+Groups themselves are abstract types without any functionality (at least for now). However,
```
borisdevos: Mention `GroupElement` sector here perhaps?

docs/src/man/sectors.md
```diff
@@ -1,464 +1,508 @@
...
-automatic conversion and pretty printing are provided, as illustrated by the following
-example
+you should not worry about the details of `HalfInt` and additional methods for automatic
+conversion and pretty printing are provided, as illustrated by the following example
 ```@repl sectors
 Irrep[U₁](0.5)
 U1Irrep(0.4)
 U1Irrep(1) ⊗ Irrep[U₁](1//2)
 u = first(U1Irrep(1) ⊗ Irrep[U₁](1//2))
 Nsymbol(u, conj(u), one(u))
```
borisdevos:
```suggestion
Nsymbol(u, dual(u), unit(u))
```

docs/src/man/sectors.md
```diff
@@ -1,464 +1,508 @@
...
+the fusion rules are implemented lazily using the `SectorProductIterator` defined above.
+Furthermore, the topological data (i.e. `Nsymbol` and `Fsymbol`) are provided by the package
+[WignerSymbols.jl](https://github.com/Jutho/WignerSymbols.jl). Note that, while
+WignerSymbols.jl is able to generate the required data in arbitrary precision, we have
+explicitly restricted the scalar type of `SU2Irrep` to `Float64` for efficiency.
+
+The following example illustrates the usage of `SU2Irrep`
 ```@repl sectors
 s = SU2Irrep(3//2)
 conj(s)
```
borisdevos:
```suggestion
dual(s)
```

docs/src/man/sectors.md
```diff
@@ -471,18 +515,18 @@ for s2 in s ⊗ s
...
-``-n``, and turns them together in a joint 2-dimensional index, except for the case
-``n=0``. Irreps are therefore labeled by integers ``n ≧ 0``, however for ``n=0`` the ``ℤ₂``
-symmetry can be realized trivially or non-trivially, resulting in an even and odd one-
-dimensional irrep with ``\mathsf{U})_1`` charge ``0``. Given
-``\mathsf{U}_1 ≂ \mathsf{SO}_2``, this group is also simply known as ``\mathsf{O}_2``, and
-the two representations with `` n = 0`` are the scalar and pseudo-scalar, respectively.
-However, because we also allow for half integer representations, we refer to it as
-`Irrep[CU₁]` or `CU1Irrep` in full.
+Other non-abelian groups for which the irreps are implemented are the dihedral groups
+``\mathsf{D}_N``, and the semidirect product ``\mathsf{U}₁ ⋉ ℤ_2``. In the context of
```
borisdevos:
```suggestion
``\mathsf{D}_N``, the alternating group of order 4 ``\mathsf{A}_4``, and the semidirect product ``\mathsf{U}₁ ⋉ ℤ_2``. In the context of
```

docs/src/man/sectors.md
```diff
@@ -471,18 +515,18 @@ for s2 in s ⊗ s
...
+quantum systems, the latter occurs in the case of systems with particle hole symmetry and
+the non-trivial element of ``ℤ_2`` acts as charge conjugation ``C``. It has the effect of
+interchaning ``\mathsf{U}_1`` irreps ``n`` and ``-n``, and turns them together in a joint
```
borisdevos:
```suggestion
interchanging ``\mathsf{U}_1`` irreps ``n`` and ``-n``, and turns them together in a joint
```

docs/src/man/sectors.md
```diff
@@ -471,18 +515,18 @@ for s2 in s ⊗ s
...
+2-dimensional index, except for the case ``n=0``. Irreps are therefore labeled by integers
+``n ≧ 0``, however for ``n=0`` the ``ℤ₂`` symmetry can be realized trivially or
+non-trivially, resulting in an even and odd one-dimensional irrep with ``\mathsf{U})_1``
```
borisdevos:
```suggestion
non-trivially, resulting in an even and odd one-dimensional irrep with ``\mathsf{U}_1``
```

docs/src/man/sectors.md
```diff
@@ -499,29 +543,35 @@ struct CU1Irrep <: AbstractIrrep{CU₁}
...
-group `SU{N}`, for `N>2`. Such sectors are supported through
+the different cases for the arguments of `Fsymbol`. For the dihedrial groups
+``\mathsf{D}_N``, which can be intepreted as the semidirect product ``\mathbb{Z}_N ⋉ ℤ_2``,
+the representation theory is obtained quite similarly, and is implmented as the type
+[`DNIrrep{N}`](@ref).
+
+By default, none of the groups mentioned above have a reprenenstation theory for which
+`FusionStyle(I) == GenericFusion()`, i.e. where fusion mulitplicities are required. An
+example where this does appear is for the irreps of `SU{N}` for `N>2`. Such sectors are
+supported through
```
borisdevos:
```suggestion
Of the abovementioned groups, only ``mathsf{A}_4`` has a representation theory for which
`FusionStyle(I) == GenericFusion()`, i.e. where fusion multiplicities are required. Another
example where this does appear is for the irreps of `SU{N}` for `N>2`. Such sectors are
supported through
```

docs/src/man/sectors.md
```diff
@@ -499,29 +543,35 @@ struct CU1Irrep <: AbstractIrrep{CU₁}
...
-using the binary operator `⊠`, which can be entered as `\boxtimes`+TAB. First some examples
+It is also possible to combine two or more different types of symmetry sectors, e.g. when
+the total symmetry group is a direct product of individual simple groups. Such combined
+sectors are obtained using the binary operator `⊠`, which can be entered as `\boxtimes`+TAB.
+The resulting type is called [`ProductSector`](@ref), which simply wraps the individual
+sectors, but knows how to combine their fusion and braiding data correctly. First some
+examples
 ```@repl sectors
 a = Z3Irrep(1) ⊠ Irrep[U₁](1)
 typeof(a)
```
borisdevos, line 478: A couple of `conj`s and `one`s here below should be replaced by `dual` and `unit`, respectively.

docs/src/man/sectors.md
```diff
@@ -551,15 +601,19 @@ The symbol `⊠` refers to the
...
-for convience; as `Irrep[ℤ₃] ⊠ Irrep[CU₁]` is not a subtype of the abstract type
-`AbstractIrrep{ℤ₃ × CU₁}`. That behavior cannot be obtained with the Julia's type system.
+representations of the two groups separately. But this definition also extends to other
+categories which are not associated with the representation theory of a group, as discussed
+below. Note that `⊠` also works in the type domain, i.e. `Irrep[ℤ₃] ⊠ Irrep[CU₁]` can be
+used to create `ProductSector{Tuple{Irrep[ℤ₃], Irrep[CU₁]}}`. Instances of this type can be
+constructed by giving a number of arguments, where the first argument is used to construct
+the first sector, and so forth. Furthermore, for representations of groups, we also enabled
+the notation `Irrep[ℤ₃ × CU₁]`, with `×` obtained using `\times+TAB`. However, this is
+merely for convience, as `Irrep[ℤ₃] ⊠ Irrep[CU₁]` is not a subtype of the abstract type
```
borisdevos:
```suggestion
merely for convenience, as `Irrep[ℤ₃] ⊠ Irrep[CU₁]` is not a subtype of the abstract type
```

docs/src/man/sectors.md
```diff
@@ -1315,3 +804,40 @@ Rsymbol(τ,τ,𝟙) |> polar
...
+
+Within TensorKit.jl, the only sector with `NoBraiding` is the [`PlanarTrivial`](@ref)
+sector, which is actually equivalent to the `Trivial` sector, but where the braiding has
+been "disabled" for testing purposes.
+
+Finally, as mentioned above, a recent extension prepares TensorKitSectors.jl to deal with
+multi-fusion categories, where the sectors (simple objects) are organized in a matrix-like
+structure and thus have an additional row and column index. Fusion between sectors is only
+possible when the row and column indices match appropriately; otherwise the fusion product
+is empty. In this structure, the different groups of "diagonal" sectors define separate
```
borisdevos: Is there a better word than "group" that can be used in this and the following sentence?

Jutho:
Also, any general comments? What do you think of the new structure of the manual? Is it an improvement?

docs/src/lib/sectors.md
lkdvos: The original reason for that method is that in the spaces, instead of a `Dict{Sector,Int}` for the degeneracies we are storing a `NTuple{N, Int}` and using direct indexing. This requires a finite and compile-time known number of different sectors.
I would have to double check, but I don't think we are really using that function for anything else right now.

docs/src/man/spaces.md
```diff
@@ -154,36 +156,37 @@ InnerProductStyle(ℝ^3)
...
-one-dimensional space, that is isomorphic to the scalar field underlying the space itself. Finally,
-we have also introduced the non-standard convention `V1 ⊖ V2` (obtained by typing `\ominus`+TAB.)
-in order to obtain a space that is isomorphic to the quotient space of `V1` by `V2`, or thus,
-a particular choice of complement of `V2` in `V1` such that `V1 == V2 ⊕ (V1 ⊖ V2)` is satisfied. 
+Instances of `ElementarySpace` support a number of useful operations. Firstly, we define the
+direct sum of two vector spaces `V1` and `V2` of the same `spacetype` (and with the same
+value of `isdual`) as [`V1 ⊕ V2`](@ref), where `⊕` is obtained by typing `\oplus`+TAB.
+[`zerospace(V)`](@ref) corresponds to the identity or zero element with respect to this
+direct sum operation, i.e. it corresponds to a zero-dimensional space. Furthermore,
+[`unitspace(V)`] (@ref) applied to an elementary space returns a one-dimensional space, that
```
lkdvos:
```suggestion
[`unitspace(V)`](@ref) applied to an elementary space returns a one-dimensional space, that
```

docs/src/man/spaces.md
```diff
@@ -154,36 +156,37 @@ InnerProductStyle(ℝ^3)
...
+is isomorphic to the scalar field underlying the space itself. Finally, we have also
+introduced the non-standard convention `V1 ⊖ V2` (obtained by typing `\ominus`+TAB.) in
```
lkdvos:
```suggestion
introduced the non-standard convention [`V1 ⊖ V2`](@ref) (obtained by typing `\ominus`+TAB.) in
```

docs/src/man/gradedspaces.md
lkdvos, line 1: I think throughout this file the header level of the sections is 3 instead of 2, which makes them not appear in the page list in the sidebar.

docs/src/man/fusiontrees.md
lkdvos, line 1: I think throughout this file the header level of the sections is 3 instead of 2, which makes them not appear in the page list in the sidebar.

lkdvos:
I tried briefly skimming through the updates to just give some global comments, since it is a bit hard to interleave these/not duplicate the things @borisdevos has already marked (Thanks for the huge amount of work @borisdevos !!!)
I can then read a bit more in detail once these comments have been incorporated.

Considering the markdown formatting, I have also been advocating to indeed use a more "LaTeX" style for line breaks as well.
The lack of good markdown formatter is one good argument, but I think it is incredibly hard to figure out what the actual git diff is if entire paragraphs have to be changed/reflowed due to the introduction of a single longer word.
Additionally, github suggestions in text like this also does not work very well with linebreak limits, so in the long run I would simply use the rule that for text, every sentence is on its own line.
I'm okay with splitting up longer sentences on multiple different lines, but in general it tends to be language-wise better to then simply split up the sentence itself to begin with... 
Ideally, we should just at some point decide to do this and have a single PR that implements it everywhere, so we can then spot the git diffs more easily.

---

I definitely like the division in more subpages, large pages make it quite hard to quickly jump to the wanted sections and to lose track of what is going on!

One suggestion I have is about the order of these pages in the manual section.
I think the main thing that feels slightly off to me is having `Vector Spaces` before the symmetries and sectors, and only then proceeding to `Graded Spaces`.
I can see the appeal of introducing a number of methods and concepts without the burden of the symmetries, and I like that idea.
Currently however, I feel like the Vector Spaces section isn't really a simplification of the concepts, and is still rather mathematically inclined both in language and concepts.
It seems also like some of the methods that are mentioned there are somewhat hard to illustrate, mostly because `ComplexSpace` does not have enough structure to show what is going on or why we need these methods to begin with.
I think that since most of our language is still quite mathematical in that section, it might turn out to be easier in the long run to simply bite through the sour apple (😀) and merge graded spaces and vector spaces together, after introducing sectors and symmetries, and then going to fusion trees, and finally tensors and tensor manipulations.

docs/src/appendix/categories.md
Jutho: Yes that is perhaps a good idea.

docs/src/appendix/categories.md
Jutho: Now I think the note blocks got destroyed by the automatic Markdown formatting that did happen in the final commit because of how my VSCode at the office desktop was set up.

I will have to disable that formatter and revert to manual formatting as I originally intended.

docs/src/lib/sectors.md
Jutho: Somehow I cannot commit this suggestion directly so I'll have to change it manually.

docs/src/lib/sectors.md
Jutho: But yes, it is true. There is a default `getindex(::SectorValues{I}, i)` that really just runs the `values(I)` iterator until it reaches the `i`th element. The same with `findindex(::SectorValues{I}, c::I)` that runs the iterator and counts how long it takes before the requested sector `c` is reached.

docs/src/lib/spaces.md
Jutho: This suggestion I can also not commit directly. Very strange.

docs/src/man/gradedspaces.md
Jutho: This should probably have been `For convenience` .

docs/src/man/gradedspaces.md
Jutho: If it is, then this is a bug, but it shouldn't. `boxtimes` is acting at the level of sectors, `times` is acting on groups.

Jutho: `Irrep[G1 × G2] == Irrep[G1] ⊠ Irrep[G2]`

lkdvos (approved):
You should probably quickly check everything is now done correctly, but I did read through the entire docs again, changed the markdown into one sentence per line, and fixed a few typos etc.
For me this is now definitely good to go.

docs/src/man/tensors.md
```diff
@@ -1,266 +1,164 @@
...
+This convention is opposite to the mathematical notation, e.g. ``\mathrm{Hom}(W, V)`` or ``f : W → V``, but originates from the fact that a normal matrix is also denoted as having size `m × n` or is constructed in Julia as `Array(..., (m, n))`, where the first integer `m` refers to the codomain being `m`- dimensional, and the seond integer `n` to the domain being `n`-dimensional.
+This also explains why we have consistently used the symbol ``W`` for spaces in the domain and ``V`` for spaces in the codomain.
+A tensor map ``t : (W_1 ⊗ … ⊗ W_{N_2}) → (V_1 ⊗ … ⊗ V_{N_1})`` will be created in Julia as `TensorMap(..., V1 ⊗ ... ⊗ VN₁, W1 ⊗ ... ⊗ WN₂)`.
 
-Furthermore, the abstract type `AbstractTensor{S,N}` is just a synonym for
-`AbstractTensorMap{S,N,0}`, i.e. for tensor maps with an empty domain, which is equivalent
-to the unit of the monoidal category, or thus, the field of scalars ``𝕜``.
+Furthermore, the abstract type `AbstractTensor{S, N}` is just a synonym for `AbstractTensorMap{S, N, 0}`, i.e. for tensor maps with an empty domain, which is equivalent to the unit of the monoidal category, or thus, the field of scalars ``𝕜``.
```
borisdevos:
```suggestion
Furthermore, the abstract type `AbstractTensor{T, S, N}` is just a synonym for `AbstractTensorMap{T, S, N, 0}`, i.e. for tensor maps with an empty domain, which is equivalent to the unit of the monoidal category, or thus, the field of scalars ``𝕜``.
```

borisdevos: There's a couple of these in this section

docs/src/index.md
```diff
@@ -8,34 +8,23 @@ CurrentModule = TensorKit
...
+Such tensors often have large dimensions and take on a specific structure when symmetries are present.
+By employing concepts from category theory, we can represent and manipulate tensors with a large variety of symmetries, including abelian and non-abelian symmetries, fermionic statistics, as well as generalized (a.k.a. non-invertible or anyonic) symmetries.
+
+At the same time, TensorKit.jl focusses on computational efficiency and performance.
+The underlying storage of a tensor's data can be any `DenseArray`.
+When the data is stored in main memory (corresponding to `Array`), multiple CPUs can be leveraged as many operations come with multithreaded implementations, either by distributing the different blocks in case of a structured tensor (i.e. with symmetries) or by using multithreading provided by the package [Strided.jl](https://github.com/Jutho/Strided.jl).
+Support for storing and manipulating tensors on NVidia and AMD GPUs is currently being developed, whereas support for distributed arrays is planned for the future.
```
borisdevos:
```suggestion
Support for storing and manipulating tensors on Nvidia and AMD GPUs is currently being developed, whereas support for distributed arrays is planned for the future.
```

PR #349, opened 2026-01-08
GPU-friendly truncation implementations

lkdvos:
This is an attempt to get rid of the scalar-indexing oriented approach, and instead do more global operations.

The main point is to take all the singular/eigen values at the same time, and sort them in their entirety.

### DIFF
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -21,6 +21,8 @@ end
 function MatrixAlgebraKit.findtruncated(
         values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByError
     )
+    (isfinite(strategy.p) && strategy.p > 0) ||
+        throw(ArgumentError(lazy"p-norm with p = $(strategy.p) is currently not supported."))
     ϵᵖmax = max(strategy.atol^strategy.p, strategy.rtol^strategy.p * norm(values, strategy.p))
     ϵᵖ = similar(values, typeof(ϵᵖmax))
 
--- a/src/factorizations/truncation.jl
+++ b/src/factorizations/truncation.jl
@@ -221,7 +221,9 @@ end
 # where k is determined by the cumulative truncation error of these values.
 # The strategy is therefore to sort all values, and then use a logical array to indicate
 # which ones to keep.
-function MAK.findtruncated(values::SectorVector, strategy::MAK.TruncationByError)
+function MAK.findtruncated(values::SectorVector, strategy::TruncationByError)
+    (isfinite(strategy.p) && strategy.p > 0) ||
+        throw(ArgumentError(lazy"p-norm with p = $(strategy.p) is currently not supported."))
     ϵᵖmax = max(strategy.atol^strategy.p, strategy.rtol^strategy.p * norm(values, strategy.p))
     ϵᵖ = similar(values, typeof(ϵᵖmax))
 


lkdvos:
I will have a look later, I have a local branch to fix the diagonal implementations already, we'll see what remains after. 
I'll also try if it actually works on GPU, and get a similar implementation going for the truncerror version.

lkdvos:
wait please don't I'm working on this 😆

kshyatt:
Oh no I'm sorry!

ext/TensorKitCUDAExt/truncation.jl
lkdvos, line 1: This has to be removed but requires a release of MatrixAlgebraKit

src/factorizations/diagonal.jl
lkdvos, line 1: TODO: figure out what this has to be once MatrixAlgebraKit releases another patch

kshyatt: Can't this be figured out now (at least on 1.11 or 1.12) by using `[sources]` in `Pkg.toml`

lkdvos: impossible!

src/factorizations/diagonal.jl
```diff
@@ -17,7 +17,7 @@ for f! in (:eig_full!, :eig_trunc!)
     @eval function MAK.initialize_output(
             ::typeof($f!), d::AbstractTensorMap, ::DiagonalAlgorithm
         )
-        return d, similar(d)
+        return similar(d, complex(scalartype(d))), similar(d, complex(scalartype(d)), space(d))
     end
 end
```
Jutho: Maybe the general `initialize_output` implementation is good enough for this, and we can just remove these specializations?

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -0,0 +1,26 @@
...
+
+    perm = sortperm(parent(values); strategy.by, strategy.rev)
+    cumulative_dim = cumsum(Base.permute!(parent(dims), perm))
+
+    result = similar(values, Bool)
+    parent(result)[perm] .= cumulative_dim .<= strategy.howmany
+    return result
+end
+
+# Needed until MatrixAlgebraKit patch hits...
```
kshyatt, line 47: Until which patch hits? Can we get some more description here?

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -0,0 +1,26 @@
...
+function MatrixAlgebraKit._ind_intersect(A::CuVector{Bool}, B::CuVector{Int})
+    result = fill!(similar(A), false)
+    result[B] .= @view A[B]
```
kshyatt, line 50: Dumb question but why not wrap the whole line in a `@view`?

src/factorizations/adjoint.jl
```diff
@@ -81,7 +82,7 @@ for (left_f, right_f) in zip(
 end
 
 # 3-arg functions
-for f in (:svd_full, :svd_compact)
+for f in (:svd_full, :svd_compact, :svd_trunc)
```
kshyatt, line 85: What about `:svd_trunc_no_error`?

src/factorizations/adjoint.jl
```diff
@@ -93,9 +94,16 @@ for f in (:svd_full, :svd_compact)
         return reverse(adjoint.(MAK.initialize_output($f!, adjoint(t), _adjoint(alg))))
     end
 
-    @eval function MAK.$f!(t::AdjointTensorMap, F, alg::AbstractAlgorithm)
-        F′ = $f!(adjoint(t), reverse(adjoint.(F)), _adjoint(alg))
-        return reverse(adjoint.(F′))
+    if f === :svd_trunc
```
kshyatt, line 97: Same comment as above

src/factorizations/adjoint.jl
```diff
@@ -111,6 +119,15 @@ function MAK.svd_compact!(t::AdjointTensorMap, F, alg::DiagonalAlgorithm)
     F′ = svd_compact!(adjoint(t), reverse(adjoint.(F)), _adjoint(alg))
     return reverse(adjoint.(F′))
 end
+function MAK.initialize_output(
+        ::typeof(svd_trunc!), t::AdjointTensorMap, alg::TruncatedAlgorithm
```
kshyatt, line 123: Same comment again

src/factorizations/truncation.jl
```diff
@@ -203,24 +162,38 @@ function MAK.findtruncated(values::SectorVector, ::NoTruncation)
...
-        next = _findnexttruncvalue(values, truncdim; strategy.by, strategy.rev)
-        isnothing(next) && break
-        _, cmin = next
-        truncdim[cmin] -= 1
-        totaldim -= dim(cmin)
-        truncdim[cmin] == 0 && delete!(truncdim, cmin)
+    I = sectortype(values)
+
+
+    if FusionStyle(I) isa UniqueFusion # dimensions are all 1
```
kshyatt: It might be nice to add some comments in here explaining what this code actually does and why, maybe with a graphic example?

ext/TensorKitCUDAExt/truncation.jl
lkdvos: it was supposed to be `# Needed until MatrixAlgebraKit patches this`, I just have to copy that over...

ext/TensorKitCUDAExt/truncation.jl
lkdvos: Uh, no particular reason, I think this should boil down to the same thing, but I'm not sure if there are subtle differences in which code path is taken

kshyatt: I think the logical indexing can also be treated as a view, but won't be unless the macro is all the way to the left? IDK

lkdvos: I don't think this allocates though, the semantics for `setindex!` are slightly different if I'm not mistaken.

This is basically: `setindex!.(result, @view(A[B]), B)`, so the view on the left isn't necessary since no slicing is appearing and it's just broadcasting, while I think that `@view result[B] .= A[B]` would be `copy!(@view(result[B]), @view(A[B]))`, both of which obviously being the same, in the sense that I'd expect the former to lower to the latter

src/factorizations/adjoint.jl
lkdvos: I don't think we are currently supporting this anywhere in TensorKit, so let's keep that for a separate PR

ext/TensorKitCUDAExt/truncation.jl
kshyatt: Fair, I think this isn't a huge deal either way

src/factorizations/truncation.jl
lkdvos: I tried to add some explanation, let me know what you think since I'm not sure I can come up with a good example without having to write quite a lot of text, but I'm happy to rethink the strategy if it isn't clear

kshyatt: Yeah that's definitely better, just to have some sense of what's going on

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -0,0 +1,50 @@
...
+    cumulative_dim = cumsum(Base.permute!(parent(dims), perm))
+
+    result = similar(values, Bool)
+    parent(result)[perm] .= cumulative_dim .<= strategy.howmany
+    return result
+end
+
+function MatrixAlgebraKit.findtruncated(
+        values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByError
+    )
```
Jutho, line 23: This probably needs to be special cased for `p == Inf`, if we want to support this. That in itself should be quite an easy case.

ext/TensorKitCUDAExt/truncation.jl
Jutho: To continue the nitpicking, I don't see the benefit of the `@view` macro in these cases. Both

`@view A[B]`

`view(A, B)` 

have the same number of characters, and `@view(A[B])` even has one more.

src/factorizations/truncation.jl
```diff
@@ -29,31 +29,36 @@ end
...
+    return spacetype(V)(c => _blocklength(dim(V, c), ind) for (c, ind) in pairs(inds))
 end
 
 function truncate_domain!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap, inds)
     for (c, b) in blocks(tdst)
         I = get(inds, c, nothing)
         @assert !isnothing(I)
-        copy!(b, view(block(tsrc, c), :, I))
+        b′ = block(tsrc, c)
+        b .= b′[:, I]
```
Jutho: Is it `copy!` that is not working? Can this still use a view for the right hand side? Or is the `view` the problem?

src/factorizations/truncation.jl
```diff
@@ -29,31 +29,36 @@ end
...
     end
     return tdst
 end
 function truncate_codomain!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap, inds)
     for (c, b) in blocks(tdst)
         I = get(inds, c, nothing)
         @assert !isnothing(I)
-        copy!(b, view(block(tsrc, c), I, :))
+        b′ = block(tsrc, c)
+        b .= b′[I, :]
```
Jutho: Same question.

src/factorizations/truncation.jl
lkdvos: Good catch, it was both `copy!` that isn't working, but additionally the view was a problem for `adjoint`s. However, I patched that differently by also bypassing the `svd_trunc` implementations for adjoint tensors, so this can now be a `view` again

src/factorizations/truncation.jl
```diff
@@ -29,31 +29,36 @@ end
...
     end
     return tdst
 end
 function truncate_diagonal!(Ddst::DiagonalTensorMap, Dsrc::DiagonalTensorMap, inds)
     for (c, b) in blocks(Ddst)
         I = get(inds, c, nothing)
         @assert !isnothing(I)
-        copy!(diagview(b), view(diagview(block(Dsrc, c)), I))
+        diagview(b) .= @view diagview(block(Dsrc, c))[I]
```
Jutho: `@view diagview` hurts my eyes more than this:
```suggestion
        diagview(b) .= view(diagview(block(Dsrc, c)), I)
```

ext/TensorKitCUDAExt/truncation.jl
lkdvos: I just checked, and at least in MatrixAlgebraKit we are currently not supporting this. While I don't necessarily have any strong opinions in favor or against this, I will for now simply add a check and leave this as TBA?

PR #352, opened 2026-01-16
Setup Mooncake extension

lkdvos:
This sets up the structure for beginning to deal with Mooncake in TensorKit.
I think we should merge this like it is ASAP so we can in parallel start working on:
1. Implementing the mooncake tangent type for TensorMap
2. Porting over a bunch of rules


kshyatt (approved):

ext/TensorKitMooncakeExt/tangent.jl
```diff
@@ -0,0 +1,7 @@
+function Mooncake.arrayify(A_dA::CoDual{<:TensorMap})
```
Jutho, line 1: Was there a benefit to overloading `arrayify` versus just making a TensorKit specific function `tensorify`?

ext/TensorKitMooncakeExt/tangent.jl
```diff
@@ -0,0 +1,7 @@
...
+    A = Mooncake.primal(A_dA)
+    dA_fw = Mooncake.tangent(A_dA)
+    data = dA_fw.data.data
+    dA = typeof(A)(data, A.space)
```
Jutho, line 5: Also does this work in the complex case, where `data` is now probably using Mooncake's wonderful complex `Complex` tangent type?

PR #375, opened 2026-02-18
fix: improvements to bypass scalar indexing and improve GPU support

kshyatt:
Needed to get more MPSKit examples working

### DIFF
--- a/ext/TensorKitCUDAExt/cutensormap.jl
+++ b/ext/TensorKitCUDAExt/cutensormap.jl
@@ -6,6 +6,9 @@ const AdjointCuTensorMap{T, S, N₁, N₂} = AdjointTensorMap{T, S, N₁, N₂, 
 function CuTensorMap(t::TensorMap{T, S, N₁, N₂, A}) where {T, S, N₁, N₂, A}
     return CuTensorMap{T, S, N₁, N₂}(CuArray{T}(t.data), space(t))
 end
+function TensorMap{T, S, N₁, N₂, DA}(t::TensorMap{T, S, N₁, N₂, HA}) where {T, S, N₁, N₂, DA <: CuArray{T}, HA <: Array{T}}
+    return CuTensorMap{T, S, N₁, N₂}(CuArray{T}(t.data), space(t))
+end
 
 # project_symmetric! doesn't yet work for GPU types, so do this on the host, then copy
 function TensorKit.project_symmetric_and_check(::Type{T}, ::Type{A}, data::AbstractArray, V::TensorMapSpace; tol = sqrt(eps(real(float(eltype(data)))))) where {T, A <: CuVector{T}}
@@ -101,18 +104,6 @@ function TensorKit.scalar(t::CuTensorMap{T, S, 0, 0}) where {T, S}
     return isempty(inds) ? zero(scalartype(t)) : @allowscalar @inbounds t.data[only(inds)]
 end
 
-function Base.convert(
-        TT::Type{CuTensorMap{T, S, N₁, N₂}},
-        t::AbstractTensorMap{<:Any, S, N₁, N₂}
-    ) where {T, S, N₁, N₂}
-    if typeof(t) === TT
-        return t
-    else
-        tnew = TT(undef, space(t))
-        return copy!(tnew, t)
-    end
-end
-
 function LinearAlgebra.isposdef(t::CuTensorMap)
     domain(t) == codomain(t) ||
         throw(SpaceMismatch("`isposdef` requires domain and codomain to be the same"))
@@ -138,9 +129,8 @@ function Base.promote_rule(
     return CuTensorMap{T, S, N₁, N₂}
 end
 
-TensorKit.promote_storage_rule(::Type{CuArray{T, N}}, ::Type{<:CuArray{T, N}}) where {T, N} =
+TensorKit.promote_storage_rule(::Type{<:CuArray{T, N}}, ::Type{<:CuArray{T, N}}) where {T, N} =
     CuArray{T, N, CUDA.default_memory}
-
 
 # CuTensorMap exponentation:
 function TensorKit.exp!(t::CuTensorMap)
--- a/ext/TensorKitCUDAExt/cutensormap.jl
+++ b/ext/TensorKitCUDAExt/cutensormap.jl
@@ -6,9 +6,6 @@ const AdjointCuTensorMap{T, S, N₁, N₂} = AdjointTensorMap{T, S, N₁, N₂, 
 function CuTensorMap(t::TensorMap{T, S, N₁, N₂, A}) where {T, S, N₁, N₂, A}
     return CuTensorMap{T, S, N₁, N₂}(CuArray{T}(t.data), space(t))
 end
-function TensorMap{T, S, N₁, N₂, DA}(t::TensorMap{T, S, N₁, N₂, HA}) where {T, S, N₁, N₂, DA <: CuArray{T}, HA <: Array{T}}
-    return CuTensorMap{T, S, N₁, N₂}(CuArray{T}(t.data), space(t))
-end
 
 # project_symmetric! doesn't yet work for GPU types, so do this on the host, then copy
 function TensorKit.project_symmetric_and_check(::Type{T}, ::Type{A}, data::AbstractArray, V::TensorMapSpace; tol = sqrt(eps(real(float(eltype(data)))))) where {T, A <: CuVector{T}}
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -10,7 +10,7 @@ function MatrixAlgebraKit.findtruncated(
         fill!(v, dim(c))
     end
 
-    perm = sortperm(parent(values); strategy.by, strategy.rev)
+    perm = isempty(parent(values)) ? () : sortperm(parent(values); strategy.by, strategy.rev)
     cumulative_dim = cumsum(Base.permute!(parent(dims), perm))
 
     result = similar(values, Bool)
@@ -36,7 +36,7 @@ function MatrixAlgebraKit.findtruncated(
         end
     end
 
-    perm = sortperm(parent(values); by = abs, rev = false)
+    perm = isempty(parent(values)) ? () : sortperm(parent(values); by = abs, rev = false)
     cumulative_err = cumsum(Base.permute!(parent(ϵᵖ), perm))
 
     result = similar(values, Bool)
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -10,7 +10,7 @@ function MatrixAlgebraKit.findtruncated(
         fill!(v, dim(c))
     end
 
-    perm = isempty(parent(values)) ? () : sortperm(parent(values); strategy.by, strategy.rev)
+    perm = isempty(parent(values)) ? Int64[] : sortperm(parent(values); strategy.by, strategy.rev)
     cumulative_dim = cumsum(Base.permute!(parent(dims), perm))
 
     result = similar(values, Bool)
@@ -36,7 +36,7 @@ function MatrixAlgebraKit.findtruncated(
         end
     end
 
-    perm = isempty(parent(values)) ? () : sortperm(parent(values); by = abs, rev = false)
+    perm = isempty(parent(values)) ? Int64[] : sortperm(parent(values); by = abs, rev = false)
     cumulative_err = cumsum(Base.permute!(parent(ϵᵖ), perm))
 
     result = similar(values, Bool)
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -10,7 +10,9 @@ function MatrixAlgebraKit.findtruncated(
         fill!(v, dim(c))
     end
 
-    perm = isempty(parent(values)) ? Int64[] : sortperm(parent(values); strategy.by, strategy.rev)
+    isempty(parent(values)) && return similar(values, Bool)
+    
+    perm = sortperm(parent(values); strategy.by, strategy.rev)
     cumulative_dim = cumsum(Base.permute!(parent(dims), perm))
 
     result = similar(values, Bool)
@@ -36,7 +38,9 @@ function MatrixAlgebraKit.findtruncated(
         end
     end
 
-    perm = isempty(parent(values)) ? Int64[] : sortperm(parent(values); by = abs, rev = false)
+    isempty(parent(values)) && return similar(values, Bool)
+    
+    perm = sortperm(parent(values); by = abs, rev = false)
     cumulative_err = cumsum(Base.permute!(parent(ϵᵖ), perm))
 
     result = similar(values, Bool)
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -50,6 +50,12 @@ function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::S)
     return CUDA.CUDACore.Adapt.adapt(Vector{Int}, MatrixAlgebraKit.findtruncated(values, strategy))
 end
 
+function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByOrder)
+    # returning a CuSectorVector wrecks things in truncate_{co}domain
+    # because of scalar indexing
+    return CUDA.CUDACore.Adapt.adapt(Vector{Int}, MatrixAlgebraKit.findtruncated(values, strategy))
+end
+
 function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByValue)
     atol = TensorKit.Factorizations.rtol_to_atol(values, strategy.p, strategy.atol, strategy.rtol)
     strategy′ = trunctol(; atol, strategy.by, strategy.keep_below)
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -56,6 +56,12 @@ function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::Ma
     return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
 end
 
+function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::TensorKit.Factorizations.TruncationSpace)
+    # returning a CuSectorVector wrecks things in truncate_{co}domain
+    # because of scalar indexing
+    return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
+end
+
 function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByValue)
     atol = TensorKit.Factorizations.rtol_to_atol(values, strategy.p, strategy.atol, strategy.rtol)
     strategy′ = trunctol(; atol, strategy.by, strategy.keep_below)
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -50,16 +50,12 @@ function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::S)
     return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
 end
 
-function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByOrder)
-    # returning a CuSectorVector wrecks things in truncate_{co}domain
-    # because of scalar indexing
-    return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
-end
-
-function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::TensorKit.Factorizations.TruncationSpace)
-    # returning a CuSectorVector wrecks things in truncate_{co}domain
-    # because of scalar indexing
-    return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
+for strat in (:(MatrixAlgebraKit.TruncationByOrder), :(MatrixAlgebraKit.TruncationByError), :(MatrixAlgebraKit.TruncationIntersection), :(TensorKit.Factorizations.TruncationSpace))
+    @eval function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::$strat)
+        # returning a CuSectorVector wrecks things in truncate_{co}domain
+        # because of scalar indexing
+        return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
+    end
 end
 
 function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByValue)
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -47,19 +47,19 @@ end
 function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::S) where {S <: MatrixAlgebraKit.TruncationStrategy}
     # returning a CuSectorVector wrecks things in truncate_{co}domain
     # because of scalar indexing
-    return CUDA.CUDACore.Adapt.adapt(Vector{Int}, MatrixAlgebraKit.findtruncated(values, strategy))
+    return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
 end
 
 function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByOrder)
     # returning a CuSectorVector wrecks things in truncate_{co}domain
     # because of scalar indexing
-    return CUDA.CUDACore.Adapt.adapt(Vector{Int}, MatrixAlgebraKit.findtruncated(values, strategy))
+    return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
 end
 
 function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByValue)
     atol = TensorKit.Factorizations.rtol_to_atol(values, strategy.p, strategy.atol, strategy.rtol)
     strategy′ = trunctol(; atol, strategy.by, strategy.keep_below)
-    return SectorDict(c => CUDA.CUDACore.Adapt.adapt(Vector{Int}, MatrixAlgebraKit.findtruncated_svd(d, strategy′)) for (c, d) in pairs(values))
+    return SectorDict(c => CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated_svd(d, strategy′)) for (c, d) in pairs(values))
 end
 
 # Needed until MatrixAlgebraKit patch hits...
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -44,6 +44,18 @@ function MatrixAlgebraKit.findtruncated(
     return result
 end
 
+function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::S) where {S <: MatrixAlgebraKit.TruncationStrategy}
+    # returning a CuSectorVector wrecks things in truncate_{co}domain
+    # because of scalar indexing
+    return CUDA.CUDACore.Adapt.adapt(Vector{Int}, MatrixAlgebraKit.findtruncated(values, strategy))
+end
+
+function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByValue)
+    atol = TensorKit.Factorizations.rtol_to_atol(values, strategy.p, strategy.atol, strategy.rtol)
+    strategy′ = trunctol(; atol, strategy.by, strategy.keep_below)
+    return SectorDict(c => CUDA.CUDACore.Adapt.adapt(Vector{Int}, MatrixAlgebraKit.findtruncated_svd(d, strategy′)) for (c, d) in pairs(values))
+end
+
 # Needed until MatrixAlgebraKit patch hits...
 function MatrixAlgebraKit._ind_intersect(A::CuVector{Bool}, B::CuVector{Int})
     result = fill!(similar(A), false)
--- a/ext/TensorKitCUDAExt/truncation.jl
+++ b/ext/TensorKitCUDAExt/truncation.jl
@@ -11,7 +11,7 @@ function MatrixAlgebraKit.findtruncated(
     end
 
     isempty(parent(values)) && return similar(values, Bool)
-    
+
     perm = sortperm(parent(values); strategy.by, strategy.rev)
     cumulative_dim = cumsum(Base.permute!(parent(dims), perm))
 
@@ -39,7 +39,7 @@ function MatrixAlgebraKit.findtruncated(
     end
 
     isempty(parent(values)) && return similar(values, Bool)
-    
+
     perm = sortperm(parent(values); by = abs, rev = false)
     cumulative_err = cumsum(Base.permute!(parent(ϵᵖ), perm))
 
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -53,9 +53,7 @@ storagetype(t) = storagetype(typeof(t))
 function storagetype(::Type{T}) where {T <: AbstractTensorMap}
     if T isa Union
         # attempt to be slightly more specific by promoting unions
-        Ma = storagetype(T.a)
-        Mb = storagetype(T.b)
-        return promote_storagetype(Ma, Mb)
+        return promote_storagetype(T.a, T.b)
     else
         # fallback definition by using scalartype
         return similarstoragetype(scalartype(T))
@@ -103,11 +101,19 @@ similarstoragetype(X::Type, ::Type{T}) where {T <: Number} =
 
 # implement on tensors
 similarstoragetype(::Type{TT}) where {TT <: AbstractTensorMap} = similarstoragetype(storagetype(TT))
-similarstoragetype(::Type{TT}, ::Type{T}) where {TT <: AbstractTensorMap, T <: Number} =
-    similarstoragetype(storagetype(TT), T)
+function similarstoragetype(::Type{TT}, ::Type{T}) where {TT <: AbstractTensorMap, T <: Number}
+    return similarstoragetype(storagetype(TT), T)
+end
+function similarstoragetype(::Type{<:AbstractTensorMap{T, S, N₁, N₂}}, ::Type{TA}) where {T <: Number, TA <: DenseVector, S, N₁, N₂}
+    return similarstoragetype(TA, T)
+end
+function similarstoragetype(t::AbstractTensorMap{T, S, N₁, N₂}, ::Type{TA}) where {T <: Number, TA <: DenseVector, S, N₁, N₂}
+    return similarstoragetype(typeof(t), TA)
+end
 
 # implement on arrays
 similarstoragetype(::Type{A}) where {A <: DenseVector{<:Number}} = A
+similarstoragetype(::Type{A}, ::Type{A}) where {A <: DenseVector{<:Number}} = A
 Base.@assume_effects :foldable similarstoragetype(::Type{A}) where {A <: AbstractArray{<:Number}} =
     Core.Compiler.return_type(similar, Tuple{A, Int})
 Base.@assume_effects :foldable similarstoragetype(::Type{A}, ::Type{T}) where {A <: AbstractArray, T <: Number} =
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -104,12 +104,6 @@ similarstoragetype(::Type{TT}) where {TT <: AbstractTensorMap} = similarstoraget
 function similarstoragetype(::Type{TT}, ::Type{T}) where {TT <: AbstractTensorMap, T <: Number}
     return similarstoragetype(storagetype(TT), T)
 end
-function similarstoragetype(::Type{<:AbstractTensorMap{T, S, N₁, N₂}}, ::Type{TA}) where {T <: Number, TA <: DenseVector, S, N₁, N₂}
-    return similarstoragetype(TA, T)
-end
-function similarstoragetype(t::AbstractTensorMap{T, S, N₁, N₂}, ::Type{TA}) where {T <: Number, TA <: DenseVector, S, N₁, N₂}
-    return similarstoragetype(typeof(t), TA)
-end
 
 # implement on arrays
 similarstoragetype(::Type{A}) where {A <: DenseVector{<:Number}} = A
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -113,7 +113,6 @@ end
 
 # implement on arrays
 similarstoragetype(::Type{A}) where {A <: DenseVector{<:Number}} = A
-similarstoragetype(::Type{A}, ::Type{A}) where {A <: DenseVector{<:Number}} = A
 Base.@assume_effects :foldable similarstoragetype(::Type{A}) where {A <: AbstractArray{<:Number}} =
     Core.Compiler.return_type(similar, Tuple{A, Int})
 Base.@assume_effects :foldable similarstoragetype(::Type{A}, ::Type{T}) where {A <: AbstractArray, T <: Number} =
--- a/src/tensors/adjoint.jl
+++ b/src/tensors/adjoint.jl
@@ -22,6 +22,8 @@ Base.adjoint(t::AbstractTensorMap) = AdjointTensorMap(t)
 space(t::AdjointTensorMap) = adjoint(space(parent(t)))
 dim(t::AdjointTensorMap) = dim(parent(t))
 storagetype(::Type{AdjointTensorMap{T, S, N₁, N₂, TT}}) where {T, S, N₁, N₂, TT} = storagetype(TT)
+similarstoragetype(::AdjointTensorMap{T, S, N₁, N₂, TT}, ::Type{T′}) where {T, S, N₁, N₂, TT, T′ <: Number} = similarstoragetype(TT, T′)
+similarstoragetype(::AdjointTensorMap{T, S, N₁, N₂, TT}, ::Type{TA}) where {T, S, N₁, N₂, TT, TA <: DenseVector} = similarstoragetype(TT, TA)
 
 # Blocks and subblocks
 #----------------------
--- a/src/tensors/adjoint.jl
+++ b/src/tensors/adjoint.jl
@@ -50,8 +50,6 @@ Base.@propagate_inbounds function subblock(t::AdjointTensorMap, (f₁, f₂)::Tu
     return permutedims(conj(data), (domainind(tp)..., codomainind(tp)...))
 end
 
-to_cpu(t::AdjointTensorMap) = adjoint(to_cpu(adjoint(t)))
-
 # Show
 #------
 function Base.showarg(io::IO, t::AdjointTensorMap, toplevel::Bool)
--- a/src/tensors/adjoint.jl
+++ b/src/tensors/adjoint.jl
@@ -22,8 +22,6 @@ Base.adjoint(t::AbstractTensorMap) = AdjointTensorMap(t)
 space(t::AdjointTensorMap) = adjoint(space(parent(t)))
 dim(t::AdjointTensorMap) = dim(parent(t))
 storagetype(::Type{AdjointTensorMap{T, S, N₁, N₂, TT}}) where {T, S, N₁, N₂, TT} = storagetype(TT)
-similarstoragetype(::AdjointTensorMap{T, S, N₁, N₂, TT}, ::Type{T′}) where {T, S, N₁, N₂, TT, T′ <: Number} = similarstoragetype(TT, T′)
-similarstoragetype(::AdjointTensorMap{T, S, N₁, N₂, TT}, ::Type{TA}) where {T, S, N₁, N₂, TT, TA <: DenseVector} = similarstoragetype(TT, TA)
 
 # Blocks and subblocks
 #----------------------
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -189,15 +189,13 @@ end
 has_shared_permute(t::BraidingTensor, ::Index2Tuple) = false
 function add_transform!(
         tdst::AbstractTensorMap,
-        tsrc::BraidingTensor{T, S},
+        tsrc::BraidingTensor{T, S, A},
         (p₁, p₂)::Index2Tuple,
         fusiontreetransform,
         α::Number, β::Number, backend::AbstractBackend...
-    ) where {T, S}
-    tsrc_map = similar(tdst, storagetype(tdst), space(tsrc))
-    copy!(tsrc_map, tsrc)
+    ) where {T, S, A}
     return add_transform!(
-        tdst, tsrc_map, (p₁, p₂), fusiontreetransform, α, β,
+        tdst, TensorMap(tsrc), (p₁, p₂), fusiontreetransform, α, β,
         backend...
     )
 end
@@ -265,8 +263,7 @@ function planarcontract!(
     )
     # special case only defined for contracting all 4 indices of B (2 contracted + 2 open)
     if length.(pB) != (2, 2)
-        tB = TensorMapWithStorage{eltype(B), similarstoragetype(A, eltype(B)), spacetype(tB′), numout(tB′), numin(tB′)}(B)
-        return planarcontract!(C, A, pA, tB, pB, pAB, α, β, backend, allocator)
+        return planarcontract!(C, A, pA, TensorMap(B), pB, pAB, α, β, backend, allocator)
     end
 
     spacecheck_contract(C, A, pA, false, B, pB, false, pAB)
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -261,8 +261,7 @@ function planarcontract!(
     )
     # special case only defined for contracting all 4 indices of B (2 contracted + 2 open)
     length.(pB) == (2, 2) ||
-        planarcontract!(C, A, pA, TensorMap(B), pB, pAB, α, β, backend, allocator)
-
+        return planarcontract!(C, A, pA, TensorMap(B), pB, pAB, α, β, backend, allocator)
 
     spacecheck_contract(C, A, pA, false, B, pB, false, pAB)
 
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -189,8 +189,6 @@ end
 has_shared_permute(t::BraidingTensor, ::Index2Tuple) = false
 function add_transform!(
         tdst::AbstractTensorMap,
-        tsrc::BraidingTensor,
-        (p₁, p₂)::Index2Tuple,
         tsrc::BraidingTensor, (p₁, p₂)::Index2Tuple,
         α::Number, β::Number, backend::AbstractBackend...
     )
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -189,12 +189,15 @@ end
 has_shared_permute(t::BraidingTensor, ::Index2Tuple) = false
 function add_transform!(
         tdst::AbstractTensorMap,
-        tsrc::BraidingTensor, (p₁, p₂)::Index2Tuple,
+        tsrc::BraidingTensor{T, S},
+        (p₁, p₂)::Index2Tuple,
         fusiontreetransform,
         α::Number, β::Number, backend::AbstractBackend...
-    )
+    ) where {T, S}
+    tsrc_map = similar(tdst, storagetype(tdst), space(tsrc))
+    copy!(tsrc_map, tsrc)
     return add_transform!(
-        tdst, TensorMap(tsrc), (p₁, p₂), fusiontreetransform, α, β,
+        tdst, tsrc_map, (p₁, p₂), fusiontreetransform, α, β,
         backend...
     )
 end
@@ -261,8 +264,11 @@ function planarcontract!(
         backend, allocator
     )
     # special case only defined for contracting all 4 indices of B (2 contracted + 2 open)
-    length.(pB) == (2, 2) ||
-        return planarcontract!(C, A, pA, TensorMap(B), pB, pAB, α, β, backend, allocator)
+    if length.(pB) != (2, 2)
+        tB′ = TensorMap(B)
+        tB = TensorMapWithStorage{eltype(B), similarstoragetype(A, eltype(B)), spacetype(tB′), numout(tB′), numin(tB′)}(tB′)
+        return planarcontract!(C, A, pA, tB, pB, pAB, α, β, backend, allocator)
+    end
 
     spacecheck_contract(C, A, pA, false, B, pB, false, pAB)
 
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -190,6 +190,7 @@ has_shared_permute(t::BraidingTensor, ::Index2Tuple) = false
 function add_transform!(
         tdst::AbstractTensorMap,
         tsrc::BraidingTensor, (p₁, p₂)::Index2Tuple,
+        fusiontreetransform,
         α::Number, β::Number, backend::AbstractBackend...
     )
     return add_transform!(
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -189,11 +189,11 @@ end
 has_shared_permute(t::BraidingTensor, ::Index2Tuple) = false
 function add_transform!(
         tdst::AbstractTensorMap,
-        tsrc::BraidingTensor{T, S, A},
+        tsrc::BraidingTensor,
         (p₁, p₂)::Index2Tuple,
         fusiontreetransform,
         α::Number, β::Number, backend::AbstractBackend...
-    ) where {T, S, A}
+    )
     return add_transform!(
         tdst, TensorMap(tsrc), (p₁, p₂), fusiontreetransform, α, β,
         backend...
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -191,7 +191,7 @@ function add_transform!(
         tdst::AbstractTensorMap,
         tsrc::BraidingTensor,
         (p₁, p₂)::Index2Tuple,
-        fusiontreetransform,
+        tsrc::BraidingTensor, (p₁, p₂)::Index2Tuple,
         α::Number, β::Number, backend::AbstractBackend...
     )
     return add_transform!(
@@ -262,9 +262,9 @@ function planarcontract!(
         backend, allocator
     )
     # special case only defined for contracting all 4 indices of B (2 contracted + 2 open)
-    if length.(pB) != (2, 2)
-        return planarcontract!(C, A, pA, TensorMap(B), pB, pAB, α, β, backend, allocator)
-    end
+    length.(pB) == (2, 2) ||
+        planarcontract!(C, A, pA, TensorMap(B), pB, pAB, α, β, backend, allocator)
+
 
     spacecheck_contract(C, A, pA, false, B, pB, false, pAB)
 
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -265,8 +265,7 @@ function planarcontract!(
     )
     # special case only defined for contracting all 4 indices of B (2 contracted + 2 open)
     if length.(pB) != (2, 2)
-        tB′ = TensorMap(B)
-        tB = TensorMapWithStorage{eltype(B), similarstoragetype(A, eltype(B)), spacetype(tB′), numout(tB′), numin(tB′)}(tB′)
+        tB = TensorMapWithStorage{eltype(B), similarstoragetype(A, eltype(B)), spacetype(tB′), numout(tB′), numin(tB′)}(B)
         return planarcontract!(C, A, pA, tB, pB, pAB, α, β, backend, allocator)
     end
 
--- a/src/tensors/diagonal.jl
+++ b/src/tensors/diagonal.jl
@@ -280,7 +280,7 @@ end
 # ----------------
 function TO.tensoradd_type(TC, A::DiagonalTensorMap, ::Index2Tuple{1, 1}, ::Bool)
     M = similarstoragetype(A, TC)
-    return DiagonalTensorMap{TC, spacetype(A), M}
+    return DiagonalTensorMap{scalartype(M), spacetype(A), M}
 end
 
 function TO.tensorcontract_type(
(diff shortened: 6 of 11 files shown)


ext/TensorKitCUDAExt/auxiliary.jl
```diff
@@ -0,0 +1,28 @@
+function TensorKit._copyto!(A::StridedView{TA, 1, <:CuArray{TA}}, B::StridedView{TB, 2, <:CuArray{TB}}) where {TA, TB}
```
lkdvos: Does this make sense to include, and should this not simply fall back to the default `copyto!`?
This really is just a performance optimization to avoid a bunch of the overhead of Strided.jl, but I would be surprised that building the indexarrays like this really gives an improvement over just a regular strided `copyto!`.

I think this entire thing should boil down to the following, which is not obvious and I should have added a comment/fallback definition: (up to some off-by-one errors though)

```julia
A[A.offset:stride(A, 1):end] .= B.op.(view(B, div(B.offset, stride(B, 2)):stride(B, 1):size(B, 1), 1:stride(B, 2):size(B, 2)))
```

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -7,6 +7,17 @@ function CuTensorMap(t::TensorMap{T, S, N₁, N₂, A}) where {T, S, N₁, N₂,
     return CuTensorMap{T, S, N₁, N₂}(CuArray{T}(t.data), space(t))
 end
 
+#=function TensorKit.TensorMap{T, S₁, N₁, N₂, A}(
```
lkdvos: leftover?

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -17,6 +28,10 @@ function TensorKit.project_symmetric_and_check(::Type{T}, ::Type{A}, data::Abstr
     return TensorKit.TensorMapWithStorage{T, A}(A(h_t.data), V)
 end
 
+function TensorKit.blocktype(::Type{<:CuTensorMap{T, S}}) where {T, S}
+    return SubArray{T, 1, CuVector{T, CUDA.DeviceMemory}, Tuple{UnitRange{Int}}, true}
```
lkdvos: I somehow had expected the blocktype to be `CuMatrix`, with the way that `CUDA` handles views. If this isn't the case, can we force it to be?

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -140,6 +167,8 @@ end
 
 TensorKit.promote_storage_rule(::Type{CuArray{T, N}}, ::Type{<:CuArray{T, N}}) where {T, N} =
     CuArray{T, N, CUDA.default_memory}
+TensorKit.promote_storage_rule(::Type{<:CuArray{T, N}}, ::Type{CuArray{T, N}}) where {T, N} =
+    CuArray{T, N, CUDA.default_memory}
```
lkdvos:
```suggestion
TensorKit.promote_storage_rule(::Type{<:CuArray{T, N}}, ::Type{<:CuArray{T, N}}) where {T, N} =
CuArray{T, N, CUDA.default_memory}
```
I should have written the rules in such a way that it is symmetric, so we shouldn't have to define both directions. However, I do think both sides need `<:` to account for the third type parameter being there, which I also missed in the last PR.

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -102,9 +117,21 @@ function TensorKit.scalar(t::CuTensorMap{T, S, 0, 0}) where {T, S}
 end
 
 function Base.convert(
```
lkdvos: I'm again a bit confused by the necessity of this function, is that not the same definition as the regular `TensorMap` one?

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -102,9 +117,21 @@ function TensorKit.scalar(t::CuTensorMap{T, S, 0, 0}) where {T, S}
...
+    ) where {T, S, N₁, N₂, A <: CuArray{T}, AA}
+    if typeof(t) === TT
+        return t
+    else
+        tnew = TT(undef, space(t))
+        return copy!(tnew, t)
+    end
+end
+
+function Base.convert(
```
lkdvos: Same comment here

src/tensors/braidingtensor.jl
```diff
@@ -171,12 +171,15 @@ end
...
 function add_transform!(
         tdst::AbstractTensorMap,
-        tsrc::BraidingTensor, (p₁, p₂)::Index2Tuple,
+        tsrc::BraidingTensor{T, S},
+        (p₁, p₂)::Index2Tuple,
         fusiontreetransform,
         α::Number, β::Number, backend::AbstractBackend...
-    )
+    ) where {T, S}
+    tsrc_map = TensorMapWithStorage{scalartype(tdst), storagetype(tdst)}(undef, (tsrc.V2 ⊗ tsrc.V1) ← (tsrc.V1 ⊗ tsrc.V2))
```
lkdvos:
```suggestion
    tsrc_map = similar(tdst, storagetype(tdst), space(tsrc))
```
This might be a little cleaner/not use that much "internal"s

src/tensors/treetransformers.jl
```diff
@@ -46,7 +46,7 @@ function AbelianTreeTransformer(transform, p, Vdst, Vsrc)
 end
 
 const _GenericTransformerData{T, N} = Tuple{
-    Matrix{T},
+    DenseMatrix{T},
```
lkdvos: I think this change makes the types below abstractly typed, do we need this?

ext/TensorKitCUDAExt/auxiliary.jl
kshyatt: It seems to be necessary to avoid scalar indexing sadness 🤷 . Happy to use the fallback, though!

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: 😰

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Actually it wanted it to be `ReshapedArray` of this `SubArray` 😱 . Really painful. I can swap this to just being a `CuMatrix`.

src/tensors/treetransformers.jl
kshyatt: Yes, in order to allow device-side matrices to get passed in. Otherwise you get attempts to multiply `CuMatrix` * `Matrix` outside of constructors

lkdvos: Ok, but in that case we would really have to make that an additional type parameter in the `GenericTreeTransformer` struct -- these were introduced to hyper specialize and get maximal efficiency, so I don't think we can eat a type-instability here.

kshyatt: OK, it would have been helpful to have had a comment or anything that this was why they were there

kshyatt:
Let's make this a draft too to cut down on CI thrash

ext/TensorKitCUDAExt/auxiliary.jl
lkdvos: Just investigated this a bit more, couple comments:

- My fallback is needlessly complicated, and should have just been `Base.copyto!(A, B)`, which then dispatches to Strided.jl
- If that fails, the fallback is `copy!(sreshape(A, size(B)), B)`, which I think works with your last changes.

It might be reasonable to turn around the logic here, and simply go from opt-out to opt-in, i.e. `TensorKit._copyto!(A, B) = copyto!(A, B)` and then only specialize this for `<:Vector` + `<:Memory` parent types.

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: This got fixed.

ext/TensorKitCUDAExt/auxiliary.jl
kshyatt: TBH all this might be obviated by the fixes that have now been merged into `Strided` and `StridedViews`, right? Why don't I just nuke this and we can see if we need it?

lkdvos: Yes, but this function still bypasses all of the Strided stuff because in this really specific case I have a bit more information and could squeeze out a tiny bit more performance. Ultimately though, if this turns out to be too much of a hassle it might be reasonable to choose maintainability and simply replace this at the callsites

kshyatt: Giving this a trial run, let's see how it goes...

src/tensors/treetransformers.jl
kshyatt: Might no longer be needed, hooray!

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -168,3 +159,30 @@ for f in (:sqrt, :log, :asin, :acos, :acosh, :atanh, :acoth)
...
+end
+
+function TensorKit.allocate_buffers(
+        tdst::CuTensorMap, tsrc::CuTensorMap, transformer::TensorKit.GenericTreeTransformer
+    )
+    sz = TensorKit.buffersize(transformer)
+    # force zeros to ensure the buffers are empty
+    # otherwise memory re-use can fill them with garbage data
+    return CUDA.zeros(eltype(tdst.data), sz), CUDA.zeros(eltype(tsrc.data), sz)
+end
```
lkdvos: This is slightly confusing to me, the `zeros` shouldn't be necessary (in fact, the implementation reuses the start of the buffer for each of the different blocks anyways), so I would have guessed that the `similar(tsrc.data, sz)` calls should be sufficient and correctly allocate device arrays here?

src/tensors/abstracttensor.jl
```diff
@@ -53,9 +53,11 @@ storagetype(t) = storagetype(typeof(t))
 function storagetype(::Type{T}) where {T <: AbstractTensorMap}
     if T isa Union
         # attempt to be slightly more specific by promoting unions
-        Ma = storagetype(T.a)
-        Mb = storagetype(T.b)
-        return promote_storagetype(Ma, Mb)
+        return promote_storagetype(T.a, T.b)
+    elseif eltype(T) isa Union
```
lkdvos: Is this to better support `BlockTensorMap`? Do we ever have tensors with union scalartypes?

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -168,3 +159,30 @@ for f in (:sqrt, :log, :asin, :acos, :acosh, :atanh, :acoth)
...
+        # Special case without intermediate buffers whenever there is only a single block
+        if length(subtransformer[1]) == 1
+            TensorKit._add_transform_single!(tdst, tsrc, p, subtransformer, α, β, backend...)
+        else
+            cu_subtransformer = tuple(CUDA.adapt(CuArray, subtransformer[1]), subtransformer[2:end]...)
+            TensorKit._add_transform_multi!(tdst, tsrc, p, cu_subtransformer, buffers, α, β, backend...)
+        end
+    end
+    return nothing
+end
```
lkdvos: I guess the only change here is to promote the unitary basis transformation into a CuArray, which probably makes more sense to just support at the `mul` callsite (which I think @kshyatt already fixed, so this might no longer be required?)

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -17,6 +17,10 @@ function TensorKit.project_symmetric_and_check(::Type{T}, ::Type{A}, data::Abstr
     return TensorKit.TensorMapWithStorage{T, A}(A(h_t.data), V)
 end
 
+function TensorKit.blocktype(::Type{<:CuTensorMap{T, S}}) where {T, S}
+    return CuMatrix{T, CUDA.DeviceMemory}
+end
+
```
lkdvos:
```suggestion
```
I think this is now more properly addressed through type inference.

src/tensors/abstracttensor.jl
```diff
@@ -103,11 +105,19 @@ similarstoragetype(X::Type, ::Type{T}) where {T <: Number} =
 
 # implement on tensors
 similarstoragetype(::Type{TT}) where {TT <: AbstractTensorMap} = similarstoragetype(storagetype(TT))
-similarstoragetype(::Type{TT}, ::Type{T}) where {TT <: AbstractTensorMap, T <: Number} =
-    similarstoragetype(storagetype(TT), T)
+function similarstoragetype(::Type{TT}, ::Type{T}) where {TT <: AbstractTensorMap, T <: Number}
+    return similarstoragetype(storagetype(TT), T)
+end
```
lkdvos, line 106: This is just a formatting change right?

src/tensors/indexmanipulations.jl
```diff
@@ -342,13 +344,14 @@ See also [`insertrightunit`](@ref insertrightunit(::AbstractTensorMap, ::Val{i})
 """
 function insertleftunit(
         t::AbstractTensorMap, ::Val{i} = Val(numind(t) + 1);
-        copy::Bool = false, conj::Bool = false, dual::Bool = false
+        copy::Bool = false, conj::Bool = false, dual::Bool = false,
+        storage_type::Type = storagetype(t)
```
lkdvos: This feels a bit strange to include a `storage_type` to me, where did this come up? Do we ever need the performance of both changing the storage type and adding unit spaces in one go? Can we bypass this with `adapt(TA, insertleftunit(args...))`?

In any case, the logic of the implementation seems to not fully take this into account for all cases either, but I would rather avoid introducing this.

src/tensors/indexmanipulations.jl
```diff
@@ -371,13 +374,14 @@ See also [`insertleftunit`](@ref insertleftunit(::AbstractTensorMap, ::Val{i}) w
 """
 function insertrightunit(
         t::AbstractTensorMap, ::Val{i} = Val(numind(t));
-        copy::Bool = false, conj::Bool = false, dual::Bool = false
+        copy::Bool = false, conj::Bool = false, dual::Bool = false,
+        storage_type::Type = storagetype(t),
```
lkdvos: Same comment here

src/tensors/indexmanipulations.jl
```diff
@@ -397,12 +401,12 @@ If `copy=false`, `tdst` might share data with `tsrc` whenever possible. Otherwis
 This operation undoes the work of [`insertleftunit`](@ref insertleftunit(::AbstractTensorMap, ::Val{i}) where {i}) 
 and [`insertrightunit`](@ref insertrightunit(::AbstractTensorMap, ::Val{i}) where {i}).
 """
-function removeunit(t::AbstractTensorMap, ::Val{i}; copy::Bool = false) where {i}
+function removeunit(t::AbstractTensorMap, ::Val{i}; copy::Bool = false, storage_type::Type = storagetype(t)) where {i}
```
lkdvos: Same comment here.

src/tensors/tensoroperations.jl
```diff
@@ -376,9 +376,10 @@ function blas_contract!(
         twistB = false
     end
 
+    TTC = storagetype(C)
```
lkdvos: I guess this effectively means that we are deciding to promote inputs to the storagetype of the output. I'm not sure if I am fully convinced that we should solve this automatically at all, since I think that is also inconsistent with how regular matrices work (same for adding):
```julia
julia> CUDA.rand(2, 2) * rand(Float32, 2, 2)
ERROR: Scalar indexing is disallowed.
```
I do think that this might be the right approach, and requiring explicit conversions in the cases of mixed inputs seems like the right call to me. (Even though I can see how that is annoying for MPSKit 😉 )

lkdvos (commented):
Left some comment throughout, there are some things that I am not entirely convinced by but the rest looks great, thanks for working through all of this!

For the `similarstoragetype(tensor, storagetype)` calls that you added, this seems like something we should probably discuss over a separate PR, and it would be great if we could consolidate this one to get the remainder of the fixes in.
Would you be up for splitting these two things, and then getting this merged?

The same kind of holds for some of the comments I made too, if we can just postpone the things that are not obvious, but already get the other parts in, that would probably be helpful.

(Note that I am very much aware that none of this is your fault and this PR has lived for too long so the design shifts a bit, for which I do apologize!)

kshyatt:
It's completely fine!! This has stayed open as I work through adding more tests for MPSKit, so I think we can pare off the simpler stuff we agree on, and then discuss things that are more contentious.

src/tensors/abstracttensor.jl
kshyatt: Yes, it's for the block case. I don't think we can have scalar unions?

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Let's remove it and see! 😈

src/tensors/indexmanipulations.jl
kshyatt: It's a bit of a kludge as I went through trying to get various MPO operations to work. Let's annihilate this and see what we can do without it now that the deeper storage type and promotion stuff is in.

src/tensors/abstracttensor.jl
lkdvos: It's a bit weird to support that here, since for generic `AbstractTensorMap` `eltype` would return a `Number`, which is why I asked about the scalartype thing. Maybe we can just copy this definition and overload in BlockTensorKit for `AbstractBlockTensorMap`?

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Yes I think in the end this turned out not to be needed, I'll pull it out

src/tensors/tensoroperations.jl
kshyatt: No, it's for 
```julia
    TTC = storagetype(C)                                                                                                                                                                                   
    # Bring A in the correct form for BLAS contraction                                                                                                                                                     
    if copyA                                                                                                                                                                                               
        Anew = TO.tensoralloc_add(TTC, A, pA, false, Val(true), allocator)                                                                                                                                 
        Anew = TO.tensoradd!(Anew, A, pA, false, One(), Zero(), backend, allocator)                                                                                                                        
        twistA && twist!(Anew, filter(!isdual ∘ Base.Fix1(space, Anew), domainind(Anew)))                                                                                                                  
    else                                                                                                                                                                                                   
        Anew = permute(A, pA)                                                                                                                                                                              
    end
```
Without this change, `Anew` will **always** have `Vector{scalartype(T)}` storage even if `A` was a `BraidingTensor` or some other object that only gets instantiated here. With the changes in https://github.com/QuantumKitHub/TensorKit.jl/pull/393 this won't be necessary. It's more than "annoying", with this change or #393 you have to define new `tensoralloc` methods for the mixed case, it's quite painful 😭

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: This is anyway handled by the BraidingTensor PR

src/tensors/braidingtensor.jl
lkdvos, line 1: I guess the changes here are no longer necessary since `BraidingTensor` now knows its own storagetype?

src/tensors/indexmanipulations.jl
```diff
@@ -342,11 +344,11 @@ See also [`insertrightunit`](@ref insertrightunit(::AbstractTensorMap, ::Val{i})
 """
 function insertleftunit(
         t::AbstractTensorMap, ::Val{i} = Val(numind(t) + 1);
-        copy::Bool = false, conj::Bool = false, dual::Bool = false
+        copy::Bool = false, conj::Bool = false, dual::Bool = false,
     ) where {i}
     W = insertleftunit(space(t), Val(i); conj, dual)
     if t isa TensorMap
-        return TensorMap{scalartype(t)}(copy ? Base.copy(t.data) : t.data, W)
+        return TensorMapWithStorage{scalartype(t), storagetype(t)}(copy ? Base.copy(t.data) : t.data, W)
```
lkdvos: This doesn't look necessary, isn't the storagetype automatically deduced from the provided `data` vector?

src/tensors/indexmanipulations.jl
```diff
@@ -371,11 +373,11 @@ See also [`insertleftunit`](@ref insertleftunit(::AbstractTensorMap, ::Val{i}) w
 """
 function insertrightunit(
         t::AbstractTensorMap, ::Val{i} = Val(numind(t));
-        copy::Bool = false, conj::Bool = false, dual::Bool = false
+        copy::Bool = false, conj::Bool = false, dual::Bool = false,
     ) where {i}
     W = insertrightunit(space(t), Val(i); conj, dual)
     if t isa TensorMap
-        return TensorMap{scalartype(t)}(copy ? Base.copy(t.data) : t.data, W)
+        return TensorMapWithStorage{scalartype(t), storagetype(t)}(copy ? Base.copy(t.data) : t.data, W)
```
lkdvos: Same comment here

src/tensors/indexmanipulations.jl
```diff
@@ -400,7 +402,7 @@ and [`insertrightunit`](@ref insertrightunit(::AbstractTensorMap, ::Val{i}) wher
 function removeunit(t::AbstractTensorMap, ::Val{i}; copy::Bool = false) where {i}
     W = removeunit(space(t), Val(i))
     if t isa TensorMap
-        return TensorMap{scalartype(t)}(copy ? Base.copy(t.data) : t.data, W)
+        return TensorMapWithStorage{scalartype(t), storagetype(t)}(copy ? Base.copy(t.data) : t.data, W)
```
lkdvos: Same comment here

src/tensors/braidingtensor.jl
kshyatt: I'm checking this now with `MPSKit`!

src/tensors/abstracttensor.jl
```diff
@@ -103,11 +101,19 @@ similarstoragetype(X::Type, ::Type{T}) where {T <: Number} =
...
+function similarstoragetype(::Type{<:AbstractTensorMap{T, S, N₁, N₂}}, ::Type{TA}) where {T <: Number, TA <: DenseVector, S, N₁, N₂}
+    return similarstoragetype(TA, T)
+end
+function similarstoragetype(t::AbstractTensorMap{T, S, N₁, N₂}, ::Type{TA}) where {T <: Number, TA <: DenseVector, S, N₁, N₂}
+    return similarstoragetype(typeof(t), TA)
+end
 
 # implement on arrays
 similarstoragetype(::Type{A}) where {A <: DenseVector{<:Number}} = A
+similarstoragetype(::Type{A}, ::Type{A}) where {A <: DenseVector{<:Number}} = A
```
lkdvos: Are we using these somewhere? I'm slightly confused about the meaning of these functions with last argument `A <: DenseVector`. In general, I thought this function mostly meant: for given tensor `T1` (and possibly `T2`), create a new storagetype that is compatible with `T1` but has scalartype `T`. In the case where we pass in a storagetype to begin with, doesn't that mean we already know that the answer has to be `A`?.

src/tensors/adjoint.jl
```diff
@@ -22,6 +22,8 @@ Base.adjoint(t::AbstractTensorMap) = AdjointTensorMap(t)
 space(t::AdjointTensorMap) = adjoint(space(parent(t)))
 dim(t::AdjointTensorMap) = dim(parent(t))
 storagetype(::Type{AdjointTensorMap{T, S, N₁, N₂, TT}}) where {T, S, N₁, N₂, TT} = storagetype(TT)
+similarstoragetype(::AdjointTensorMap{T, S, N₁, N₂, TT}, ::Type{T′}) where {T, S, N₁, N₂, TT, T′ <: Number} = similarstoragetype(TT, T′)
```
lkdvos: Isn't this also automatic from the definition of `similarstoragetype(t::AbstractTensorMap, T) = similarstoragetype(storagetype(t), T)` along with `storagetype` the line above?

src/tensors/adjoint.jl
```diff
@@ -22,6 +22,8 @@ Base.adjoint(t::AbstractTensorMap) = AdjointTensorMap(t)
...
+similarstoragetype(::AdjointTensorMap{T, S, N₁, N₂, TT}, ::Type{TA}) where {T, S, N₁, N₂, TT, TA <: DenseVector} = similarstoragetype(TT, TA)
```
lkdvos: Same comment here about why the second argument would already be a `<:DenseVector`.

src/tensors/indexmanipulations.jl
```diff
@@ -17,6 +17,8 @@ for (operation, manipulation) in (
         $promote_op(::Type{T}, ::Type{I}) where {T <: Number, I <: Sector} =
             sectorscalartype(I) <: Integer ? T :
             sectorscalartype(I) <: Real ? float(T) : complex(T)
+        $promote_op(::Type{TA}, ::Type{I}) where {TA <: DenseVector, I <: Sector} =
+            similarstoragetype(TA, $promote_op(eltype(TA), I))
```
lkdvos: This is also not really used I guess?

src/tensors/tensoroperations.jl
lkdvos: reopening this since it seems like you changed this again: I'm really confused about why you are claiming that `Anew` will have the wrong storagetype: it should have the same storage type as `A`, not as `C` no? It sounds like you are trying to make contractions with mismatching storagetypes in the inputs work, while I would have expected that this is not what we want to support? Am I missing something here, or am I wrong?

kshyatt: What seems to be happening is that:

- `TTC = scalartype(C)` gives you a `Float64` (let's say)
- `Anew = TO.tensoralloc_add(TTC, A, pA, false, Val(true), allocator)` then spits out something with storage type `Vector{Float64}` if `A` is a `SparseBlockMatrix{AbstractTensorMap}` even if it should in fact be giving you `Anew` having storage type `CuVector{Float64}` -- this is because of the fact that `AbstractTensorMap` storage type forcibly defaults to `Vector`, I think

So it's not that I'm **trying** to mix incompatible storage types, it's that TensorKit as currently set up forces me to do so without this change.

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -6,6 +6,9 @@ const AdjointCuTensorMap{T, S, N₁, N₂} = AdjointTensorMap{T, S, N₁, N₂,
 function CuTensorMap(t::TensorMap{T, S, N₁, N₂, A}) where {T, S, N₁, N₂, A}
     return CuTensorMap{T, S, N₁, N₂}(CuArray{T}(t.data), space(t))
 end
+function TensorMap{T, S, N₁, N₂, DA}(t::TensorMap{T, S, N₁, N₂, HA}) where {T, S, N₁, N₂, DA <: CuArray{T}, HA <: Array{T}}
+    return CuTensorMap{T, S, N₁, N₂}(CuArray{T}(t.data), space(t))
+end
```
lkdvos:
```suggestion
function TensorMap{T, S, N₁, N₂, A}(t::TensorMap) where {T, S, N₁, N₂, A}
    return TensorMap{T, S, N₁, N₂, A}(A(t.data), space(t))
end
```
Should we just add this as a generic constructor in the main package?

lkdvos (commented):
It seems like some of the rebasing and the github UI has made it hard to spot the comments I left before, although I think many of them are still unresolved and could be discussed :)

kshyatt:
I'm happy to discuss them, you'll note this PR is still in draft state.

src/tensors/braidingtensor.jl
kshyatt: This got completely detached from wherever it was supposed to be 🤔

lkdvos:
My bad, I got my notifications messed up and thought this was a review request 🙃 I will still blame github, but this one might actually also be (partially) on me

src/tensors/tensoroperations.jl
kshyatt: Does this make sense? I think the `JordanMPOTensor` changes could obviate this but I think we should link that issue in a comment and keep this moving

kshyatt:
GitHub UI against the world

src/tensors/braidingtensor.jl
```diff
@@ -294,11 +297,15 @@ function planarcontract!(
         backend, allocator
     )
     # special case only defined for contracting 2 indices
-    length(oindB) == length(cindB) == 2 ||
+    if !(length(oindB) == length(cindB) == 2)
+        # horrible!!!!!
+        tB′ = TensorMap(B)
+        tB = TensorMapWithStorage{eltype(B), similarstoragetype(A, eltype(B)), spacetype(tB′), numout(tB′), numin(tB′)}(tB′)
```
Jutho: Why does the following not work directly, and if so, can we fix that?


```suggestion
        tB = TensorMapWithStorage{eltype(B), similarstoragetype(A, eltype(B)), spacetype(B), numout(B), numin(B)}(B)
```

src/tensors/tensor.jl
```diff
@@ -34,6 +33,8 @@ struct TensorMap{T, S <: IndexSpace, N₁, N₂, A <: DenseVector{T}} <: Abstrac
         return new{T, S, N₁, N₂, A}(data, space)
     end
 end
+# constructors from another TensorMap -- no-op
+TensorMap{T, S, N₁, N₂, A}(t::TensorMap{T, S, N₁, N₂, A}) where {T, S <: IndexSpace, N₁, N₂, A <: DenseVector{T}} = t
```
Jutho: One could add Lukas' suggestion from the very top here:

```suggestion
TensorMap{T, S, N₁, N₂, A}(t::TensorMap{T, S, N₁, N₂, A}) where {T, S <: IndexSpace, N₁, N₂, A <: DenseVector{T}} = t
TensorMap{T, S, N₁, N₂, A}(t::TensorMap{T, S, N₁, N₂}) where {T, S <: IndexSpace, N₁, N₂, A <: DenseVector{T}} = TensorMap(A(t.data), space(t))
```

test/cuda/tensors.jl
```diff
@@ -293,29 +300,26 @@ for V in spacelist
                     ht2 = permute(TensorKit.to_cpu(t), (p1, p2))
                     @test ht2 ≈ TensorKit.to_cpu(dt2)
                 end
-
-                dt3 = CUDA.@allowscalar repartition(t, k)
+                dt3 = repartition(t, k)
                 ht3 = repartition(TensorKit.to_cpu(t), k)
                 @test ht3 ≈ TensorKit.to_cpu(dt3)
```
Jutho: Do we still have/support the `to_cpu` function? Above you replaced these calls with `adapt(Vector{T}, ...)`.

Jutho: And then there is also the shorthand `ad` that is sometimes being used.

Jutho: I find the `ad` quite non-descriptive, so maybe the definition of `ad` can be replaced with `to_cpu = adapt(Array)` and then this can be used consistently?

kshyatt: I think we can get rid of `to_cpu` indeed

src/tensors/tensor.jl
kshyatt: Odd or old?

Jutho: Was supposed to be "add". Apparently I cannot type anymore.

Jutho: I had to edit my message no less than 5 times to finally get it correct 😄 .

src/tensors/braidingtensor.jl
lkdvos: It was supposed to be with the entire file, given that braidingtensors now know about their storagetype, they should just be converted to tensors of their own storage type?

kshyatt: Ok, great, I think @Jutho also mentioned this below, can I resolve this and fold it into the open discussion there?

src/tensors/tensoroperations.jl
lkdvos: That does make sense, but that mostly seems to indicate that we should remove the `storagetype(::Type{<:AbstractTensorMap})` default, as that is just incorrect, and instead make `storagetype(x::SparseBlockTensorMap{AbstractTensorMap})` do a runtime computation of the storagetype when this isn't possible to deduce from the type itself.

kshyatt: Maybe so but does this have to happen in this PR?

kshyatt: It raises additional problems if the `SparseBlockTensorMap` or `BlockTensorMap` has empty `blocks`

src/tensors/braidingtensor.jl
lkdvos: Actually I think my comment is distinct, I think these specializations should never be required, because if `BraidingTensor` knows its storagetype, `TensorMap(::BraidingTensor)` should return a `TensorMap` of the correct storagetype, after which this should follow the same rules as regular `TensorMap` instances and indeed not work with mixed storagetypes. If mixed storagetypes still show up because of other parts of the code, we should either fixed that elsewhere, or make a decision to globally support mixed storagetype stuff (also for TensorMap), where I would lean towards the former.

src/tensors/indexmanipulations.jl
kshyatt: This is needed for `promote_permute`, in fact

src/tensors/tensor.jl
kshyatt: Added!

src/tensors/abstracttensor.jl
kshyatt: Lines 107-112 are needed for the `SparseBlockTensorMap` reasons outlined above. Line 116 is unnecessary and is now gone.

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -10,7 +10,7 @@ function MatrixAlgebraKit.findtruncated(
         fill!(v, dim(c))
     end
 
-    perm = sortperm(parent(values); strategy.by, strategy.rev)
+    perm = isempty(parent(values)) ? () : sortperm(parent(values); strategy.by, strategy.rev)
```
Jutho: I assume `sortperm` returns a type of `AbstractVector{Int}`, not sure if it is `Vector{Int}` or `CuVector{Int}` in this case. Is the problem that empty arrays fail? Should the `isempty` case return `(Cu)Vector{Int}(undef, 0)` to avoid a type stability?

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -36,7 +36,7 @@ function MatrixAlgebraKit.findtruncated(
         end
     end
 
-    perm = sortperm(parent(values); by = abs, rev = false)
+    perm = isempty(parent(values)) ? () : sortperm(parent(values); by = abs, rev = false)
```
Jutho: Same

src/tensors/braidingtensor.jl
```diff
@@ -265,8 +265,7 @@ function planarcontract!(
     )
     # special case only defined for contracting all 4 indices of B (2 contracted + 2 open)
     if length.(pB) != (2, 2)
-        tB′ = TensorMap(B)
-        tB = TensorMapWithStorage{eltype(B), similarstoragetype(A, eltype(B)), spacetype(tB′), numout(tB′), numin(tB′)}(tB′)
+        tB = TensorMapWithStorage{eltype(B), similarstoragetype(A, eltype(B)), spacetype(tB′), numout(tB′), numin(tB′)}(B)
```
Jutho: `tB'` is still referenced here, so this cannot be correct.

src/tensors/tensor.jl
Jutho: Odded!

ext/TensorKitCUDAExt/truncation.jl
kshyatt: The problem is deeper within GPUArrays, it tries to launch a kernel of size 0 which is not allowed

src/tensors/braidingtensor.jl
kshyatt: I think this is fixed by the latest change which also addressed Lukas' top level comment

ext/TensorKitCUDAExt/truncation.jl
kshyatt: @Jutho are you ok keeping this in? If not I can move it to another PR related to truncation fixes

Jutho: I am certainly ok with keeping it, I was just worried about the potential type stability. Why an empty tuple instead of an empty vector of the correct type?

kshyatt: Good catch, fixed

kshyatt:
GitHub UI acting up again. Does anyone have any objections to squashing this later today if tests pass?

src/tensors/braidingtensor.jl
```diff
@@ -189,10 +189,11 @@ end
 has_shared_permute(t::BraidingTensor, ::Index2Tuple) = false
 function add_transform!(
         tdst::AbstractTensorMap,
-        tsrc::BraidingTensor, (p₁, p₂)::Index2Tuple,
+        tsrc::BraidingTensor{T, S, A},
+        (p₁, p₂)::Index2Tuple,
         fusiontreetransform,
         α::Number, β::Number, backend::AbstractBackend...
-    )
+    ) where {T, S, A}
```
Jutho: Are these type parameters needed?

kshyatt: I think not, removed now

src/tensors/tensor.jl
```diff
@@ -34,6 +33,9 @@ struct TensorMap{T, S <: IndexSpace, N₁, N₂, A <: DenseVector{T}} <: Abstrac
...
```
```suggestion
```

I would actually be in favor of only having the second implementation, since there is a subtle semantic difference that I would like to retain:

In general, (although this is very poorly documented), the difference between a constructor and a converter is that a converter is meant to be as lossless as possible, and is allowed to take ownership of all parts of the original data. On the other hand, a constructor is typically meant to make an independent copy. You can see this for regular `Array`s in the example below:

Where it becomes a bit more vague is of course that you want to have a constructor that takes the different fields and builds up an object from that, without having to create an independent copy (e.g. the inner constructors of a type). This is also where the Julia docs are completely underspecified, so this is more something we should try and figure out ourselves.

I would say that in this case though, creating an independent copy is probably the safest solution, and it is probably easier to reason in generic code that if you call `TensorMap{...}(t::TensorMap)`, you get something that does not share data with `t`, but I am of course open to other opinions :)

```julia
julia> a = rand(3)
3-element Vector{Float64}:
 0.3155466339261911
 0.6265705423537413
 0.6035244268238111

julia> b = Array(a)
3-element Vector{Float64}:
 0.3155466339261911
 0.6265705423537413
 0.6035244268238111

julia> b[1] = 1
1

julia> a
3-element Vector{Float64}:
 0.3155466339261911
 0.6265705423537413
 0.6035244268238111

julia> c = convert(Array, a)
3-element Vector{Float64}:
 0.3155466339261911
 0.6265705423537413
 0.6035244268238111

julia> c[1] = 1
1

julia> a
3-element Vector{Float64}:
 1.0
 0.6265705423537413
 0.6035244268238111
```

src/tensors/indexmanipulations.jl
lkdvos: I don't think this is called anymore right?

src/tensors/indexmanipulations.jl
```diff
@@ -342,7 +344,7 @@ See also [`insertrightunit`](@ref insertrightunit(::AbstractTensorMap, ::Val{i})
 """
 function insertleftunit(
         t::AbstractTensorMap, ::Val{i} = Val(numind(t) + 1);
-        copy::Bool = false, conj::Bool = false, dual::Bool = false
+        copy::Bool = false, conj::Bool = false, dual::Bool = false,
```
```suggestion
        copy::Bool = false, conj::Bool = false, dual::Bool = false
```

src/tensors/indexmanipulations.jl
```diff
@@ -371,7 +373,7 @@ See also [`insertleftunit`](@ref insertleftunit(::AbstractTensorMap, ::Val{i}) w
 """
 function insertrightunit(
         t::AbstractTensorMap, ::Val{i} = Val(numind(t));
-        copy::Bool = false, conj::Bool = false, dual::Bool = false
+        copy::Bool = false, conj::Bool = false, dual::Bool = false,
```
```suggestion
        copy::Bool = false, conj::Bool = false, dual::Bool = false
```

src/tensors/indexmanipulations.jl
lkdvos: (Just trying to minimize the diff)

src/tensors/indexmanipulations.jl
kshyatt: Let me check again

kshyatt: There is a huge amount of this code that is tested only by MPSKit which makes verifying any of this... painful

src/tensors/diagonal.jl
```diff
@@ -280,7 +280,7 @@ end
 # ----------------
 function TO.tensoradd_type(TC, A::DiagonalTensorMap, ::Index2Tuple{1, 1}, ::Bool)
     M = similarstoragetype(A, TC)
-    return DiagonalTensorMap{TC, spacetype(A), M}
+    return DiagonalTensorMap{scalartype(M), spacetype(A), M}
```
lkdvos: I'm a bit torn by this, does this ever happen that the two don't match? I think we would like `similarstoragetype(A, TC) -> Type{<:DenseVector{TC}}` to always hold, and it might actually be nicer if this explicitly errors instead of silently giving a different result?

---

I just realized that this is probably because of trying to pass `TC <: Type{<:DenseVector}`, which I am not a big fan of. I think that is kind of misusing the TensorOperations interface slightly?

src/tensors/braidingtensor.jl
```diff
@@ -189,7 +189,8 @@ end
 has_shared_permute(t::BraidingTensor, ::Index2Tuple) = false
 function add_transform!(
         tdst::AbstractTensorMap,
-        tsrc::BraidingTensor, (p₁, p₂)::Index2Tuple,
+        tsrc::BraidingTensor,
+        (p₁, p₂)::Index2Tuple,
         fusiontreetransform,
```
lkdvos, line 193:
```suggestion
        tsrc::BraidingTensor, (p₁, p₂)::Index2Tuple,
```
Again just minimizing the diff

src/tensors/braidingtensor.jl
```diff
@@ -261,8 +262,9 @@ function planarcontract!(
         backend, allocator
     )
     # special case only defined for contracting all 4 indices of B (2 contracted + 2 open)
-    length.(pB) == (2, 2) ||
+    if length.(pB) != (2, 2)
         return planarcontract!(C, A, pA, TensorMap(B), pB, pAB, α, β, backend, allocator)
+    end
```
lkdvos:
```suggestion
    length.(pB) == (2, 2) ||
        planarcontract!(C, A, pA, TensorMap(B), pB, pAB, α, β, backend, allocator)


```

src/tensors/abstracttensor.jl
```diff
@@ -103,8 +101,15 @@ similarstoragetype(X::Type, ::Type{T}) where {T <: Number} =
...
-    similarstoragetype(storagetype(TT), T)
+function similarstoragetype(::Type{TT}, ::Type{T}) where {TT <: AbstractTensorMap, T <: Number}
+    return similarstoragetype(storagetype(TT), T)
+end
+function similarstoragetype(::Type{<:AbstractTensorMap{T, S, N₁, N₂}}, ::Type{TA}) where {T <: Number, TA <: DenseVector, S, N₁, N₂}
+    return similarstoragetype(TA, T)
+end
+function similarstoragetype(t::AbstractTensorMap{T, S, N₁, N₂}, ::Type{TA}) where {T <: Number, TA <: DenseVector, S, N₁, N₂}
+    return similarstoragetype(typeof(t), TA)
+end
```
lkdvos:
```suggestion
```

I think my comment got lost in the github UI, but these methods seem a bit strange to me. Since `similarstoragetype(t, T::Type{<:Number})` is supposed to give you a storagetype that is compatible with `t`, but has the scalartype `T`, I would infer that `similarastoragetype(t, A::Type{<:DenseVector})` should give you a storagetype that is compatible with storagetype `A`, which is simply `A` itself. Am I missing something here?

src/tensors/tensoroperations.jl
```diff
@@ -379,7 +379,7 @@ function blas_contract!(
     bstyle = BraidingStyle(sectortype(C))
     bstyle isa SymmetricBraiding ||
         throw(SectorMismatch("only tensors with symmetric braiding rules can be contracted; try `@planar` instead"))
-    TC = scalartype(C)
+    TC = storagetype(C) # without this, Anew below has wrong storagetype
```
lkdvos:
```suggestion
    TC = scalartype(C)
```

I understand that this is a way to make the issues go away, but I don't think this is really the right strategy. If `Anew` has the wrong storagetype, that means `A` had the wrong storagetype to begin with, and I would like to make `tensorcontract` just error in that case, since I don't think we have any support for mixed storagetypes at the moment?

For example, what `TensorOperations` does with the `cuTENSOR` backend is that the implementation first sends everything to the GPU: https://github.com/QuantumKitHub/TensorOperations.jl/blob/880d046330a665488f6592a6e6979fdd8af665ec/ext/TensorOperationscuTENSORExt.jl#L80-L91
I think I would be more in favor to have similar behavior, where if you want to run things on a GPU you should either make sure everything is on the GPU from the start, or explicitly select a backend that will copy everything over.

We might also just want to schedule a meeting about this and talk through the implications, since I do agree that there are advantages and disadvantages to making things work more automatic, and I might just be overly dense about this.
For the purpose of getting this PR merged ASAP though, I think it would be nice to just already get the smaller tweaks in, without having to make these decisions?

lkdvos (commented):
I left some comments, and tried to minimize some of the diff. I think overall this PR looks pretty good, and especially the tests are starting to look like they contain less and less `@allowscalar` and more idiomatic code so this is great work!

How would you feel about punting on the more in-depth design decisions, such as making the TensorOperations functions auto-promote storagetypes, and merging the rest?

src/tensors/indexmanipulations.jl
kshyatt: This is still needed from `tensoradd_type` which we discussed previously

kshyatt: Here's a trace for posterity

```
Error During Test at /home/kshyatt/.julia/dev/MPSKit/test/gpu/cuda/operators.jl:228                                                                                                                             
  Got exception outside of a @test                                                                                                                                                                              
  MethodError: no method matching promote_permute(::Type{Vector{ComplexF64}}, ::Type{TensorKitSectors.SU2Irrep})                                                                                                
  The function `promote_permute` exists, but no method is defined for this combination of argument types.                                                                                                       
                                                                                                                                                                                                                
  Closest candidates are:                                                                                                                                                                                       
    promote_permute(!Matched::Type{T}, ::Type{I}) where {T<:Number, I<:TensorKitSectors.Sector}                                                                                                                 
     @ TensorKit ~/.julia/dev/TensorKit/src/tensors/indexmanipulations.jl:17                                                                                                                                    
    promote_permute(!Matched::TensorKit.AbstractTensorMap)                                                                                                                                                      
     @ TensorKit ~/.julia/dev/TensorKit/src/tensors/indexmanipulations.jl:14                                                                                                                                    
    promote_permute(!Matched::Type{T}) where T<:TensorKit.AbstractTensorMap                                                                                                                                     
     @ TensorKit ~/.julia/dev/TensorKit/src/tensors/indexmanipulations.jl:15                                                                                                                                    
                                                                                                                                                                                                                
  Stacktrace:                                                                                                                                                                                                   
    [1] tensoradd_type(TC::Type, A::TensorKit.AdjointTensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 1, 2, TensorKit.TensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 2, 1, Vector{ComplexF64}}}, ::Tuple{Tuple{Int64, Int64}, Tuple{Int64}}, ::Bool)      @ TensorKit ~/.julia/dev/TensorKit/src/tensors/tensoroperations.jl:56                                                                                                                                         [2] tensoralloc_add(TC::Type, A::TensorKit.AdjointTensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 1, 2, TensorKit.TensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 2, 1, Vector{ComplexF64}}}, pA::Tuple{Tuple{Int64, Int64}, Tuple{Int64}}, conjA::Bool, istemp::Val{true}, allocator::TensorOperations.DefaultAllocator)                                                                                                                                       
      @ TensorOperations ~/.julia/packages/TensorOperations/FjqJj/src/implementation/allocator.jl:117                                                                                                           
    [3] blas_contract!(C::TensorKit.TensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 1, 2, Vector{ComplexF64}}, A::TensorKit.AdjointTensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 1, 2, TensorKit.TensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 2, 1, Vector{ComplexF64}}}, pA::Tuple{Tuple{Int64, Int64}, Tuple{Int64}}, B::TensorKit.TensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 1, 1, Vector{ComplexF64}}, pB::Tuple{Tuple{Int64}, Tuple{Int64}}, pAB::Tuple{Tuple{Int64}, Tuple{Int64, Int64}}, α::VectorInterface.One, β::VectorInterface.Zero, backend::TensorOperations.DefaultBackend, allocator::TensorOperations.DefaultAllocator)                                                                     
      @ TensorKit ~/.julia/dev/TensorKit/src/tensors/tensoroperations.jl:406                                                                                                                                    
    [4] contract!(C::TensorKit.TensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 1, 2, Vector{ComplexF64}}, A::TensorKit.TensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 1, 1, Vector{ComplexF64}}, pA::Tuple{Tuple{Int64}, Tuple{Int64}}, B::TensorKit.AdjointTensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 1, 2, TensorKit.TensorMap{ComplexF64, TensorKit.GradedSpace{TensorKitSectors.SU2Irrep, TensorKit.SortedVectorDict{TensorKitSectors.SU2Irrep, Int64}}, 2, 1, Vector{ComplexF64}}}, pB::Tuple{Tuple{Int64}, Tuple{Int64, Int64}}, pAB::Tuple{Tuple{Int64}, Tuple{Int64, Int64}}, α::VectorInterface.One, β::VectorInterface.Zero, backend::TensorOperations.DefaultBackend, allocator::TensorOperations.DefaultAllocator)                                                                          
      @ TensorKit ~/.julia/dev/TensorKit/src/tensors/tensoroperations.jl:346 
```

src/tensors/tensoroperations.jl
kshyatt: The problem is not having this breaks all of the more complex MPSKit support, which is why I added it in the first place.

kshyatt: Is reworking how all of this works right now worth forcing everyone in the group who wants to try GPU + MPSKit to wait (be realistic here) another 6 months?

src/tensors/abstracttensor.jl
kshyatt: Yes because the problem here is again the **AbstractTensorMap**, which has a bad default storagetype of `Vector`. This is again working around the other design decision about `JordanMPOTensor` and could be gotten rid of once that is changed.

kshyatt:
> How would you feel about punting on the more in-depth design decisions, such as making the TensorOperations functions auto-promote storagetypes, and merging the rest?

This will completely negate the point of this PR, which is to finally get things to actually work for MPSKit and also possibly break quite a few of the tests and force the return of `@allowscalar`.

src/tensors/tensoroperations.jl
kshyatt: All of the relevant arrays **are** already on the GPU but this is currently undetectable because of previously made decisions that obviously never took the concept of "different array storage locations" into account. My strong preference would be:

- Merge things more or less as is with a comment including a flag at each location I had to work around this awful `AbstractTensorMap` stuff so it is easy to find later
- Open an issue and call out the flag
- Fix the `JordanMPOTensor` behaviour that is causing it, without yet again forcing everyone who wants to use the GPU stuff to wait ages for an internal change
- Return to the previously flagged comments and fix things

Then at least we will have a working version to see if things break, and MPSKit can (_finally_) start moving again.

src/tensors/abstracttensor.jl
lkdvos: But who is **calling** this function?
Where do you end up with `similarstoragetype(AbstractTensorMap{...}, CuVector{T})` and if you want this to return `CuVector{T}`, why are you calling this function in the first place? Alternatively, if this has to go through the `promote_storagetype` tree, are we okay with hardcoding that `CuVector{T}` + `Vector{T}` promotes to `CuVector{T}`? I understand your point of view, but I'm not sure I want to make these decisions based only on this fixing a bad design choice in upstream packages.

For example, I'm assuming this is a result of `JordanMPOTensor` having a `Union` eltype, which then becomes a `AbstractTensorMap` eltype upon adding/permuting/...
Should we not instead try and make this result in the correct `TensorMap` eltype and tackle this code instead: https://github.com/QuantumKitHub/BlockTensorKit.jl/blob/5a6612600c705aaa8cef478ebcede6c8d26a32fe/src/tensors/abstractblocktensor/abstractarray.jl#L286-L305

kshyatt: Trying this over at https://github.com/QuantumKitHub/BlockTensorKit.jl/pull/54/changes/677d141f276ddf619f105829bcd1b9d7219a448c

src/tensors/tensoroperations.jl
kshyatt: OK, this now ???? seems ???? to be working with some changes that also worked in BlockTensorKit... let's see if the TK tests also pass

src/tensors/abstracttensor.jl
kshyatt: Removed!!!! 🔪 🔪 🔪

src/tensors/diagonal.jl
kshyatt: I'm running MPSKit tests again with some other changes but will check next whether this can now be removed

kshyatt: Looks good

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -36,14 +36,32 @@ function MatrixAlgebraKit.findtruncated(
...
 
     result = similar(values, Bool)
     parent(result)[perm] .= cumulative_err .> ϵᵖmax
     return result
 end
 
+function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::S) where {S <: MatrixAlgebraKit.TruncationStrategy}
+    # returning a CuSectorVector wrecks things in truncate_{co}domain
+    # because of scalar indexing
+    return CUDA.CUDACore.Adapt.adapt(Vector{Int}, MatrixAlgebraKit.findtruncated(values, strategy))
```
lkdvos: I don't think we are guaranteed to always return a `<:AbstractVector{Int}`, sometimes this can also be a `<:AbstractVector{Bool}`. Do you think we could put a hook into the `truncate_(co)domain` functions directly and then just specialize there?

kshyatt: Hm, could we instead split this into two lines and do 

```
raw_ind = MatrixAlgebraKit.findtruncated(values, strategy)
return adapt(Vector{eltype(raw_ind)}, raw_ind)
```

For whatever reason, `adapt(Vector, )` wasn't working :(

src/tensors/tensor.jl
kshyatt: I guess this and the adapt thing are the main outstanding points, right?

ext/TensorKitCUDAExt/truncation.jl
kshyatt: Seems to be working now, I guess the specialization issues were hiding this. Hooray!

kshyatt:
Think I addressed everything and cleaned up the diff a bit as well

kshyatt:
AMD test fail is unrelated. Does anyone have objections to this getting merged today?

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -10,7 +10,7 @@ function MatrixAlgebraKit.findtruncated(
         fill!(v, dim(c))
     end
 
-    perm = sortperm(parent(values); strategy.by, strategy.rev)
+    perm = isempty(parent(values)) ? Int64[] : sortperm(parent(values); strategy.by, strategy.rev)
```
lkdvos: I think this is now type-unstable for the GPU arrays: `sortperm` would otherwise return a `CuArray{Int, ...}`.
Probably this gets union split etc, but maybe the following is a bit nicer?


```suggestion
    isempty(parent(values)) && return similar(values, Bool)
    
    perm = sortperm(parent(values); strategy.by, strategy.rev)
```

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -36,14 +36,34 @@ function MatrixAlgebraKit.findtruncated(
         end
     end
 
-    perm = sortperm(parent(values); by = abs, rev = false)
+    perm = isempty(parent(values)) ? Int64[] : sortperm(parent(values); by = abs, rev = false)
```
lkdvos:
```suggestion
    isempty(parent(values)) && return similar(values, Bool)
    
    perm = sortperm(parent(values); by = abs, rev = false)
```

src/tensors/abstracttensor.jl
```diff
@@ -53,9 +53,7 @@ storagetype(t) = storagetype(typeof(t))
 function storagetype(::Type{T}) where {T <: AbstractTensorMap}
     if T isa Union
         # attempt to be slightly more specific by promoting unions
-        Ma = storagetype(T.a)
-        Mb = storagetype(T.b)
-        return promote_storagetype(Ma, Mb)
+        return promote_storagetype(T.a, T.b)
     else
         # fallback definition by using scalartype
         return similarstoragetype(scalartype(T))
```
lkdvos, line 59: Do we want to immediately return `AbstractVector{T}` here?

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -36,14 +36,34 @@ function MatrixAlgebraKit.findtruncated(
...
+        return CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated(values, strategy))
+    end
+end
+
+function MatrixAlgebraKit.findtruncated_svd(values::CuSectorVector, strategy::MatrixAlgebraKit.TruncationByValue)
+    atol = TensorKit.Factorizations.rtol_to_atol(values, strategy.p, strategy.atol, strategy.rtol)
+    strategy′ = trunctol(; atol, strategy.by, strategy.keep_below)
+    return SectorDict(c => CUDA.CUDACore.Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated_svd(d, strategy′)) for (c, d) in pairs(values))
+end
+
```
lkdvos, line 70: Would it make sense to overload the `truncate_domain!` and `truncate_codomain!` and `truncate_diagonal!` functions instead?
This looks like it is quite prone to ambiguity, and I guess we will also have to copy this for the `findtruncated` version too if we want to have eigenvalue decompositions as well.


```suggestion
function TensorKit.Factorizations.truncate_domain!(tdst::CuTensorMap, tsrc::CuTensorMap, inds)
    for (c, b) in blocks(tdst)
        I = get(inds, c, nothing)
        @assert !isnothing(I)
        I = CUDA.CUDACore.Adapt.adapt(Vector, I)
        b′ = block(tsrc, c)
        b .= view(b′, :, I)
    end
    return tdst
end
function TensorKit.Factorizations.truncate_codomain!(tdst::CuTensorMap, tsrc::CuTensorMap, inds)
    for (c, b) in blocks(tdst)
        I = get(inds, c, nothing)
        @assert !isnothing(I)
        I = CUDA.CUDACore.Adapt.adapt(Vector, I)
        b′ = block(tsrc, c)
        b .= view(b′, I, :)
    end
    return tdst
end
function TensorKit.Factorizations.truncate_diagonal!(Ddst::DiagonalCuTensorMap, Dsrc::DiagonalCuTensorMap, inds)
    for (c, b) in blocks(Ddst)
        I = get(inds, c, nothing)
        @assert !isnothing(I)
        I = CUDA.CUDACore.Adapt.adapt(Vector, I)
        diagview(b) .= view(diagview(block(Dsrc, c)), I)
    end
    return Ddst
end
```

(Warning, did not try to run this!)
Also, should this be `adapt` or `collect`?
Also, I added the `DiagonalCuTensorMap`, not sure if we have that type alias yet. (And also if this is required for them?)

ext/TensorKitCUDAExt/truncation.jl
```diff
@@ -36,14 +36,34 @@ function MatrixAlgebraKit.findtruncated(
...
 # Needed until MatrixAlgebraKit patch hits...
 function MatrixAlgebraKit._ind_intersect(A::CuVector{Bool}, B::CuVector{Int})
```
lkdvos, line 72: I think we should get rid of this one now?

ext/TensorKitCUDAExt/truncation.jl
kshyatt: I tried overriding those and the problem was I was getting compilation errors from GPUCompiler, it was enough of a rabbit hole that I thought it made more sense to punt this for now

ext/TensorKitCUDAExt/truncation.jl
kshyatt: This one is the one with `# Needed until MatrixAlgebraKit patch hits...` ?

ext/TensorKitCUDAExt/truncation.jl
kshyatt: Sorry, why `Bool` here? If `sortperm` would return `CuVector{Int}`, I think `Bool` is only one byte but `Int` is (usually) 8, no?

src/tensors/abstracttensor.jl
kshyatt: here = `similarstoragetype(scalartype(T))`?

lkdvos: yes, wow

ext/TensorKitCUDAExt/truncation.jl
lkdvos: Because if the `parent` is empty, we don't have to do any of the logic below, and can immediately return an empty `SectorVector{Bool}`, which is what would have been returned after the logic below. (I.e., I'm not replacing the `sortperm` output, just bypassing the whole function)

src/tensors/abstracttensor.jl
kshyatt: Oh, I didn't want to hard code it in case we change the way `similarstoragetype(::Number)` works in the future (for example, allowing users to set the default)

ext/TensorKitCUDAExt/truncation.jl
lkdvos: yep, sorry for that

kshyatt: Mind if I do that in a separate PR?

kshyatt: https://github.com/QuantumKitHub/TensorKit.jl/pull/435

lkdvos (dismissed):
I think overall the current changes look good to me. The only remaining part is the one on the factorizations, but if this unblocks it, we can always revisit if this shows up in profilers.

kshyatt:
Sorry by factorizations u mean `truncate_{co}domain`?

lkdvos (approved):

PR #379, opened 2026-03-11
Update indexmanipulations.jl

kshyatt:
Remove unneeded unit op (I think?)

I think this `tsdt` isn't actually used so we can remove some unnecessary work

### DIFF
--- a/ext/TensorKitMooncakeExt/indexmanipulations.jl
+++ b/ext/TensorKitMooncakeExt/indexmanipulations.jl
@@ -283,7 +283,7 @@ for insertunit in (:insertleftunit, :insertrightunit)
                 tsrc_cache = copy(tsrc)
                 tdst_Δtdst = CoDual(
                     $insertunit(tsrc, ival; kwargs...),
-                    $insertunit(Mooncake.tangent(tsrc_Δtsrc), ival; kwargs...)
+                    $insertunit(Δtsrc, ival; kwargs...)
                 )
             else
                 tsrc_cache = nothing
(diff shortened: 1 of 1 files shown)


lkdvos (dismissed):
Wooops good catch!

ext/TensorKitMooncakeExt/indexmanipulations.jl
```diff
@@ -281,7 +281,6 @@ for insertunit in (:insertleftunit, :insertrightunit)
             # sharing address spaces
             if tsrc isa TensorMap && !get(kwargs, :copy, false)
                 tsrc_cache = copy(tsrc)
-                tdst = $insertunit(tsrc, ival; kwargs...)
                 tdst_Δtdst = CoDual(
                     $insertunit(tsrc, ival; kwargs...),
                     $insertunit(Mooncake.tangent(tsrc_Δtsrc), ival; kwargs...)
```
Jutho: Should this also be
```suggestion
                    $insertunit(Δtsrc, ival; kwargs...)
```

I am still a bit confused about when `arrify` needs to be called and when not. Is `Mooncake.tangent` calling `arrify`?

kshyatt: `arrayify` does some reinterpretation wizardry on an array of tuples to give us both the primal and tangent arrays. tangent just gives us the "raw" tangent array, which may have named tuple elements. So now with the changes to make Complex-valued arrays "first class", we don't need arrayify here... but if we support BigFloats or Quaternions or something odd later, we might.

Jutho: Is my suggestion then not the more robust solution?

lkdvos: Yes, I think here we really do want the `arrayify`'d version, for example for `AdjointTensorMap` this would be necessary, (even though this is in a `tsrc isa TensorMap` guarded-if-branch).
For consistency though, I think Jutho's suggestion just looks nicest

lkdvos (approved):

Jutho (approved):

PR #380, opened 2026-03-12
Incremental work on Enzyme support

kshyatt:
Building on top of https://github.com/QuantumKitHub/TensorKit.jl/pull/356

First I need to get all the tests working. I am getting some odd space mismatch errors I need to dig into further as well as a few Enzyme-related segfaults.

Then, I'd like to refactor the joint code shared between Mooncake and Enzyme into the main `src` folder so it can be reused (much as we've done with MatrixAlgebraKit).


kshyatt:
We're getting kind of hammered here by the fact that Enzyme seems to be compiling many many different versions of [this line](https://github.com/EnzymeAD/Enzyme.jl/blob/main/lib/EnzymeCore/src/rules.jl#L351) in the tests -- probably exacerbated by the fact that we have a lot of different types running around in TensorKit.

kshyatt:
OK I gotta figure out why the Enzyme tests take so freaking long in compilation, this is absurd

kshyatt:
The `eig_vals`, `eigh_vals`, and `svd_vals` are wrong for `SU2Irrep` and `FibonacciAnyon` which is rather weird, I'll look into this more.

kshyatt:
The factorization tests are passing locally... why do these tests worship death

kshyatt:
I think everything here has been pulled out so this can be closed

PR #381, opened 2026-03-13
Increase code and test coverage

borisdevos:
This PR achieves multiple things:

- Increase code coverage in the symmetry-related code by testing `A4Irrep`, currently the only available sector type in TensorKitSectors with `GenericFusion`.
- Also test `ZNElement` sector types which have `UniqueFusion` but non-trivial F-symbols, revealing an error in `foldright`.
- Remove some functions which are unused, even in downstream packages AFAIK.

I started on this a while ago, but refrained from progressing because #261 seemed to be picking up speed, but since then it's remained unmerged, so I thought to just finish this and not have to worry about some of the issues we knew were present. 

Edit: after #261 and #389 this PR does less than summed above :)


test/setup.jl
```diff
@@ -145,14 +145,30 @@ function test_dim_isapprox(V::ProductSpace, d::Int)
...
+multifusionsectorlist = (
     IsingBimodule, IsingBimodule ⊠ SU2Irrep, IsingBimodule ⊠ IsingBimodule,
+    IsingBimodule ⊠ Z3Element{1}, IsingBimodule ⊠ FibonacciAnyon ⊠ A4Irrep,
+)
+
+sectorlist = (
+    uniquefusionsectorlist...,
+    simplefusionsectorlist...,
+    genericfusionsectorlist...,
+    multifusionsectorlist...,
```
Jutho, line 189: Did this come from the ongoing vectorize fusiontree PR? I hope that if this gets merged first, a rebase of the other PR will allow to combine this without too many conflicts...

borisdevos: It's indeed a copy-paste, but I think I may have added a sector or two

lkdvos:
hoping that the merge commit will fix the faulty actions as well...

src/TensorKit.jl
```diff
@@ -42,7 +43,8 @@ export infimum, supremum, isisomorphic, ismonomorphic, isepimorphic
 # Reexport methods for sectors and properties thereof
 export sectortype, sectors, hassector
 export unit, rightunit, leftunit, allunits, isunit, otimes
-export Nsymbol, Fsymbol, Rsymbol, Bsymbol, frobenius_schur_phase, frobenius_schur_indicator, twist, sectorscalartype, deligneproduct
+export Nsymbol, Fsymbol, Rsymbol, Bsymbol, frobenius_schur_phase, frobenius_schur_indicator, twist, fusiontensor
+export sectorscalartype, fusionscalartype, braidingscalartype, deligneproduct
```
Jutho: For consistency (which wasn't there before), I feel like `deligneproduct` should be two lines up, on the same line of `unit` and `otimes`, as it is mapping sectors to sectors. The next line is mapping sectors to data, and then this final line is about `scalartype`.

Jutho (dismissed):

borisdevos:
Compat check is still failing so I can't merge

PR #383, opened 2026-03-18
Small simplification of indexmanipulations

lkdvos:
This PR removes a (possibly micro-)optimization that bypassed the Strided.jl machinery for a small gain in performance. Here I removed that part again since it severely complicates things for the GPU implementations.
It's also safer in the sense that the current implementation was only correct since it was implicitly never called on `conj`'d `StridedViews`, so I do think that in the long term this is more maintainable anyways.

I will try and see if I can run the benchmark suite to actually gauge if there are any measurable performance implications here, but I would guess that this is a reasonable trade-off.

### DIFF
--- a/src/auxiliary/auxiliary.jl
+++ b/src/auxiliary/auxiliary.jl
@@ -57,30 +57,6 @@ function _interleave(a::NTuple{N}, b::NTuple{N}) where {N}
     return (a[1], b[1], _interleave(tail(a), tail(b))...)
 end
 
-# Low-overhead implementation of `copyto!` for specific case of `stride(B, 1) < stride(B, 2)`
-# used in indexmanipulations: avoids the overhead of Strided.jl
-function _copyto!(A::StridedView{<:Any, 1}, B::StridedView{<:Any, 2})
-    length(A) == length(B) || throw(DimensionMismatch())
-
-    Adata = parent(A)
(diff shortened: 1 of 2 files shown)


lkdvos:
See also https://github.com/QuantumKitHub/Strided.jl/pull/52

kshyatt (approved):

Jutho:
I had never even noticed when this was introduced (or had forgotten about it). Was this the reason for the special `copyto!` implementation https://github.com/QuantumKitHub/Strided.jl/pull/52 ?

kshyatt:
@Jutho It's the reason for one of them, yes!

lkdvos:
Feel free to merge without the benchmarkresults, I'm pretty sure it's not realistic for us to maintain things like this anyways.

Jutho:
@kshyatt , what is the other one?

Jutho (approved):

kshyatt:
@Jutho It's in the logic for `flip`! The offending line is [here](https://github.com/QuantumKitHub/TensorKit.jl/blob/main/src/tensors/indexmanipulations.jl#L44)

kshyatt:
With the `copyto!` + `Broadcasted` changes in the `Strided.jl` PR, we can avoid scalar indexing for all the `flip` operations 🤠

PR #387, opened 2026-03-23
Reduce cache footprint by decoupling degeneracy-dependent data

### LINKED ISSUE #384: `fusionblockstructure` should reuse data that is degeneracy-independent
Within the cached data for `fusionblockstructure`, there is data that depends on the sectors only, and data that depends on the sectors and degeneracies.
Decoupling these should decrease the memory pressure and increase the overall speed of the package, since we could avoid recomputing fusiontrees in workflows where the degeneracies are often fluctuating slightly, but the charges are somewhat fixed.

lkdvos:
Refactors the internal block structure representation for `TensorMap` by splitting `FusionBlockStructure` into two separately-cached structs:

- **`SectorStructure`** (cached by sector structure, shared across spaces with the same sectors): coupled sectors and valid fusion tree pairs as `Dictionaries.Indices`.
- **`DegeneracyStructure`** (cached per `HomSpace`): total dimension, block sizes/ranges, and sub-block strides as plain `Vector`s.

The public `blockstructure(W)` and `subblockstructure(W)` combine these on the fly into a `Dictionary`.

Previously, `FusionBlockStructure` stored a `fusiontreelist`, a `fusiontreeindices` lookup dict, and a `fusiontreestructure` vector — the tree pairs were stored twice and lookup required manual indirection. `Dictionaries.jl` handles this naturally via its token system.

A secondary benefit: `DegeneracyStructure{N}` is parameterized only by the number of indices `N`, not the sector type. Kernels that only need sizes/strides/offsets can therefore be compiled once and reused across different symmetry groups — a potential route to much faster precompilation.

Also moves the tensor structure computation out of `homspace.jl` into a new `tensorstructure.jl`.

---

**Open questions:**
- Should we replace all remaining internal dicts with `Dictionaries.jl` types to avoid maintaining two APIs?
- Are there other places that would benefit from a similar sector/degeneracy split?


src/auxiliary/dicts.jl
```diff
@@ -263,3 +263,24 @@ function Base.:(==)(d1::SortedVectorDict, d2::SortedVectorDict)
...
+struct Hashed{T, Hash, Eq}
+    val::T
+end
+
+Hashed(val, hash = Base.hash, eq = Base.isequal) = Hashed{typeof(val), hash, eq}(val)
+
+Base.parent(h::Hashed) = h.val
+
+# hash overload
+Base.hash(h::Hashed{T, Hash}, seed::UInt) where {T, Hash} = Hash(parent(h), seed)
```
Jutho: I find it a bit confusing if type parameters encode a function, that they still use capital letters. I was looking for a type called `Hash`, before realising that this is actually the custom hash function encoded in the type. I understand that using `hash` for the type parameter is out of the question.

A related question is whether you know if there is any benefits or downsides of this design compared to using the function as a field, i.e.
```julia
struct Hashed{T, H, E}
    val::T
    hashf::H
    eqf::E
end
```
and then doing things like `h.hashf(parent(h), seed)`. I am asking because that is how [`InnerProductVec`](https://github.com/Jutho/KrylovKit.jl/blob/master/src/innerproductvec.jl) over at KrylovKit works, and if it has downsides (other than the stylistic differences), I should probably change this.

lkdvos: I actually don't think it does, if these are singleton types (like `Function`) the size remains the same, so AFAIK this is equivalent and just a style choice. I'll refactor along with the required rebase :)

src/tensors/tensorstructure.jl
```diff
@@ -0,0 +1,276 @@
+# Block and fusion tree ranges: structure information for building tensors
+#--------------------------------------------------------------------------
```
Jutho: Even though some of this information is very much tailored towards the tensor construction, in my mind spaces come before tensors, and could in principle exist on their own. However, the current code in the spaces subfolder is broken without the definitions in this file, so I would prefer this file to live in the spaces folder. (There is not a single mention of `TensorMap` in this file).

src/tensors/tensorstructure.jl
```diff
@@ -0,0 +1,276 @@
...
+function sectorequal(W₁::HomSpace, W₂::HomSpace)
+    check_spacetype(W₁, W₂)
+    (numout(W₁) == numout(W₂) && numin(W₁) == numin(W₂)) || return false
+    for (w₁, w₂) in zip(codomain(W₁), codomain(W₂))
+        isdual(w₁) == isdual(w₂) || return false
+        isequal(sectors(w₁), sectors(w₂)) || return false
+    end
+    for (w₁, w₂) in zip(domain(W₁), domain(W₂))
+        isdual(w₁) == isdual(w₂) || return false
+        isequal(sectors(w₁), sectors(w₂)) || return false
```
Jutho: I don't think the return value of `sectors` was really intended for direct comparison:

```julia-repl
julia> isequal(sectors(SU2Space(0=>2)), sectors(SU2Space(0=>2)))
false
```

Jutho: Maybe `isempty(symdiff(sectors(w₁), sectors(w₂)))` ?

src/spaces/productspace.jl
```diff
@@ -147,23 +150,11 @@ that make up the `ProductSpace` instance.
...
-        for s in sectors(P)
-            for c in ⊗(s...)
-                if !(c in bs)
-                    push!(bs, c)
-                end
-            end
-        end
+    bs = Indices{I}()
+    for s in sectors(P), c in ⊗(s...)
+        set!(bs, c)
```
Jutho: I don't think this covers the `N==0` case, and is also overly complicated for the `N==1` case.

src/spaces/structure.jl
```diff
@@ -0,0 +1,276 @@
...
+                n₁ = length(trees) - codom_start + 1
+            else
+                # Subsequent f₂s: the codomain trees are already in the list at
+                # codom_start:codom_start+n₁-1, so read them back instead of recomputing.
+                for j in codom_start:(codom_start + n₁ - 1)
+                    push!(trees, (trees[j][1], f₂))
+                end
+            end
+        end
+    end
```
Jutho, line 69: I'd say there are quite a few unnecessary +1 and -1s because of your choice of using the starting index, instead of the offset:
```suggestion
    for c in _blocksectors(W)
        push!(bs, c)
        offset = length(trees)
        n₁ = 0
        for f₂ in fusiontrees(dom, c)
            if n₁ == 0
                # First f₂ for this sector: enumerate codomain trees and record how many there are.
                for f₁ in fusiontrees(codom, c)
                    push!(trees, (f₁, f₂))
                end
                n₁ = length(trees) - offset
            else
                # Subsequent f₂s: the codomain trees are already in the list at
                # offset .+ (1:n₁), so read them back instead of recomputing.
                for j in offset .+ (1:n₁)
                    push!(trees, (trees[j][1], f₂))
                end
            end
        end
    end
```

src/tensors/tensorstructure.jl
lkdvos: `issetequal` is probably the correct abstraction in this case, although it would be nice to be able to avoid this (the resulting sectorset is really only there because it needs to account for duals, but we don't need that since it is already checked). I'll think a bit more about this, and just factor it out into a function so we can specialize later if needed.

Jutho: Since sectors are sorted I think we probably just want to `collect` them and use normal array comparison. Quickly glancing over the implementation, all the more fancy set methods like `issetequal` and `symdiff` will anyway first call `Set` on them, which will also collect them.

src/tensors/tensorstructure.jl
lkdvos: I now moved a bunch of the functions around, let me know if this feels more natural!

src/spaces/structure.jl
```diff
@@ -0,0 +1,198 @@
...
+        f₁, f₂ = gettokenvalue(treelist, tree_index)
+        c = f₁.coupled
+
+        # compute subblock structure
+        # splitting tree data
+        empty!(splittingstructure)
+        offset₁ = 0
+        for i in tree_index:L
+            f₁′, f₂′ = gettokenvalue(treelist, i)
+            f₂′ == f₂ || break
```
Jutho, line 141: I am confused by this. We have
```julia
f₁, f₂ = gettokenvalue(treelist, tree_index)
```
and then the first value of `i in tree_index:L` is `i == tree_index`, and then we have
```julia
f₁′, f₂′ = gettokenvalue(treelist, i)
f₂′ == f₂ || break
```
Doesn't this loop break straight away in the first iteration?

Jutho: Oh, I misinterpreted this. You want `f₂′ == f₂`. Never mind the above.

src/spaces/vectorspaces.jl
```diff
@@ -363,6 +363,18 @@ Return an iterator over the different sectors of `V`.
 """
 function sectors end
 
+function sectorequal(V₁::ElementarySpace, V₂::ElementarySpace)
+    isdual(V₁) == isdual(V₂) || return false
+    return issetequal(sectors(V₁), sectors(V₂))
```
Jutho, line 368: Since `sectorhash` does in principle depend on the order in which the sectors are iterated (which is fine, since they are anyway ordered), it does perhaps make sense to make sure that `sectorequal` is fully consistent and also explicitly iterates through the sectors. Then it can also be made completely allocation free, but is of course a bit more code. Since `length(sectors(...))` doesn't generally work, I can also not use `zip` so I found no better solution then writing out the iteration exaclty
```suggestion
    s₁, s₂ = sectors(V₁), sectors(V₂)
    next₁, next₂ = iterate(s₁), iterate(s₂)
    while !isnothing(next₁) && !isnothing(next₂)
        val₁, state₁ = next₁
        val₂, state₂ = next₂
        val₁ == val₂ || return false
    end
    return isnothing(next₁) && isnothing(next₂)
```

lkdvos: While I'm okay with this, another approach could be to simply not define this fallback definition at all. I have added specializations for our specific GradedSpace implementations, so I think this method currently only ends up being called with Cartesian or complex spaces, for which I probably should also just overload this. In the end, this code would then never be called and it might be better to just have a methoderror if we end up accidentally changing something in the future?

src/tensors/treetransformers.jl
Jutho, line 1: Since I still even have to review this file from the vectorize fusiontrees PR, I am skipping it now. I will get to it at some point 😄 .

src/tensors/braidingtensor.jl
```diff
@@ -165,10 +165,9 @@ function block(b::BraidingTensor, s::Sector)
...
     end
 
-    structure = fusionblockstructure(b)
-    base_offset = first(structure.blockstructure[s][2]) - 1
+    base_offset = first(blockstructure(b)[s][2]) - 1
 
-    for ((f₁, f₂), (sz, str, off)) in zip(structure.fusiontreelist, structure.fusiontreestructure)
+    for ((f₁, f₂), (sz, str, off)) in pairs(subblockstructure(space(b)))
         (f₁.coupled == f₂.coupled == s) || continue
         r = _braiding_factor(f₁, f₂)
```
Jutho, line 172: This is not part of this PR, but I am wondering why this call to `_braiding_factor` does not include the information in `b.adjoint`, whereas the subblock call (getindex with fusion trees) does?

src/spaces/vectorspaces.jl
Jutho: Yes that would definitely be ok with me.

src/tensors/braidingtensor.jl
lkdvos: That seems to be an error on my behalf, great catch! I'll fix in a separate PR and rebase this.

Jutho (approved):
JET breaking on nightly seems to be still a thing; I thought we disabled JET on nightly, or all-together (?), but that was probably in MAK?

Anyway, PR approved.

PR #390, opened 2026-03-25
MAK v0.6.5 updates

lkdvos:
This PR consists of some small fixes to unbreak the tests.
None of these should affect downstream users, as the ambiguity only hits a new feature, but it is required to make the tests of all current PRs pass.

The CUDA arrays where wrong `blocktype` implementations, which made the algorithm selection trip. This was already tackled in #375, but I handled it slightly differently here by just making the overload more specific, to avoid accidentally running into this in the future. It does mean we rely a bit more on type inference for that, but I think this is a reasonable trade-off.


src/factorizations/truncation.jl
```diff
@@ -281,6 +281,22 @@ function MAK.findtruncated_svd(values::SectorVector, strategy::TruncationInterse
...
     )
 end
+function MAK.findtruncated(values::SectorVector, strategy::TruncationUnion)
+    inds = map(Base.Fix1(MAK.findtruncated, values), strategy.components)
+    return SectorDict(
+        c => reduce(
+                MatrixAlgebraKit._ind_union, [ind[c] for ind in inds if haskey(ind, c)]
+            ) for c in union(map(keys, inds)...)
+    )
+end
```
Jutho, line 298: I am not a big fan of these complicated nested expressions with generators; I would be surprised if that is type stable.

```suggestion
function MAK.findtruncated(values::SectorVector, strategy::TruncationUnion)
    inds = map(Base.Fix1(MAK.findtruncated, values), strategy.components)
    cs = union(map(keys, inds)...)
    union_ind = let inds = inds, empty_ind = Int[], _union = MatrixAlgebraKit._ind_union
        (c => mapreduce(ind->get(ind, c, empty_ind), _union, inds) for c in cs)
    end
    return SectorDict(union_ind)
end
```

src/factorizations/truncation.jl
```diff
@@ -281,6 +281,22 @@ function MAK.findtruncated_svd(values::SectorVector, strategy::TruncationInterse
...
+function MAK.findtruncated_svd(values::SectorVector, strategy::TruncationUnion)
+    inds = map(Base.Fix1(MAK.findtruncated_svd, values), strategy.components)
+    return SectorDict(
+        c => reduce(
+                MatrixAlgebraKit._ind_union, [ind[c] for ind in inds if haskey(ind, c)]
+            ) for c in union(map(keys, inds)...)
+    )
+end
```
Jutho, line 307: Same here:
```suggestion
function MAK.findtruncated_svd(values::SectorVector, strategy::TruncationUnion)
    inds = map(Base.Fix1(MAK.findtruncated_svd, values), strategy.components)
    cs = union(map(keys, inds)...)
    union_ind = let inds = inds, empty_ind = Int[], _union = MatrixAlgebraKit._ind_union
        (c => mapreduce(ind->get(ind, c, empty_ind), _union, inds) for c in cs)
    end
    return SectorDict(union_ind)
end
```

Jutho (commented):
Two code suggestions but otherwise approved!

lkdvos:
Test failures are due to `allequal` not existing on 1.12, I will update this but let the tests finish first

PR #389, opened 2026-03-24
Overhaul test infrastructure

lkdvos:
## Summary

- Moves test dependencies into a dedicated `test/Project.toml`, separating them from the main package manifest
- Replaces the monolithic test runner with [ParallelTestRunner.jl](https://github.com/vchuravy/ParallelTestRunner.jl), running each test file in its own worker process
- Adds `--fast` mode that skips AD test groups and reduces sector/scalar-type coverage for quick iteration
- Adds `test/README.md` documenting how to run tests, available groups, fast mode, and how to add new test files
- Updates CI to auto-discover test groups as matrix jobs; draft PRs run a reduced matrix (ubuntu + Julia 1 only) while ready PRs run the full matrix

PR #393, opened 2026-04-03
Allow BraidingTensor to have a custom storage type

kshyatt:
This is probably the nicest way to unblock the various MPSKit changes. I found the `similarmatrixtype` approach the least gross looking way to achieve this, but open to other ideas.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "TensorKit"
 uuid = "07d1fe3e-3e46-537d-9eac-e9e13d0d4cec"
-version = "0.16.4"
+version = "0.17.0"
 authors = ["Jutho Haegeman, Lukas Devos"]
 
 [deps]
--- a/ext/TensorKitAdaptExt.jl
+++ b/ext/TensorKitAdaptExt.jl
@@ -15,8 +15,12 @@ function Adapt.adapt_structure(to, x::DiagonalTensorMap)
     data′ = adapt(to, x.data)
     return DiagonalTensorMap(data′, x.domain)
 end
-function Adapt.adapt_structure(::Type{TorA}, x::BraidingTensor) where {TorA <: Union{Number, DenseArray{<:Number}}}
-    return BraidingTensor{scalartype(TorA)}(space(x), x.adjoint)
+function Adapt.adapt_structure(::Type{T}, x::BraidingTensor{T′, S, A}) where {T <: Number, T′, S, A}
+    A′ = TensorKit.similarstoragetype(A, T)
+    return BraidingTensor{T, S, A′}(space(x), x.adjoint)
+end
+function Adapt.adapt_structure(::Type{TA}, x::BraidingTensor{T, S, A}) where {T′, TA <: DenseArray{T′}, T, S, A}
+    return BraidingTensor{T′, S, TA}(space(x), x.adjoint)
 end
 
 end
--- a/ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
+++ b/ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
@@ -22,11 +22,14 @@ include("cutensormap.jl")
 include("truncation.jl")
 
 function TensorKit.fill_braidingsubblock!(data::TD, val) where {T, TD <: Union{<:CuMatrix{T}, <:StridedViews.StridedView{T, 4, <:CuArray{T}}}}
+    # COV_EXCL_START
+    # kernels are not reachable by coverage
     @kernel function fill_subblock_kernel!(subblock, val)
         idx = @index(Global, Cartesian)
         idx_val = idx[1] == idx[4] && idx[2] == idx[3] ? val : zero(val)
         @inbounds subblock[idx] = idx_val
     end
+    # COV_EXCL_STOP
     kernel = fill_subblock_kernel!(get_backend(data))
     kernel(data, val; ndrange = size(data))
     return data
--- a/ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
+++ b/ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
@@ -5,8 +5,7 @@ using CUDA: @allowscalar
 using cuTENSOR: cuTENSOR
 using Strided: StridedViews
 import CUDA: rand as curand, rand! as curand!, randn as curandn, randn! as curandn!
-using CUDA.KernelAbstractions
-using CUDA.KernelAbstractions: @kernel, @index
+using CUDA.KernelAbstractions: @kernel, @index, get_backend
 
 using TensorKit
 using TensorKit.Factorizations
@@ -28,7 +27,7 @@ function TensorKit.fill_braidingsubblock!(data::TD, val) where {T, TD <: Union{<
         idx_val = idx[1] == idx[4] && idx[2] == idx[3] ? val : zero(val)
         @inbounds subblock[idx] = idx_val
     end
-    kernel = fill_subblock_kernel!(KernelAbstractions.get_backend(data))
+    kernel = fill_subblock_kernel!(get_backend(data))
     kernel(data, val; ndrange = size(data))
     return data
 end
--- a/ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
+++ b/ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
@@ -3,14 +3,17 @@ module TensorKitCUDAExt
 using CUDA, CUDA.CUBLAS, CUDA.CUSOLVER, LinearAlgebra
 using CUDA: @allowscalar
 using cuTENSOR: cuTENSOR
+using Strided: StridedViews
 import CUDA: rand as curand, rand! as curand!, randn as curandn, randn! as curandn!
+using CUDA.KernelAbstractions
+using CUDA.KernelAbstractions: @kernel, @index
 
 using TensorKit
 using TensorKit.Factorizations
 using TensorKit.Strided
 using TensorKit.Factorizations: AbstractAlgorithm
 using TensorKit: SectorDict, tensormaptype, scalar, similarstoragetype, AdjointTensorMap, scalartype, project_symmetric_and_check
-import TensorKit: randisometry, rand, randn
+import TensorKit: randisometry, rand, randn, fill_braidingsubblock!
 
 using TensorKit: MatrixAlgebraKit
 
@@ -18,5 +21,16 @@ using Random
 
 include("cutensormap.jl")
 include("truncation.jl")
+
+function TensorKit.fill_braidingsubblock!(data::TD, val) where {T, TD <: Union{<:CuMatrix{T}, <:StridedViews.StridedView{T, 4, <:CuArray{T}}}}
+    @kernel function fill_subblock_kernel!(subblock, val)
+        idx = @index(Global, Cartesian)
+        idx_val = idx[1] == idx[4] && idx[2] == idx[3] ? val : zero(val)
+        @inbounds subblock[idx] = idx_val
+    end
+    kernel = fill_subblock_kernel!(KernelAbstractions.get_backend(data))
+    kernel(data, val; ndrange = size(data))
+    return data
+end
 
 end
--- a/ext/TensorKitCUDAExt/cutensormap.jl
+++ b/ext/TensorKitCUDAExt/cutensormap.jl
@@ -168,3 +168,7 @@ for f in (:sqrt, :log, :asin, :acos, :acosh, :atanh, :acoth)
         return tf
     end
 end
+
+function TensorKit._add_transform_multi!(tdst::CuTensorMap, tsrc, p, (U, structs_dst, structs_src)::Tuple{<:Array, TD, TS}, buffers, alpha, beta, backend...) where {TD, TS}
+    return TensorKit._add_transform_multi!(tdst, tsrc, p, (CUDA.Adapt.adapt(CuArray, U), structs_dst, structs_src), buffers, alpha, beta, backend...)
+end
--- a/src/planar/preprocessors.jl
+++ b/src/planar/preprocessors.jl
@@ -184,7 +184,8 @@ function _construct_braidingtensors!(ex, preargs, indexmap, non_braiding) # ex i
             if foundV1 && foundV2
                 s = gensym(:τ)
                 storageex = Expr(:call, GlobalRef(TensorKit, :promote_storagetype), non_braiding...)
-                constructex = Expr(:call, GlobalRef(TensorKit, :BraidingTensor), storageex, V1, V2)
+                braidingex = Expr(:call, GlobalRef(TensorKit, :braidingtensortype), V1, V2, storageex)
+                constructex = Expr(:call, braidingex, V1, V2)
                 push!(preargs, Expr(:(=), s, constructex))
                 obj = _is_adjoint(obj) ? _add_adjoint(s) : s
                 success = true
--- a/src/planar/preprocessors.jl
+++ b/src/planar/preprocessors.jl
@@ -83,6 +83,23 @@ _add_adjoint(ex) = Expr(TO.prime, ex)
 # spaces from the rest of the expression. Construct the explicit BraidingTensor objects and
 # insert them in the expression.
 function _construct_braidingtensors(ex)
+    function filter_f(expr)
+        if TO.istensor(expr)
+            return _remove_adjoint(TO.decomposetensor(expr)[1]) != :τ
+        elseif TO.istensorexpr(expr)
+            return any(filter_f, expr.args)
+        else
+            return false
+        end
+    end
+    function extract_tensors(tensor_ex)
+        if TO.istensor(tensor_ex)
+            return [TO.decomposetensor(tensor_ex)[1]]
+        elseif TO.istensorexpr(tensor_ex)
+            return collect(Iterators.flatmap(extract_tensors, filter(filter_f, tensor_ex.args)))
+        end
+    end
+    # get storagetype
     ex isa Expr || return ex
     if ex.head == :macrocall && ex.args[1] == Symbol("@notensor")
         return ex
@@ -104,7 +121,9 @@ function _construct_braidingtensors(ex)
                 )
             end
         end
-        newrhs, success = _construct_braidingtensors!(rhs, preargs, indexmap)
+        # if this is a definition, the lhs tensor is NOT yet defined
+        no_τ_ex = reduce(vcat, Iterators.flatmap(extract_tensors, filter(filter_f, rhs.args)); init = Symbol[])
+        newrhs, success = _construct_braidingtensors!(rhs, preargs, indexmap, no_τ_ex)
         success ||
             throw(ArgumentError("cannot determine the spaces of all braiding tensors in $ex"))
         pre = Expr(
@@ -115,7 +134,8 @@ function _construct_braidingtensors(ex)
     elseif TO.istensorexpr(ex)
         preargs = Vector{Any}()
         indexmap = Dict{Any, Any}()
-        newex, success = _construct_braidingtensors!(ex, preargs, indexmap)
+        no_τ_ex = reduce(vcat, Iterators.flatmap(extract_tensors, filter(filter_f, ex.args)); init = Symbol[])
+        newex, success = _construct_braidingtensors!(ex, preargs, indexmap, no_τ_ex)
         success ||
             throw(ArgumentError("cannot determine the spaces of all braiding tensors in $ex"))
         pre = Expr(
@@ -128,7 +148,7 @@ function _construct_braidingtensors(ex)
     end
 end
 
-function _construct_braidingtensors!(ex, preargs, indexmap) # ex is guaranteed to be a single tensor expression
+function _construct_braidingtensors!(ex, preargs, indexmap, non_braiding) # ex is guaranteed to be a single tensor expression
     if TO.isscalarexpr(ex)
         # ex could be tensorscalar call with more braiding tensors
         return _construct_braidingtensors(ex), true
@@ -163,7 +183,8 @@ function _construct_braidingtensors!(ex, preargs, indexmap) # ex is guaranteed t
             end
             if foundV1 && foundV2
                 s = gensym(:τ)
-                constructex = Expr(:call, GlobalRef(TensorKit, :BraidingTensor), V1, V2)
+                storageex = Expr(:call, GlobalRef(TensorKit, :promote_storagetype), non_braiding...)
+                constructex = Expr(:call, GlobalRef(TensorKit, :BraidingTensor), storageex, V1, V2)
                 push!(preargs, Expr(:(=), s, constructex))
                 obj = _is_adjoint(obj) ? _add_adjoint(s) : s
                 success = true
@@ -196,7 +217,7 @@ function _construct_braidingtensors!(ex, preargs, indexmap) # ex is guaranteed t
         newargs = Vector{Any}(undef, length(args))
         success = true
         for i in 1:length(ex.args)
-            newargs[i], successa = _construct_braidingtensors!(args[i], preargs, indexmap)
+            newargs[i], successa = _construct_braidingtensors!(args[i], preargs, indexmap, non_braiding)
             success = success && successa
         end
         newex = Expr(ex.head, newargs...)
@@ -212,7 +233,7 @@ function _construct_braidingtensors!(ex, preargs, indexmap) # ex is guaranteed t
             for i in 2:length(ex.args)
                 successes[i] && continue
                 newargs[i], successa = _construct_braidingtensors!(
-                    args[i], preargs, indexmap
+                    args[i], preargs, indexmap, non_braiding
                 )
                 successes[i] = successa
             end
@@ -232,7 +253,7 @@ function _construct_braidingtensors!(ex, preargs, indexmap) # ex is guaranteed t
         indices = [TO.getindices(arg) for arg in args]
         for i in 2:length(ex.args)
             indexmapa = copy(indexmap)
-            newargs[i], successa = _construct_braidingtensors!(args[i], preargs, indexmapa)
+            newargs[i], successa = _construct_braidingtensors!(args[i], preargs, indexmapa, non_braiding)
             for l in indices[i]
                 if !haskey(indexmap, l) && haskey(indexmapa, l)
                     indexmap[l] = indexmapa[l]
@@ -243,10 +264,10 @@ function _construct_braidingtensors!(ex, preargs, indexmap) # ex is guaranteed t
         newex = Expr(ex.head, newargs...)
         return newex, success
     elseif isexpr(ex, :call) && ex.args[1] == :/ && length(ex.args) == 3
-        newarg, success = _construct_braidingtensors!(ex.args[2], preargs, indexmap)
+        newarg, success = _construct_braidingtensors!(ex.args[2], preargs, indexmap, non_braiding)
         return Expr(:call, :/, newarg, ex.args[3]), success
     elseif isexpr(ex, :call) && ex.args[1] == :\ && length(ex.args) == 3
-        newarg, success = _construct_braidingtensors!(ex.args[3], preargs, indexmap)
+        newarg, success = _construct_braidingtensors!(ex.args[3], preargs, indexmap, non_braiding)
         return Expr(:call, :\, ex.args[2], newarg), success
     else
         error("unexpected expression $ex")
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -166,8 +166,6 @@ function fill_braidingblock!(data, b::BraidingTensor, s::Sector)
         r = _braiding_factor(f₁, f₂, b.adjoint)
         # change offset to account for single block
         subblock = StridedView(data, sz, str, off - base_offset)
-        # without the zero-value, the non-trivial block is not set
-        # correctly in the GPU case
         isnothing(r) ? zerovector!(subblock) : fill_braidingsubblock!(subblock, r)
     end
     return data
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -53,25 +53,30 @@ function BraidingTensor{T}(V::HomSpace, adjoint::Bool = false) where {T}
         throw(SpaceMismatch("Cannot define a braiding on $V"))
     return BraidingTensor{T}(V[2], V[1], adjoint)
 end
+
+function Base.adjoint(b::BraidingTensor{T, S, A}) where {T, S, A}
+    return BraidingTensor{T, S, A}(b.V1, b.V2, !b.adjoint)
+end
+
 # these are here to make the preprocessing for `@planar` expressions less painful
-function BraidingTensor(TorA::Type, V1::S, V2::S, adjoint::Bool = false) where {S <: IndexSpace}
+function braidingtensortype(::Type{S}, ::Type{TorA}) where {S <: IndexSpace, TorA}
     T = eltype(TorA)
     TA = similarstoragetype(TorA)
-    return BraidingTensor{T, S, TA}(V1, V2, adjoint)
+    return BraidingTensor{T, S, TA}
 end
-function BraidingTensor(TorA::Type, V1::IndexSpace, V2::IndexSpace, adjoint::Bool = false)
-    return BraidingTensor(TorA, promote(V1, V2)..., adjoint)
+braidingtensortype(V::S, ::Type{TorA}) where {S <: IndexSpace, TorA} = braidingtensortype(S, TorA)
+braidingtensortype(V1::S, V2::S, ::Type{TorA}) where {S <: IndexSpace, TorA} = braidingtensortype(S, TorA)
+function braidingtensortype(V1::IndexSpace, V2::IndexSpace, ::Type{TorA}) where {TorA}
+    S = promote(V1, V2)
+    return braidingtensortype(S..., TorA)
 end
-function BraidingTensor(TorA::Type, V::HomSpace, adjoint::Bool = false)
+function braidingtensortype(V::HomSpace, ::Type{TorA}) where {TorA}
     domain(V) == reverse(codomain(V)) ||
         throw(SpaceMismatch("Cannot define a braiding on $V"))
     T = eltype(TorA)
     S = spacetype(V[2])
-    TA = storagetype(TorA)
-    return BraidingTensor{T, S, TA}(V[2], V[1], adjoint)
-end
-function Base.adjoint(b::BraidingTensor{T, S, A}) where {T, S, A}
-    return BraidingTensor{T, S, A}(b.V1, b.V2, !b.adjoint)
+    TA = similarstoragetype(TorA)
+    return braidingtensortype(S, TA)
 end
 
 storagetype(::Type{BraidingTensor{T, S, A}}) where {T, S, A} = A
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -69,8 +69,6 @@ function braidingtensortype(V1::IndexSpace, V2::IndexSpace, ::Type{TorA}) where 
     return braidingtensortype(S..., TorA)
 end
 function braidingtensortype(V::HomSpace, ::Type{TorA}) where {TorA}
-    domain(V) == reverse(codomain(V)) ||
-        throw(SpaceMismatch("Cannot define a braiding on $V"))
     return braidingtensortype(spacetype(V), TorA)
 end
 
--- a/src/tensors/braidingtensor.jl
+++ b/src/tensors/braidingtensor.jl
@@ -2,72 +2,80 @@
 # special (2,2) tensor that implements a standard braiding operation
 #====================================================================#
 """
-    struct BraidingTensor{T,S<:IndexSpace} <: AbstractTensorMap{T, S, 2, 2}
+    struct BraidingTensor{T, S <: IndexSpace, A <: DenseVector{T}} <: AbstractTensorMap{T, S, 2, 2}
     BraidingTensor(V1::S, V2::S, adjoint::Bool=false) where {S<:IndexSpace}
+    BraidingTensor{T, S, A}(V1::S, V2::S, adjoint::Bool=false) where {T, S, A}
 
 Specific subtype of [`AbstractTensorMap`](@ref) for representing the braiding tensor that
 braids the first input over the second input; its inverse can be obtained as the adjoint.
 
 It holds that `domain(BraidingTensor(V1, V2)) == V1 ⊗ V2` and
-`codomain(BraidingTensor(V1, V2)) == V2 ⊗ V1`.
+`codomain(BraidingTensor(V1, V2)) == V2 ⊗ V1`. The storage type `TA`
+controls the array type of the braiding tensor used when indexing
+and multiplying with other tensors.
 """
-struct BraidingTensor{T, S} <: AbstractTensorMap{T, S, 2, 2}
+struct BraidingTensor{T, S, A <: DenseVector{T}} <: AbstractTensorMap{T, S, 2, 2}
     V1::S
     V2::S
     adjoint::Bool
-    function BraidingTensor{T, S}(V1::S, V2::S, adjoint::Bool = false) where {T, S <: IndexSpace}
-        for a in sectors(V1)
-            for b in sectors(V2)
-                for c in (a ⊗ b)
-                    Nsymbol(a, b, c) == Nsymbol(b, a, c) ||
-                        throw(ArgumentError("Cannot define a braiding between $a and $b"))
-                end
-            end
+    function BraidingTensor{T, S, A}(V1::S, V2::S, adjoint::Bool = false) where {T, S <: IndexSpace, A <: DenseVector{T}}
+        for a in sectors(V1), b in sectors(V2), c in (a ⊗ b)
+            Nsymbol(a, b, c) == Nsymbol(b, a, c) ||
+                throw(ArgumentError("Cannot define a braiding between $a and $b"))
         end
-        return new{T, S}(V1, V2, adjoint)
+        return new{T, S, A}(V1, V2, adjoint)
         # partial construction: only construct rowr and colr when needed
     end
 end
 function BraidingTensor{T}(V1::S, V2::S, adjoint::Bool = false) where {T, S <: IndexSpace}
-    return BraidingTensor{T, S}(V1, V2, adjoint)
-end
-function BraidingTensor{T}(V1::IndexSpace, V2::IndexSpace, adjoint::Bool = false) where {T}
-    return BraidingTensor{T}(promote(V1, V2)..., adjoint)
-end
-function BraidingTensor(V1::IndexSpace, V2::IndexSpace, adjoint::Bool = false)
-    return BraidingTensor(promote(V1, V2)..., adjoint)
+    TA = similarstoragetype(T)
+    return BraidingTensor{T, S, TA}(V1, V2, adjoint)
 end
 function BraidingTensor(V1::S, V2::S, adjoint::Bool = false) where {S <: IndexSpace}
     T = BraidingStyle(sectortype(S)) isa SymmetricBraiding ? Float64 : ComplexF64
-    return BraidingTensor{T, S}(V1, V2, adjoint)
+    return BraidingTensor{T}(V1, V2, adjoint)
+end
+function BraidingTensor(V1::IndexSpace, V2::IndexSpace, adjoint::Bool = false)
+    return BraidingTensor(promote(V1, V2)..., adjoint)
 end
 function BraidingTensor(V::HomSpace, adjoint::Bool = false)
     domain(V) == reverse(codomain(V)) ||
         throw(SpaceMismatch("Cannot define a braiding on $V"))
     return BraidingTensor(V[2], V[1], adjoint)
 end
+function BraidingTensor{T, S, A}(V::HomSpace, adjoint::Bool = false) where {T, S, A}
+    domain(V) == reverse(codomain(V)) ||
+        throw(SpaceMismatch("Cannot define a braiding on $V"))
+    return BraidingTensor{T, S, A}(V[2], V[1], adjoint)
+end
 function BraidingTensor{T}(V::HomSpace, adjoint::Bool = false) where {T}
     domain(V) == reverse(codomain(V)) ||
         throw(SpaceMismatch("Cannot define a braiding on $V"))
     return BraidingTensor{T}(V[2], V[1], adjoint)
 end
-function Base.adjoint(b::BraidingTensor{T, S}) where {T, S}
-    return BraidingTensor{T, S}(b.V1, b.V2, !b.adjoint)
+# these are here to make the preprocessing for `@planar` expressions less painful
+function BraidingTensor(TorA::Type, V1::S, V2::S, adjoint::Bool = false) where {S <: IndexSpace}
+    T = eltype(TorA)
+    TA = similarstoragetype(TorA)
+    return BraidingTensor{T, S, TA}(V1, V2, adjoint)
+end
+function BraidingTensor(TorA::Type, V1::IndexSpace, V2::IndexSpace, adjoint::Bool = false)
+    return BraidingTensor(TorA, promote(V1, V2)..., adjoint)
+end
+function BraidingTensor(TorA::Type, V::HomSpace, adjoint::Bool = false)
+    domain(V) == reverse(codomain(V)) ||
+        throw(SpaceMismatch("Cannot define a braiding on $V"))
+    T = eltype(TorA)
+    S = spacetype(V[2])
+    TA = storagetype(TorA)
+    return BraidingTensor{T, S, TA}(V[2], V[1], adjoint)
+end
+function Base.adjoint(b::BraidingTensor{T, S, A}) where {T, S, A}
+    return BraidingTensor{T, S, A}(b.V1, b.V2, !b.adjoint)
 end
 
+storagetype(::Type{BraidingTensor{T, S, A}}) where {T, S, A} = A
 space(b::BraidingTensor) = b.adjoint ? b.V1 ⊗ b.V2 ← b.V2 ⊗ b.V1 : b.V2 ⊗ b.V1 ← b.V1 ⊗ b.V2
-
-# specializations to ignore the storagetype of BraidingTensor
-promote_storagetype(::Type{A}, ::Type{B}) where {A <: BraidingTensor, B <: AbstractTensorMap} = storagetype(B)
-promote_storagetype(::Type{A}, ::Type{B}) where {A <: AbstractTensorMap, B <: BraidingTensor} = storagetype(A)
-promote_storagetype(::Type{A}, ::Type{B}) where {A <: BraidingTensor, B <: BraidingTensor} = storagetype(A)
-
-promote_storagetype(::Type{T}, ::Type{A}, ::Type{B}) where {T <: Number, A <: BraidingTensor, B <: AbstractTensorMap} =
-    similarstoragetype(B, T)
-promote_storagetype(::Type{T}, ::Type{A}, ::Type{B}) where {T <: Number, A <: AbstractTensorMap, B <: BraidingTensor} =
-    similarstoragetype(A, T)
-promote_storagetype(::Type{T}, ::Type{A}, ::Type{B}) where {T <: Number, A <: BraidingTensor, B <: BraidingTensor} =
-    similarstoragetype(A, T)
 
 function Base.getindex(b::BraidingTensor)
     sectortype(b) === Trivial || throw(SectorMismatch())
@@ -99,6 +107,13 @@ function _braiding_factor(f₁, f₂, inv::Bool = false)
     return r
 end
 
+# generates scalar indexing errors on GPU
+function fill_braidingsubblock!(data, val)
+    f(I) = ((I[1] == I[4]) & (I[2] == I[3])) * val
+    return data .= f.(CartesianIndices(data))
+end
+
+
 @inline function subblock(
         b::BraidingTensor, (f₁, f₂)::Tuple{FusionTree{I, 2}, FusionTree{I, 2}}
     ) where {I <: Sector}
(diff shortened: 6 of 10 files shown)


kshyatt:
The only thing missing here is the parser preprocessing for `@planar`, where we need to extract the storagetype to hand it over to the `BraidingTensor` constructor.

Jutho:
I will try to review asap, but I first have to catch up on the latest status of the treetransform stuff, as I never managed to review that part of the vectorize fusiontrees PR.

src/tensors/braidingtensor.jl
```diff
@@ -115,15 +135,12 @@ end
     d = (dims(codomain(b), f₁.uncoupled)..., dims(domain(b), f₂.uncoupled)...)
     n1 = d[1] * d[2]
     n2 = d[3] * d[4]
-    data = sreshape(StridedView(Matrix{eltype(b)}(undef, n1, n2)), d)
+    data_t = similarmatrixtype(storagetype(b))(undef, (n1, n2))
+    data = sreshape(StridedView(data_t), d)
```
Jutho: I am a bit confused of why we first want to create `data_t` as a `Matrix`. I see this was the case in the old code, but even there I don't know why this was still in there. For regular tensors, `subblock` will create a `StridedView` with a `Vector`/`CuVector` parent, so we can do the same thing here, i.e.

```julia
parent = storagetype(b)(undef, prod(d))
data = sreshape(StridedView(parent), d)
```

Jutho: This would eliminate the need for all of `similarmatrixtype`.

src/tensors/braidingtensor.jl
```diff
@@ -145,36 +185,17 @@ function block(b::BraidingTensor, s::Sector)
     # TODO: probably always square?
     m = blockdim(codomain(b), s)
     n = blockdim(domain(b), s)
-    data = Matrix{eltype(b)}(undef, (m, n))
```
Jutho, line 148: Here too, there is no need to ever create a Matrix. Our `block` function on regular `TensorMap` uses `reshape` and `view` (the Base one, not the StridedViews one) into a vector to return the block.

src/tensors/braidingtensor.jl
kshyatt: I think because deeper down in the indexing code, the subblock logic creates new `StridedView`s of `data`, and creating a `StridedView` from a `StridedView` doesn't work.

Jutho: I don't think that's true. `subblock` is already the lowest level of indexing. (Also, but I think unrelated, `StridedView` acting on a `StridedView` should work).

kshyatt: I'm telling you what I observed from running the actual tests, so apparently it is true? I will try to modify things to work around this

kshyatt: `similarmatrixtype` is dead 🔪 🔪 🔪 . Things seem to be working locally, let's see how it does on CI. Thanks for the tip to simplify all this!

kshyatt:
Anything more need doing here or is this ready to go after https://github.com/QuantumKitHub/TensorKit.jl/pull/389?

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,85 @@
...
+        for a in sectors(V1), b in sectors(V2), c in (a ⊗ b)
+            Nsymbol(a, b, c) == Nsymbol(b, a, c) ||
+                throw(ArgumentError("Cannot define a braiding between $a and $b"))
         end
-        return new{T, S}(V1, V2, adjoint)
+        return new{T, S, A}(V1, V2, adjoint)
         # partial construction: only construct rowr and colr when needed
     end
 end
+function BraidingTensor{T, S}(V1::S, V2::S, ::Type{A}, adjoint::Bool = false) where {T, S <: IndexSpace, A}
```
lkdvos: I think I'm not completely convinced by this constructor syntax, it feels like at this point we might as well just call `BraidingTensor{T, S, A}` directly, rather than including the `Type{A}` as an argument. This also just cuts down slightly on the number of constructors we need to have, under the *less-is-more* strategy :)

src/tensors/braidingtensor.jl
```diff
@@ -99,6 +111,14 @@ function _braiding_factor(f₁, f₂, inv::Bool = false)
     return r
 end
 
+function _set_subblock!(data, val)
+    @inbounds for i in axes(data, 1), j in axes(data, 2)
+        data[i, j, j, i] = val
+    end
+    return data
+end
```
lkdvos:
```suggestion
function _set_subblock!(data, val)
    f(I) = ((I[1] == I[4]) & (I[2] == I[3])) * val
    data .= f.(CartesianIndices(data))
end
```

Would something like this work immediately on GPU, or does that still give issues? At least locally it seems to work on `JLArray`s.

lkdvos: (If it does, also need to remove the `fill!(data, zero...)` at the different callsites)

src/tensors/braidingtensor.jl
```diff
@@ -169,14 +173,31 @@ function block(b::BraidingTensor, s::Sector)
...
+    # TODO: probably always square?
+    m = blockdim(codomain(b), s)
+    n = blockdim(domain(b), s)
+
+    data = reshape(storagetype(b)(undef, m * n), (m, n))
+
+    m * n == 0  && return data # s ∉ blocksectors(b)
+    fill!(data, zero(eltype(b)))
+
+    if sectortype(b) === Trivial
```
lkdvos: Mostly out of curiosity, is this an optimization or was this required?

test/cuda/planar.jl
```diff
@@ -0,0 +1,300 @@
...
+        ρ = CUDA.randn(T, Vmera ⊗ Vmera ⊗ Vmera ← Vmera ⊗ Vmera ⊗ Vmera)
+        h = CUDA.randn(T, Vmera ⊗ Vmera ⊗ Vmera ← Vmera ⊗ Vmera ⊗ Vmera)
+
+        u′ = force_planar(u)
+        w′ = force_planar(w)
+        ρ′ = force_planar(ρ)
+        h′ = force_planar(h)
+
+        for alloc in
+            (TensorOperations.DefaultAllocator(), TensorOperations.CUDAAllocator())
```
lkdvos: I think this might just run the same test twice, (although I might be wrong) does the default allocator not coincide with the CUDAAllocator here?

Project.toml
```diff
@@ -36,12 +36,16 @@ TensorKitMooncakeExt = "Mooncake"
 [workspace]
 projects = ["test"]
 
+[extras]
+GPUArrays = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
```
lkdvos: This might be a merge error, and probably belongs in a different project.toml?

ext/TensorKitCUDAExt/cutensormap.jl
```diff
@@ -168,3 +168,23 @@ for f in (:sqrt, :log, :asin, :acos, :acosh, :atanh, :acoth)
...
+        # Special case without intermediate buffers whenever there is only a single block
+        if length(subtransformer[1]) == 1
+            TensorKit._add_transform_single!(tdst, tsrc, p, subtransformer, α, β, backend...)
+        else
+            cu_subtransformer = tuple(CUDA.adapt(CuArray, subtransformer[1]), subtransformer[2:end]...)
+            TensorKit._add_transform_multi!(tdst, tsrc, p, cu_subtransformer, buffers, α, β, backend...)
+        end
+    end
+    return nothing
+end
```
lkdvos, line 174: This probably is something separate from this PR right? (Also, might no longer be necessary of the `mul!` specializations of `StridedView`s are correctly handled in the later versions of Strided.jl

Project.toml
kshyatt: Ah no this was here because of a GPUArrays bug that's now fixed

src/tensors/braidingtensor.jl
kshyatt: I'll try it, might also be worth checking if this impacts performance considering it now touches every element of `data`?

src/tensors/braidingtensor.jl
kshyatt: Sorry, what's "this"?

src/tensors/braidingtensor.jl
lkdvos: It does, but previously we had to touch every element, and then alter the specific ones on top (`fill!(data, zero(T)); _set_subblock!(data, val)` is now replaced with the single `_set_subblock!(data, val)`). So really this merges the two together, should be quite friendly data-locality-wise, so it might actually be faster?

lkdvos: Can confirm:

<details>

```julia
julia> function f!(data, val)
       fill!(data, zero(eltype(data)))
       @inbounds for i in axes(data, 1), j in axes(data, 2)
       data[i, j, j, i] = val
       end
       return data
       end
f! (generic function with 1 method)

julia> function f_ordered!(data, val)
       fill!(data, zero(eltype(data)))
       @inbounds for j in axes(data, 2), i in axes(data, 1) # stride-1 in tightest loop
       data[i, j, j, i] = val
       end
       return data
       end
f_ordered! (generic function with 1 method)

julia> helper(I::CartesianIndex{4}) = I[1] == I[4] & I[2] == I[3]
helper (generic function with 1 method)

julia> function g!(data, val)
       data .= helper.(CartesianIndices(data)) .* Ref(val)
       return data
       end
g! (generic function with 1 method)

julia> N = M = 100;

julia> @benchmark f!(data, $(rand())) setup = (data = rand($N, $M, $M, $N))
BenchmarkTools.Trial: 16 samples with 1 evaluation per sample.
 Range (min … max):  106.735 ms … 107.348 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     107.098 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   107.115 ms ± 139.543 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                  █
  ▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆▁▁▁▁▁█▁▆▆▁▆▆▁▁▆▁▁▆▁▁▆▁▁▁▁▁▆▆▁▁▁▁▁▆ ▁
  107 ms           Histogram: frequency by time          107 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark f_ordered!(data, $(rand())) setup = (data = rand($N, $M, $M, $N))
BenchmarkTools.Trial: 16 samples with 1 evaluation per sample.
 Range (min … max):  106.576 ms … 107.345 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     107.034 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   106.977 ms ± 214.293 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

      ▃                              ▃ █
  ▇▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▇█▁█▁▇▁▇▁▁▇▁▁▁▁▇▁▁▁▁▁▁▁▁▁▁▇ ▁
  107 ms           Histogram: frequency by time          107 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark g!(data, $(rand())) setup = (data = rand($N, $M, $M, $N))
BenchmarkTools.Trial: 18 samples with 1 evaluation per sample.
 Range (min … max):  85.037 ms …  85.668 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     85.173 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   85.221 ms ± 178.977 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▁▁  ██   ▁▁▁  ▁▁▁  ▁▁      ▁    ▁                      ▁   ▁
  ██▁▁██▁▁▁███▁▁███▁▁██▁▁▁▁▁▁█▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁█ ▁
  85 ms           Histogram: frequency by time         85.7 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

</details>

It does intuitively make sense, this thing must be completely memory-bound anyways so just the fact of merging the data accesses should just account for the difference in speed.
(As an aside, I also tried switching the loop order which doesn't make a difference here but actually slows things down in smaller tests, where I think the jumps for the middle strides still fit in cachelines somewhat).

As a final note on the entire topic: this is actually something that really only ever gets called in fallback paths (where we instantiate a `BraidingTensor` to a `TensorMap` because we are too lazy to work with the lazy BraidingTensor representation. If this ends up being the bottleneck somewhere, the better approach would be to start working with the `BraidingTensor` directly, and less so to try and make the `TensorMap` conversion faster :) (I am fully aware that this is completely not obvious from anything that is documented, definitely my apologies for that!)

kshyatt: Nice, thanks for running the tests. This also lets us get rid of a custom kernel in the extension.

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: No, this was needed for a test to work, let me look up which one in a moment...

src/tensors/braidingtensor.jl
lkdvos: My bad, I meant the split between the trivial and non-trivial implementation.

kshyatt: Oh it was so that both could dispatch to `set_sublock` (which I could specialize) but do their own setup, and to cleanup a quite long function.

src/tensors/braidingtensor.jl
kshyatt: Actually this breaks on CUDA 😭 . I'll see if I can't figure out a fix as it would be nice to not have to write a custom kernel...

src/tensors/braidingtensor.jl
kshyatt: Trimmed this down quite a bit where possible

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: Yep, it's needed for `braid()` for `FibonacciAnyon`...

test/cuda/planar.jl
kshyatt: I'll just drop down to `DefaultAllocator`

ext/TensorKitCUDAExt/cutensormap.jl
lkdvos: Ok, but is it still just the `mul!` call with `StridedView` that is giving issues? Should we not fix that there instead?

ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
```diff
@@ -19,4 +23,16 @@ using Random
...
 
+function TensorKit._set_subblock!(data::TD, val) where {T, TD <: Union{<:CuMatrix{T}, <:StridedViews.StridedView{T, 4, <:CuArray{T}}}}
+    @kernel function fill_subblock_kernel!(subblock, val)
+        idx = @index(Global, Cartesian)
+        @inbounds subblock[idx[1], idx[2], idx[2], idx[1]] = val
+    end
+    kernel = fill_subblock_kernel!(KernelAbstractions.get_backend(data))
+    d1 = size(data, 1)
+    d2 = size(data, 2)
+    kernel(data, val; ndrange = (d1, d2))
```
lkdvos:
```suggestion
    @kernel function fill_subblock_kernel!(subblock, val)
        idx = @index(Global, Cartesian)
        @inbounds subblock[idx] = ((idx[1] == idx[4]) & [idx[2] == idx[3])) * val
    end
    kernel = fill_subblock_kernel!(KernelAbstractions.get_backend(data))
    kernel(data, val; ndrange = size(data))
```

Given the benchmarks I did, I think we still should just merge the "setting to zero" and the "setting to val" parts.

src/tensors/braidingtensor.jl
```diff
@@ -99,6 +93,12 @@ function _braiding_factor(f₁, f₂, inv::Bool = false)
     return r
 end
 
+function _set_subblock!(data, val)
```
lkdvos:
```suggestion
function fill_braidingsubblock!(data, val)
```
This is mostly just to make things clearer, since it's not really just setting the subblock. (also possible: `fill_braidingtensor!`, `fill_braidingblock!`, ...

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,67 @@
...
-function BraidingTensor(V1::S, V2::S, adjoint::Bool = false) where {S <: IndexSpace}
-    T = BraidingStyle(sectortype(S)) isa SymmetricBraiding ? Float64 : ComplexF64
-    return BraidingTensor{T, S}(V1, V2, adjoint)
-end
 function BraidingTensor(V::HomSpace, adjoint::Bool = false)
     domain(V) == reverse(codomain(V)) ||
         throw(SpaceMismatch("Cannot define a braiding on $V"))
     return BraidingTensor(V[2], V[1], adjoint)
 end
+function BraidingTensor(V::HomSpace, A, adjoint::Bool = false)
```
lkdvos:
```suggestion
function BraidingTensor{T, S, A}(V::HomSpace, adjoint::Bool = false) where {T, S, A}
```
I still think we should stay consistent with our type parameters not as positional arguments for the constructors, even if that requires us to add some more of these passing-through constructors

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,67 @@
...
+                throw(ArgumentError("Cannot define a braiding between $a and $b"))
         end
-        return new{T, S}(V1, V2, adjoint)
+        return new{T, S, A}(V1, V2, adjoint)
         # partial construction: only construct rowr and colr when needed
     end
 end
 function BraidingTensor{T}(V1::S, V2::S, adjoint::Bool = false) where {T, S <: IndexSpace}
-    return BraidingTensor{T, S}(V1, V2, adjoint)
+    return BraidingTensor{T, S, Vector{T}}(V1, V2, adjoint)
```
lkdvos:
```suggestion
    return BraidingTensor{T, S, similarstoragetype(T)}(V1, V2, adjoint)
```
This is the default value used by `TensorMap`, so we can couple this to that for if we ever end up changing to `Memory{T}`

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,67 @@
...
 end
-function BraidingTensor{T}(V1::IndexSpace, V2::IndexSpace, adjoint::Bool = false) where {T}
-    return BraidingTensor{T}(promote(V1, V2)..., adjoint)
+function BraidingTensor(V1::S, V2::S, adjoint::Bool = false) where {S <: IndexSpace}
+    T = BraidingStyle(sectortype(S)) isa SymmetricBraiding ? Float64 : ComplexF64
+    return BraidingTensor{T, S, Vector{T}}(V1, V2, adjoint)
```
lkdvos:
```suggestion
    return BraidingTensor{T}(V1, V2, adjoint)
```

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,67 @@
 # special (2,2) tensor that implements a standard braiding operation
 #====================================================================#
 """
-    struct BraidingTensor{T,S<:IndexSpace} <: AbstractTensorMap{T, S, 2, 2}
-    BraidingTensor(V1::S, V2::S, adjoint::Bool=false) where {S<:IndexSpace}
+    struct BraidingTensor{T,S<:IndexSpace,A<:DenseVector{T}} <: AbstractTensorMap{T, S, 2, 2}
+    BraidingTensor(V1::S, V2::S, ::Type{A}, adjoint::Bool=false) where {S<:IndexSpace, A <: DenseVector{<:Number}}
```
lkdvos:
```suggestion
    struct BraidingTensor{T, S <: IndexSpace, A <: DenseVector{T}} <: AbstractTensorMap{T, S, 2, 2}
    BraidingTensor(V1::S, V2::S, adjoint::Bool = false) where {S <: IndexSpace}
    BraidingTensor{T, S, A}(V1::S, V2::S, adjoint::Bool = false) where {T, S, A}
```

lkdvos (commented):
I guess the main thing missing is still the `@planar` changes... I will try to find some time, but I seem to be having a million things 🙃

kshyatt:
I'll work on the preprocessor, don't worry about it! But my solution might be a bit ugly...

ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
Jutho: Which one is faster? First zeroing everything and then only setting the elements, or this integrated approach from your suggestion?

(Should probably be `@inbounds subblock[idx[1], idx[2], idx[3], idx[4]] = ...`)

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,67 @@
...
-    return BraidingTensor{T}(promote(V1, V2)..., adjoint)
+function BraidingTensor(V1::S, V2::S, adjoint::Bool = false) where {S <: IndexSpace}
+    T = BraidingStyle(sectortype(S)) isa SymmetricBraiding ? Float64 : ComplexF64
+    return BraidingTensor{T, S, Vector{T}}(V1, V2, adjoint)
+end
+# necessary due to HomSpace ctor below
+function BraidingTensor(V1::S, V2::S, A, adjoint::Bool = false) where {S <: IndexSpace}
+    T = eltype(A)
+    return BraidingTensor{T, S, A}(V1, V2, adjoint)
 end
```
Jutho: Lukas' next comment then also applies here. Or probably this constructor definition can go completely, as it doesn't have a counterpart if you only know `A` and not `T`.

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,67 @@
...
-end
 function BraidingTensor(V::HomSpace, adjoint::Bool = false)
     domain(V) == reverse(codomain(V)) ||
         throw(SpaceMismatch("Cannot define a braiding on $V"))
     return BraidingTensor(V[2], V[1], adjoint)
 end
+function BraidingTensor(V::HomSpace, A, adjoint::Bool = false)
+    domain(V) == reverse(codomain(V)) ||
+        throw(SpaceMismatch("Cannot define a braiding on $V"))
+    return BraidingTensor(V[2], V[1], A, adjoint)
```
Jutho: This will then also need to change.

src/tensors/braidingtensor.jl
```diff
@@ -134,33 +131,20 @@ TensorMap(b::BraidingTensor) = copy!(similar(b), b)
...
-    # TODO: probably always square?
-    m = blockdim(codomain(b), s)
-    n = blockdim(domain(b), s)
-    data = Matrix{eltype(b)}(undef, (m, n))
-
-    length(data) == 0 && return data # s ∉ blocksectors(b)
-
-    data = fill!(data, zero(eltype(b)))
-
+function _trivial_subblock!(data, b::BraidingTensor)
```
Jutho: In analogy of Lukas' suggestion:


```suggestion
function fill_trivial_braidingblock!(data, b::BraidingTensor)
```

src/tensors/braidingtensor.jl
```diff
@@ -134,33 +131,20 @@ TensorMap(b::BraidingTensor) = copy!(similar(b), b)
...
-        end
-        return data
-    end
+    d1, d2 = dim(V1), dim(V2)
+    subblock = sreshape(StridedView(data), (d1, d2, d2, d1))
+    _set_subblock!(subblock, one(eltype(b)))
+    return data
+end
 
+function _nontrivial_subblock!(data, b::BraidingTensor, s::Sector)
```
Jutho:
```suggestion
function fill_nontrivial_braidingblock!(data, b::BraidingTensor, s::Sector)
```

Also, these could probably be within the same method `fill_braidingblock` where also in the trivial case you pass a `Trivial` sector, and then either dispatch on the `Sector` type, or in the function body have an explict

```
if s isa Trivial
   # trivial implementation
else
   #
end
```

ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
lkdvos: At least on the CPU, this integrated approach was faster. The benchmarks are presented in one of the comments here: https://github.com/QuantumKitHub/TensorKit.jl/pull/393#discussion_r3094542942
I was hoping to have that avoid the need of a kernel to begin with, this morning @kshyatt and I were discussing something like this too:

```julia
a = 1:2
a1 = reshape(a, 2, 1, 1, 1)
a2 = reshape(a, 1, 2, 1, 1)
a3 = reshape(a, 1, 1, 2, 1)
a4 = reshape(a, 1, 1, 1, 2)
data .= (a1 .== a4 .& a3 .== a4) .* val
```

kshyatt: The really annoying thing here is that both `CartesianIndices` and `OneTo` (like the result of `axes`) play very poorly on the GPU, and I'd rather not instantiate a full CPU array of all the indices then copy it to the GPU. This was the origin of the kernel...

kshyatt: OK, I went back and forth on this over the weekend and it seems the `CartesianIndices` really doesn't play well with the GPU. I did simplify the custom kernel quite a bit to mostly match what the CPU side does.

kshyatt:
OK, a lot of comments resolved (hopefully!)

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: I guess I don't want to silently transfer all CPU-side arrays for `mul!` *in general* as this will have pretty bad performance impacts and hide bugs. I've done it specifically here because there's currently no way to instantiate a GPU-side `GenericTreeTransformer`, which I can try to fix, although I think that should be a separate PR as this has already been open for weeks.

src/tensors/braidingtensor.jl
kshyatt: FWIW this change made the preprocessor changes much uglier

kshyatt:
CUDA test fail here seems unrelated, I'm running locally to see if I can repro

kshyatt:
Aaaaaand I can't repro locally!

ext/TensorKitCUDAExt/cutensormap.jl
lkdvos: Good point, I forgot that is indeed not what we want in general...
Let me make one final suggestion about this then, just to reduce some of the other logic duplication:
```julia
function TensorKit._add_transform_multi!(tdst::CuTensorMap, tsrc, p, (U, structs_dst, structs_src), buffers, alpha, beta, backend...)
    return TensorKit_add_transform_multi!(tdst, tsrc, p, (adapt(CuArray, U), structs_dst, structs_src), bueffers, alpha, beta, backend...)
end
```

ext/TensorKitAdaptExt.jl
```diff
@@ -15,8 +15,11 @@ function Adapt.adapt_structure(to, x::DiagonalTensorMap)
     data′ = adapt(to, x.data)
     return DiagonalTensorMap(data′, x.domain)
 end
-function Adapt.adapt_structure(::Type{TorA}, x::BraidingTensor) where {TorA <: Union{Number, DenseArray{<:Number}}}
-    return BraidingTensor{scalartype(TorA)}(space(x), x.adjoint)
+function Adapt.adapt_structure(::Type{T}, x::BraidingTensor{T′, S, A}) where {T <: Number, T′, S, A}
+    return BraidingTensor{T}(space(x), x.adjoint)
```
lkdvos: I think this might now lose the storagetype information, i.e. `adapt(Float32, BraidingTensor{Float64, ComplexSpace, CuVector{Float64}}) == BraidingTensor{Float32, ComplexSpace, Vector{Float32}})`.


```suggestion
    A = TensorKit.similarstoragetype(A, T)
    return BraidingTensor{T, S, A}(space(x), x.adjoint)
```

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,81 @@
 # special (2,2) tensor that implements a standard braiding operation
 #====================================================================#
 """
-    struct BraidingTensor{T,S<:IndexSpace} <: AbstractTensorMap{T, S, 2, 2}
+    struct BraidingTensor{T,S<:IndexSpace,A<:DenseVector{T}} <: AbstractTensorMap{T, S, 2, 2}
```
lkdvos:
```suggestion
    struct BraidingTensor{T, S <: IndexSpace, A <: DenseVector{T}} <: AbstractTensorMap{T, S, 2, 2}
```

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,81 @@
...
 braids the first input over the second input; its inverse can be obtained as the adjoint.
 
 It holds that `domain(BraidingTensor(V1, V2)) == V1 ⊗ V2` and
-`codomain(BraidingTensor(V1, V2)) == V2 ⊗ V1`.
+`codomain(BraidingTensor(V1, V2)) == V2 ⊗ V1`. The storage type `TA`
+controls the array type of the braiding tensor used when indexing
+and multiplying with other tensors.
 """
-struct BraidingTensor{T, S} <: AbstractTensorMap{T, S, 2, 2}
+struct BraidingTensor{T, S, A} <: AbstractTensorMap{T, S, 2, 2}
```
lkdvos:
```suggestion
struct BraidingTensor{T, S, A <: DenseVector{T}} <: AbstractTensorMap{T, S, 2, 2}
```
Just to keep it consistent with `TensorMap`

src/tensors/braidingtensor.jl
```diff
@@ -115,15 +130,11 @@ end
...
+    data_parent = storagetype(b)(undef, prod(d))
+    data = sreshape(StridedView(data_parent), d)
     r = _braiding_factor(f₁, f₂, b.adjoint)
-    if !isnothing(r)
-        @inbounds for i in axes(data, 1), j in axes(data, 2)
-            data[i, j, j, i] = r
-        end
-    end
+    val = isnothing(r) ? zero(eltype(b)) : r
+    fill_braidingsubblock!(data, val)
```
lkdvos:
```suggestion
    isnothing(r) ? zerovector!(data) : fill_braidingsubblock!(data, r)
```

src/tensors/braidingtensor.jl
```diff
@@ -134,49 +145,53 @@ TensorMap(b::BraidingTensor) = copy!(similar(b), b)
...
-        isnothing(r) && continue
         # change offset to account for single block
         subblock = StridedView(data, sz, str, off - base_offset)
-        @inbounds for i in axes(subblock, 1), j in axes(subblock, 2)
-            subblock[i, j, j, i] = r
-        end
+        # without the zero-value, the non-trivial block is not set
+        # correctly in the GPU case
+        val = isnothing(r) ? zero(eltype(data)) : r
+        fill_braidingsubblock!(subblock, val)
```
lkdvos:
```suggestion
        isnothing(r) ? zerovector!(subblock) : fill_braidingsubblock!(subblock, r)
```

test/setup.jl
lkdvos:
```suggestion
    tdst = similar(tsrc, force_planar(codomain(tsrc)) ← force_planar(domain(tsrc)))
```
I'm not sure if this works (although it really should), and this probably should have been defined like this to begin with

lkdvos:
```suggestion
    tdst = similar(tsrc, force_planar(codomain(tsrc)) ← force_planar(domain(tsrc)))
```
same comment here

kshyatt: Why no prayer react emoji

test/setup.jl
```diff
@@ -111,20 +111,22 @@ function force_planar(V::GradedSpace)
...
         force_planar(codomain(tsrc)) ←
             force_planar(domain(tsrc))
     )
+    tdst = similar(tsrc, force_planar(codomain(tsrc)) ← force_planar(domain(tsrc)))
     copyto!(block(tdst, PlanarTrivial()), block(tsrc, Trivial()))
     return tdst
 end
 function force_planar(tsrc::TensorMap{<:Any, <:GradedSpace})
-    tdst = TensorMap{scalartype(tsrc)}(
+    tdst = TensorKit.TensorMapWithStorage{scalartype(tsrc), storagetype(tsrc)}(
```
lkdvos: Seems like I messed up my suggestion here, but I will blame it on the wonderful GitHub UI 😢

test/setup.jl
```diff
@@ -111,20 +111,22 @@ function force_planar(V::GradedSpace)
 end
 force_planar(V::ProductSpace) = mapreduce(force_planar, ⊗, V)
 function force_planar(tsrc::TensorMap{<:Any, ComplexSpace})
-    tdst = TensorMap{scalartype(tsrc)}(
+    tdst = TensorKit.TensorMapWithStorage{scalartype(tsrc), storagetype(tsrc)}(
```
lkdvos: Also here 😿

ext/TensorKitAdaptExt.jl
kshyatt: Looks like this broke some tests

kshyatt: Or maybe they were already broken?

ext/TensorKitCUDAExt/cutensormap.jl
kshyatt: "bueffers" lol

kshyatt: This change also led to `StackOverflowError`

kshyatt:
I can't reproduce the factorization failure locally...

ext/TensorKitCUDAExt/cutensormap.jl
lkdvos: 🙃

kshyatt: It's fixed now!

ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
```diff
@@ -3,14 +3,18 @@ module TensorKitCUDAExt
 using CUDA, CUDA.CUBLAS, CUDA.CUSOLVER, LinearAlgebra
 using CUDA: @allowscalar
 using cuTENSOR: cuTENSOR
+using Strided: StridedViews
 import CUDA: rand as curand, rand! as curand!, randn as curandn, randn! as curandn!
 
+using CUDA: KernelAbstractions
+using CUDA.KernelAbstractions: @kernel, @index
```
Jutho: Is the first line needed if you qualify `KernelAbstractions` on the second line as `CUDA.KernelAbstractions`?

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,81 @@
...
-                for c in (a ⊗ b)
-                    Nsymbol(a, b, c) == Nsymbol(b, a, c) ||
-                        throw(ArgumentError("Cannot define a braiding between $a and $b"))
-                end
-            end
+    function BraidingTensor{T, S, A}(V1::S, V2::S, adjoint::Bool = false) where {T, S <: IndexSpace, A <: DenseVector{T}}
+        for a in sectors(V1), b in sectors(V2), c in (a ⊗ b)
+            Nsymbol(a, b, c) == Nsymbol(b, a, c) ||
+                throw(ArgumentError("Cannot define a braiding between $a and $b"))
         end
```
Jutho, line 25: Probably we need a more stringent check for the validity of constructing a `BraidingTensor`. But I guess we do need to allow the `BraidingStyle(I) isa NoBraiding` case if one of the spaces is a unit space only, as this is explicitly used in MPSKit, correct @lkdvos ?

lkdvos: Yes, I think so indeed. I pushed a change that should be better (I think), and at least has a clearer point of specialization

ext/TensorKitCUDAExt/TensorKitCUDAExt.jl
kshyatt: Yes apparently it is needed, sorry this was masked by all the other failures

src/tensors/braidingtensor.jl
kshyatt: Hm, this seems to have broken the CPU side `tensors/planar` tests?

src/tensors/braidingtensor.jl
```diff
@@ -19,14 +19,15 @@ struct BraidingTensor{T, S, A <: DenseVector{T}} <: AbstractTensorMap{T, S, 2, 2
     V2::S
     adjoint::Bool
     function BraidingTensor{T, S, A}(V1::S, V2::S, adjoint::Bool = false) where {T, S <: IndexSpace, A <: DenseVector{T}}
-        for a in sectors(V1), b in sectors(V2), c in (a ⊗ b)
-            Nsymbol(a, b, c) == Nsymbol(b, a, c) ||
-                throw(ArgumentError("Cannot define a braiding between $a and $b"))
-        end
+        check_valid_braiding(V1, V1) ||
```
kshyatt: This seems to many tests now failing which were passing before. Were they incorrectly passing before?

src/tensors/braidingtensor.jl
kshyatt: Also, why does this more stringent check need to be part of this PR? It seems pretty unrelated to the core point of this, and this PR has been open for nearly a month. Can this change be a separate PR?

Jutho: You're right; this is mostly a reminder to myself about potential improvements when I revisit old code.

kshyatt:
I'm still confused by this factorization test failure, which I cannot locally reproduce and only occurs for `FermionParity ⊠ Irrep[SU₂] ⊠ Irrep[U₁]`. But it seems it must be related to this PR, as it doesn't occur on `main`. I'll try to make some progress on this today.

Update: this was also failing [here](https://github.com/QuantumKitHub/TensorKit.jl/actions/runs/24657483618/job/72094576165), so maybe it's unrelated

kshyatt:
In the meantime, I removed the braiding check commit and squashed a bunch of intermediate stuff that was messing around with CUDA. Let's see how GPU CI does now...

src/planar/preprocessors.jl
```diff
@@ -83,6 +83,23 @@ _add_adjoint(ex) = Expr(TO.prime, ex)
...
+            return false
+        end
+    end
+    function extract_tensors(tensor_ex)
+        if TO.istensor(tensor_ex)
+            return [TO.decomposetensor(tensor_ex)[1]]
+        elseif TO.istensorexpr(tensor_ex)
+            return collect(Iterators.flatmap(extract_tensors, filter(filter_f, tensor_ex.args)))
+        end
+    end
```
Jutho, line 101:
```suggestion
    function extract_tensors(tensor_ex)
        if TO.istensor(tensor_ex)
            obj = TO.decomposetensor(expr)[1]
            return (_remove_adjoint(obj) == :τ) ? Any[] : Any[obj]
        elseif TO.istensorexpr(tensor_ex)
            return collect(Any, Iterators.flatmap(extract_tensors, tensor_ex.args))
        end
    end
```

kshyatt: Can't we replace the `Any` here with a `Symbol`? I think we can guarantee that `obj` on the new line 88 is a `Symbol`.

src/planar/preprocessors.jl
```diff
@@ -104,7 +121,9 @@ function _construct_braidingtensors(ex)
                 )
             end
         end
-        newrhs, success = _construct_braidingtensors!(rhs, preargs, indexmap)
+        # if this is a definition, the lhs tensor is NOT yet defined
+        no_τ_ex = reduce(vcat, Iterators.flatmap(extract_tensors, filter(filter_f, rhs.args)); init = Symbol[])
```
Jutho, line 125: I don't remember if we can really assume that all tensor objects (i.e. the expression before the indexing) are really `Symbol`s already. I know we substitute them with random variables when they are more complicated expressions, but I am not sure this preprocessing step has happened already at this point. Probably yes, but in any case, I have grown a habit of only using `Any[]` arrays in the context of manipulating macro expressions.
```suggestion
        no_τ_ex = collect(Any, Iterators.flatmap(extract_tensors, rhs.args))
```

src/planar/preprocessors.jl
```diff
@@ -115,7 +134,8 @@ function _construct_braidingtensors(ex)
     elseif TO.istensorexpr(ex)
         preargs = Vector{Any}()
         indexmap = Dict{Any, Any}()
-        newex, success = _construct_braidingtensors!(ex, preargs, indexmap)
+        no_τ_ex = reduce(vcat, Iterators.flatmap(extract_tensors, filter(filter_f, ex.args)); init = Symbol[])
```
Jutho, line 137: Same as above:

```suggestion
        no_τ_ex = collect(Any, Iterators.flatmap(extract_tensors, ex.args))
```

kshyatt: Should this really be `rhs`, or `ex`?

kshyatt: (I think `rhs` is not defined in this part of the `if-else`)

src/planar/preprocessors.jl
```diff
@@ -163,7 +183,8 @@ function _construct_braidingtensors!(ex, preargs, indexmap) # ex is guaranteed t
             end
             if foundV1 && foundV2
                 s = gensym(:τ)
-                constructex = Expr(:call, GlobalRef(TensorKit, :BraidingTensor), V1, V2)
+                storageex = Expr(:call, GlobalRef(TensorKit, :promote_storagetype), non_braiding...)
+                constructex = Expr(:call, GlobalRef(TensorKit, :BraidingTensor), storageex, V1, V2)
```
Jutho: As an alternative to needing the special purpose constructor with the storage type as a regular first argument, it could also be an option to generate the code for
`:(adapt($storageex, BraidingTensor($V1, $V2)))`

src/planar/preprocessors.jl
Jutho: Yes this was me doing copy paste.

Jutho: I've corrected it.

src/planar/preprocessors.jl
Jutho: See below; probably when the `_construct_braidingtensor` preprocessor runs, the tensor objects can be guaranteed to be `Symbol`s indeed, thanks to an earlier preprocessor that extracts more complicated expressions. Although the fact that we still need to do `_remove_adjoint` makes me think that we could still have tensor objects which are adjoint expressions, i.e. things like `A'[l1 l2 l3; r1 r2]`. I am not sure if these are also replaced.

src/planar/preprocessors.jl
kshyatt: What most people don't know is you get infinite copy-paste mulligans before noon

src/planar/preprocessors.jl
kshyatt: This would require pulling the adapt extension into the main package, wouldn't it?

src/planar/preprocessors.jl
kshyatt: In that case let's keep it as `Any` for now

src/tensors/braidingtensor.jl
```diff
@@ -115,15 +130,10 @@ end
     d = (dims(codomain(b), f₁.uncoupled)..., dims(domain(b), f₂.uncoupled)...)
     n1 = d[1] * d[2]
     n2 = d[3] * d[4]
```
Jutho: It seems `n1` and `n2` are no longer needed now:

```suggestion
```

kshyatt: I really wish GitHub would show me the lines surrounding these two!

src/planar/preprocessors.jl
Jutho: Yes, you're right. Let's keep it like this for now.

kshyatt: I'm definitely open to changing this in the future, I tried also "just" constructing the parametric type expr which was very very nasty.

src/tensors/braidingtensor.jl
```diff
@@ -134,49 +142,52 @@ TensorMap(b::BraidingTensor) = copy!(similar(b), b)
...
 
+function block(b::BraidingTensor, s::Sector)
+    I = sectortype(b)
+    I == typeof(s) || throw(SectorMismatch())
+
+    # TODO: probably always square?
+    m = blockdim(codomain(b), s)
+    n = blockdim(domain(b), s)
+
+    data = reshape(storagetype(b)(undef, m * n), (m, n))
```
Jutho, line 180: To really have exactly the same type as the blocks of a regular `TensorMap`, this should probably be

```suggestion
    data = reshape(view(storagetype(b)(undef, m * n), 1:(m * n)), (m, n))
```

though I am not sure whether that is really important.

test/setup.jl
```diff
@@ -111,20 +111,22 @@ function force_planar(V::GradedSpace)
...
         undef,
         force_planar(codomain(tsrc)) ←
             force_planar(domain(tsrc))
     )
+    tdst = similar(tsrc, force_planar(codomain(tsrc)) ← force_planar(domain(tsrc)))
```
Jutho, line 114: Something seems off here; `tdst` is constructed twice? Probably line 114 - 118 can go?

test/setup.jl
```diff
@@ -111,20 +111,22 @@ function force_planar(V::GradedSpace)
...
     return tdst
 end
 function force_planar(tsrc::TensorMap{<:Any, <:GradedSpace})
-    tdst = TensorMap{scalartype(tsrc)}(
+    tdst = TensorKit.TensorMapWithStorage{scalartype(tsrc), storagetype(tsrc)}(
         undef,
         force_planar(codomain(tsrc)) ←
             force_planar(domain(tsrc))
     )
+    tdst = similar(tsrc, force_planar(codomain(tsrc)) ← force_planar(domain(tsrc)))
```
Jutho, line 119: Same comment here.

Jutho:
Ok, this looks good to me. I haven't read through the test file in detail; I hope to get to a more massive test review and update at some later stage.

kshyatt:
Thanks! I agree we should take another swing at all these tests

src/planar/preprocessors.jl
kshyatt: Looks like something in this set of changes broke the `cuda/planar` tests.

kshyatt: OK, these changes are just not working. I will revert them and we can return to readdress later, I think.

Jutho: I am not too familiar with the `flatmap` iterator, I just tried to mimick the collect that you were doing before. Surely replacing the `filter` part with simply doing that filter at the level of `_extract_tensor` must work, no?

kshyatt: I'll try it locally

kshyatt: No, this already breaks things.

kshyatt: The reason is actually pretty obvious in the end: neither `rhs.args` nor `tensor_ex.args` are a `tensorexpr` or `tensor`. Unless there's a strong objection to keeping the working version of this, I will leave it as-is for now.

src/tensors/braidingtensor.jl
kshyatt: This also broke things because `StridedView` will not accept a `ReshapedArray` as an argument. I'll revert it.

kshyatt:
I really cannot figure out this factorization test, which I can repro failing locally on `main` and appears on various other branches. Unsure what the cause is or if it's related to this at all.

kshyatt:
Now it passes!!! I seriously wonder if the `--fast` is causing this

lkdvos:
In guessing the interaction with the RNG changes in the --fast specification . In any case, id like to have a look at this for a final round of review but might have to be later today since I still have some other things to get to

kshyatt:
Sure, makes sense to me. I'm just completely wowed by this bug if it's real, it's so silly

src/tensors/braidingtensor.jl
```diff
@@ -2,73 +2,81 @@
...
+    return BraidingTensor{T, S, A}(V[2], V[1], adjoint)
+end
 function BraidingTensor{T}(V::HomSpace, adjoint::Bool = false) where {T}
     domain(V) == reverse(codomain(V)) ||
         throw(SpaceMismatch("Cannot define a braiding on $V"))
     return BraidingTensor{T}(V[2], V[1], adjoint)
 end
-function Base.adjoint(b::BraidingTensor{T, S}) where {T, S}
-    return BraidingTensor{T, S}(b.V1, b.V2, !b.adjoint)
+# these are here to make the preprocessing for `@planar` expressions less painful
```
lkdvos, line 60: I get that we probably don't want to deal with that in the macros, but how do you feel about a setup similar to the `TensorMap` approach: there is `tensormaptype`: https://github.com/QuantumKitHub/TensorKit.jl/blob/6d02541211c41d10518ab7dcc56e982eb15ecf19/src/tensors/tensor.jl#L49-L52
which takes the role of promoting `TorA` to an actual storage type, and in the macro this would then just become `$braidingtensortype(S, TorA)(args...)`, which captures all constructors above and still retains the same structure as `TensorMap`?

One alternative, which is closer to what you have here and which I also think is okay is to go more along the route of this being a general-purpose constructor like `ones`, `rand` or `isomorphism`, `id`, which then has the syntax you have here, but maybe without capitals to distinguish it from the type constructor, i.e. `braiding` or `swap`? (although these could also be understood as creating the `TensorMap(BraidingTensor())`...

Sorry to be annoying about these constructors, it's mostly that this kind of decision is hard to undo later without breaking changes, so it is better to get that right from the beginning.

src/tensors/braidingtensor.jl
```diff
@@ -134,49 +142,52 @@ TensorMap(b::BraidingTensor) = copy!(similar(b), b)
...
         (f₁.coupled == f₂.coupled == s) || continue
         r = _braiding_factor(f₁, f₂, b.adjoint)
-        isnothing(r) && continue
         # change offset to account for single block
         subblock = StridedView(data, sz, str, off - base_offset)
-        @inbounds for i in axes(subblock, 1), j in axes(subblock, 2)
-            subblock[i, j, j, i] = r
-        end
+        # without the zero-value, the non-trivial block is not set
+        # correctly in the GPU case
```
lkdvos:
```suggestion
```

src/tensors/braidingtensor.jl
kshyatt: I will try `braidingtensortype`

kshyatt: > `$braidingtensortype(S, TorA)(args...)`

I think this has to be another `Expr(:call, ...)` but let's see

kshyatt: > Sorry to be annoying about these constructors, it's mostly that this kind of decision is hard to undo later without breaking changes, so it is better to get that right from the beginning.

I think it makes sense to be picky. As we see now, extending the old ones to new array types has not been so easy...

kshyatt: OK, `braidingtensortype` seems to have worked, thanks for the suggestion. There's a little bit of logic duplication but perhaps that is ok.

src/tensors/braidingtensor.jl
```diff
@@ -2,72 +2,85 @@
...
 
-# specializations to ignore the storagetype of BraidingTensor
-promote_storagetype(::Type{A}, ::Type{B}) where {A <: BraidingTensor, B <: AbstractTensorMap} = storagetype(B)
-promote_storagetype(::Type{A}, ::Type{B}) where {A <: AbstractTensorMap, B <: BraidingTensor} = storagetype(A)
-promote_storagetype(::Type{A}, ::Type{B}) where {A <: BraidingTensor, B <: BraidingTensor} = storagetype(A)
+# these are here to make the preprocessing for `@planar` expressions less painful
+function braidingtensortype(::Type{S}, ::Type{TorA}) where {S <: IndexSpace, TorA}
+    T = eltype(TorA)
+    TA = similarstoragetype(TorA)
+    return BraidingTensor{T, S, TA}
```
lkdvos:
```suggestion
    A = similarstoragetype(TorA)
    return BraidingTensor{scalartype(A), S, A}
```

src/tensors/braidingtensor.jl
```diff
@@ -2,72 +2,85 @@
...
         end
-        return new{T, S}(V1, V2, adjoint)
+        return new{T, S, A}(V1, V2, adjoint)
         # partial construction: only construct rowr and colr when needed
     end
 end
 function BraidingTensor{T}(V1::S, V2::S, adjoint::Bool = false) where {T, S <: IndexSpace}
-    return BraidingTensor{T, S}(V1, V2, adjoint)
+    TA = similarstoragetype(T)
+    return BraidingTensor{T, S, TA}(V1, V2, adjoint)
```
lkdvos:
```suggestion
    return braidingtensortype(S, T)(V1, V2, adjoint)
```

src/tensors/braidingtensor.jl
```diff
@@ -2,72 +2,85 @@
...
+    S = promote(V1, V2)
+    return braidingtensortype(S..., TorA)
+end
+function braidingtensortype(V::HomSpace, ::Type{TorA}) where {TorA}
+    domain(V) == reverse(codomain(V)) ||
+        throw(SpaceMismatch("Cannot define a braiding on $V"))
+    T = eltype(TorA)
+    S = spacetype(V[2])
+    TA = similarstoragetype(TorA)
+    return braidingtensortype(S, TA)
```
lkdvos:
```suggestion
    return braidingtensortype(spacetype(V), TorA)
```

kshyatt:
I've run the MPSKit tests that aren't already failing on https://github.com/QuantumKitHub/MPSKit.jl/pull/416 against this branch, everything but Aqua passes. I'm running the `algorithms` tests in detail just to make sure as I suspect that's where any problem would surface.

kshyatt:
Figured out one last MPSKit failure that was related -- an extension of `braidingtensortype` is needed at `BlockTensorKit.jl`, I'll make that PR once this is merged

src/tensors/braidingtensor.jl
```diff
@@ -2,72 +2,80 @@
...
+end
+braidingtensortype(V::S, ::Type{TorA}) where {S <: IndexSpace, TorA} = braidingtensortype(S, TorA)
+braidingtensortype(V1::S, V2::S, ::Type{TorA}) where {S <: IndexSpace, TorA} = braidingtensortype(S, TorA)
+function braidingtensortype(V1::IndexSpace, V2::IndexSpace, ::Type{TorA}) where {TorA}
+    S = promote(V1, V2)
+    return braidingtensortype(S..., TorA)
+end
+function braidingtensortype(V::HomSpace, ::Type{TorA}) where {TorA}
+    domain(V) == reverse(codomain(V)) ||
+        throw(SpaceMismatch("Cannot define a braiding on $V"))
```
Jutho: Do we want to include this check here? It might be valid to ask what the `braidingtensortype` would be for some HomSpace of some tensor, of which you are only going to braid two legs out of many. 


```suggestion
```

Jutho:
One final suggestion from my side, but this definitely looks great, so feel free to merge with or without that.

src/tensors/braidingtensor.jl
kshyatt: I included it for consistency but I agree with your proposed example. Let's remove this for now.

lkdvos (commented):
Looks good to me too, could you already increase the version to 0.17? Since this is breaking, we will have to do that anyways, so might as well and then start adding the other breaking stuff in too (CUDA etc)

kshyatt:
Done! They're calling it the end of Friday snipe commit of the century

lkdvos (approved):

PR #403, opened 2026-04-16
define `spacetype` for `TruncationSpace`

ogauthe:
This PR defines `spacetype` for `TruncationSpace` and adds a test that spacetype matches between the vector to truncate and the strategy (else it will crash with a hard to read error). I expect the check to be done at compile time without any consequence on assembly code.

### DIFF
--- a/src/factorizations/truncation.jl
+++ b/src/factorizations/truncation.jl
@@ -25,7 +25,7 @@ function truncspace(space::ElementarySpace; by = abs, rev::Bool = true)
     return TruncationSpace(space, by, rev)
 end
 
-spacetype(::Type{<:TruncationSpace{S}}) where {S} = S
+TensorKit.spacetype(::Type{<:TruncationSpace{S}}) where {S} = S
 
 # truncate!
 # ---------
--- a/src/factorizations/truncation.jl
+++ b/src/factorizations/truncation.jl
@@ -259,12 +259,12 @@ MAK.findtruncated_svd(values::SectorVector, strategy::TruncationByError) =
     MAK.findtruncated(values, strategy)
 
 function MAK.findtruncated(values::SectorVector, strategy::TruncationSpace)
-    sectortype(values) == sectortype(strategy) || throw(SpaceMismatch("sectortype of truncation strategy does not match values"))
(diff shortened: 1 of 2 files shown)


lkdvos (commented):
Definitely okay with this change, but would you mind changing the `@assert` to `throw(SpaceMismatch("spacetype of truncation strategy does not match spacetype of tensor"))` (or a better worded message if you have inspiration, just quickly typed something here)

src/factorizations/truncation.jl
```diff
@@ -257,10 +259,12 @@ MAK.findtruncated_svd(values::SectorVector, strategy::TruncationByError) =
     MAK.findtruncated(values, strategy)
 
 function MAK.findtruncated(values::SectorVector, strategy::TruncationSpace)
+    sectortype(values) == sectortype(strategy) || throw(SpaceMismatch("sectortype of truncation strategy does not match values"))
```
lkdvos:
```suggestion
    sectortype(values) == sectortype(strategy) || throw(SectorMismatch("sectortype of truncation strategy does not match values"))
```

src/factorizations/truncation.jl
```diff
@@ -257,10 +259,12 @@ MAK.findtruncated_svd(values::SectorVector, strategy::TruncationByError) =
...
     blockstrategy(c) = truncrank(dim(strategy.space, c); strategy.by, strategy.rev)
     return SectorDict(c => MAK.findtruncated(d, blockstrategy(c)) for (c, d) in pairs(values))
 end
 function MAK.findtruncated_svd(values::SectorVector, strategy::TruncationSpace)
+    sectortype(values) == sectortype(strategy) || throw(SpaceMismatch("sectortype of truncation strategy does not match values"))
```
lkdvos:
```suggestion
    sectortype(values) == sectortype(strategy) || throw(SectorMismatch("sectortype of truncation strategy does not match values"))
```

lkdvos (commented):
(you are right, `sectortype` is more appropriate :))

lkdvos (dismissed):

lkdvos (approved):

PR #404, opened 2026-04-17
Bump minimum version of CUDA and cuTENSOR

kshyatt:


lkdvos (dismissed):

kshyatt:
I'm running this locally to see if the failures for CUDA are related or not. We'll also need a TO tag first.

kshyatt:
Can't repro the factorization failures, annoying...

kshyatt:
Now needs the MAK bump which itself needs the Mooncake and Enzyme ones

kshyatt:
OK I think this is good to go now that things are passing

kshyatt:
... or not. These chainrules failures are suprising to me, it seems it should be unrelated as this PR doesn't touch that code. Could it maybe be related to the `MatrixAlgebraKit` version bump, as a lot of the failures are in factorization related tests?

lkdvos:
Actually, it turns out that the chainrules is the same issue that should be resolved here: https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/219 (at least, I assume so since it only affects the LQ decomposition)

Given that this failure is unrelated, if the tests pass we should probably try and merge this.

lkdvos (approved):

kshyatt:
Looking at the logs, all the failures are in the LQ tests mentioned above, so I am going to force-merge

PR #409, opened 2026-04-21
Various docs fixes and improvements 

lkdvos:
This should also fix the documenter failures in the current open PRs


.gitignore
```diff
@@ -5,7 +5,7 @@
 *-old
 __pycache__
 .ipynb*
-Manifest.toml
+*Manifest.toml
```
Jutho, line 8: Out of curiosity, why is this one necessary?

Jutho (approved):

.gitignore
lkdvos: The workspaces don't work in 1.10, in which case you might actually accidentally generate a manifest in the test or docs directory

Jutho: So without the asterisk "Manifest.toml" would only match with such a file at the top level and not in subdirectories? I didn't know .gitignore worked this way. Thanks.

Issue #413, opened 2026-04-23
`cond` test fails depending on whether `--fast` is used in the testsuite


kshyatt:
As discussed around [here](https://github.com/QuantumKitHub/TensorKit.jl/pull/393#issuecomment-4298123848), it's likely to do with the changed state of the RNG. This one is tough to debug for this reason, maybe we should explicitly set the RNG seeds per space?

PR #416, opened 2026-04-26
Rework index manipulation API

lkdvos:
## Summary

This PR overhauls the index manipulation API in `src/tensors/indexmanipulations.jl` to match TensorOperations dispatch conventions, reduces code duplication in the implementation, and adds a dedicated documentation page.
The goal was a bunch of code simplification, (overall number of lines reduced, even though I added some docs 🎉 )

### API changes

- **Unified in-place interface**: `permute!`, `braid!`, `transpose!`, and `repartition!` now directly accept `α`, `β`, `backend`, and `allocator` as optional arguments (with defaults `One()`, `Zero()`, `DefaultBackend()`, `DefaultAllocator()`), following the TensorOperations dispatch pattern. The old `add_permute!`, `add_braid!`, and `add_transpose!` are deprecated and forward to the new functions.
- **`allocator` support**: previously, the index manipulation functions did not support a custom `allocator` at all. It is now a positional argument in both the public and internal interfaces, consistent with TensorOperations convention.
- **Out-of-place functions** (`permute`, `braid`, `transpose`, `repartition`) gain `backend` as a new keyword argument alongside the now-supported `allocator` keyword.

### Implementation changes

- All permutation operations now route through `braid!`, eliminating duplicate codepaths.
- Dedicated `braid!` method added for `AdjointTensorMap`.
- `add_transform!` kernels for `TensorMap` refactored to operate on the raw data vector rather than the full `TensorMap`. Because the data vector has no symmetry type, this avoids recompilation for every `TensorMap` type combination, improving compilation time.
- Various minor simplifications throughout.


lkdvos:
~~This might also resolve #413, where I tried to compute `cond` with `eigh_vals` instead of `LinearAlgebra.eigvals`.~~

src/tensors/braidingtensor.jl
```diff
@@ -288,21 +288,13 @@ function planarcontract!(
...
-    )
-
-    add_braid!(
-        C, A′, (ntuple(identity, M), (N, N - 1)),
-        levels, α, β, backend,
-    )
+    p = (oindA, reverse(cindA))
+    N = length(oindA)
+    levels = (ntuple(identity, N)..., (B.adjoint ? (N + 1, N + 2) : (N + 2, N + 1))...)
+    braid!(C, A, p, levels, α, β, backend, allocator)
```
Jutho: This doesn't seem correct since it is still using `A` instead of `A′`. It is very different then the case above where `A::BraidingTensor` and `B::AbstractTensorMap`.

If this would be correct, then everything above (the comment about not a single `braid!` call and the construction of `A′` can go.

src/tensors/indexmanipulations.jl
```diff
@@ -46,33 +28,214 @@ function flip(t::AbstractTensorMap, I; inv::Bool = false)
...
+    if !all(in(allind(t)), inds)
+        msg = "Can't twist indices $inds of a tensor with only $(numind(t)) indices."
+        throw(ArgumentError(msg))
+    end
+    (scalartype(t) <: Real && !(sectorscalartype(sectortype(t)) <: Real)) &&
+        throw(ArgumentError("Can't in-place twist a real tensor with complex sector type"))
+    has_shared_twist(t, inds) && return t
+
+    (scalartype(t) <: Real && !(sectorscalartype(sectortype(t)) <: Real)) &&
+        throw(ArgumentError("No in-place `twist!` for a real tensor with complex sector type"))
```
Jutho: This seems duplicated from line 67-68 with a slightly different error message.

src/tensors/indexmanipulations.jl
```diff
@@ -46,33 +28,214 @@ function flip(t::AbstractTensorMap, I; inv::Bool = false)
...
+Apply a twist to the `i`th index of `tsrc` and return the result as a new tensor.
+If `inv = true`, use the inverse twist.
+If `copy = false`, `tdst` might share data with `tsrc` whenever possible. Otherwise, a copy is always made.
+
+See [`twist!`](@ref) for storing the result in place.
+"""
+function twist(t::AbstractTensorMap, inds; inv::Bool = false, copy::Bool = false)
+    if has_shared_twist(t, inds)
+        copy || return t
+        return copy!(similar(t), t)
```
Jutho:
```suggestion
        return copy ? Base.copy(t) : t
```

docs/src/man/indexmanipulations.md
```diff
@@ -0,0 +1,114 @@
...
+CollapsedDocStrings = true
+```
+
+```@setup indexmanip
+using TensorKit
+using LinearAlgebra
+```
+
+A `TensorMap{T, S, N₁, N₂}` is a linear map from a domain (a `ProductSpace{S, N₂}`) to a codomain (a `ProductSpace{S, N₁}`).
+In practice, the bipartition of the `N₁ + N₂` indices between domain and codomain is often not fixed: algorithms typically need to reshuffle indices between the two sides, reorder them, or change the arrow direction on individual indices before passing a tensor to a factorization or contraction.
```
Jutho:
```suggestion
In practice, the bipartition of the `N₁ + N₂` indices between domain and codomain rarely remains fixed: algorithms typically need to reshuffle indices between the two sides, reorder them, or change the arrow direction on individual indices before passing a tensor to a factorization or contraction.
```

docs/src/man/indexmanipulations.md
```diff
@@ -0,0 +1,114 @@
...
+domainind
+allind
+```
+
+## Reweighting
+
+Reweighting operations modify the entries of a tensor by applying local isomorphisms to individual indices, without changing the number of indices or their partition between domain and codomain.
+
+[`flip`](@ref) changes the arrow direction on selected indices by applying the corresponding isomorphism between a space and its dual.
+[`twist`](@ref) applies the topological spin (monoidal twist) to selected indices; for `BraidingStyle(I) == Bosonic()` this is always trivial.
```
Jutho:
```suggestion
Reweighting operations modify the entries of a tensor by applying local isomorphisms to individual indices, without changing the number of indices or their partition between domain and codomain.
In particular, [`twist`](@ref) applies the topological spin (monoidal twist) to selected indices; this operation preserves the space of the indices and is completely trivial for `BraidingStyle(I) == Bosonic()`.
In contrast, [`flip`](@ref) changes the arrow direction on selected indices by applying a (non-canonical!) isomorphism between the index space and its dual.
```

docs/src/man/indexmanipulations.md
```diff
@@ -0,0 +1,114 @@
...
+
+```@docs; canonical=false
+flip(t::AbstractTensorMap, I)
+twist(::AbstractTensorMap, ::Int)
+twist!
```
Jutho:
```suggestion
twist(::AbstractTensorMap, ::Int)
twist!
flip(t::AbstractTensorMap, I)
```

docs/src/man/indexmanipulations.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+## Inserting and removing unit spaces
+
+These functions add or remove a trivial tensor product factor at a specified index position, without affecting any other indices.
+[`insertleftunit`](@ref) inserts before position `i` and [`insertrightunit`](@ref) inserts after position `i`; [`removeunit`](@ref) undoes either insertion.
+Passing `Val(i)` instead of an `Int` for the position may improve type stability.
```
Jutho:
```suggestion
The next set of functions add or remove a trivial tensor product factor at a specified index position, without affecting any other indices.
We distinguish between [`insertleftunit`](@ref), which inserts a unit index before index `i` (the unit index becoming index `i`),
and [`insertrightunit`](@ref), which inserts after index `i` (the unit index becoming index `i + 1`);
[`removeunit`](@ref) undoes either insertion.

For tensors `t` with `UnitStyle(sectortype(t)) = SimpleUnit()`, the only relevant difference between `insertleftunit(t, i + 1)` and `insertrightunit(t, i)` is that `insertleftunit(t, numout(t) + 1)` inserts the unit index as first index in the domain, whereas `insertrightunit(t, numout(t))` will insert the unit index as last index in the codomain. 

Passing `Val(i)` instead of an integer `i` for the position may improve type stability.
```

docs/src/man/indexmanipulations.md
```diff
@@ -0,0 +1,114 @@
...
+permute!(::AbstractTensorMap, ::AbstractTensorMap, ::Index2Tuple)
+transpose(::AbstractTensorMap, ::Index2Tuple)
+transpose!
+repartition(::AbstractTensorMap, ::Int, ::Int)
+repartition!
+```
+
+## Fusing and splitting indices
+
+There is no dedicated function for fusing or splitting indices.
```
Jutho:
```suggestion
There is no dedicated functionality for fusing or splitting indices.
```

docs/src/man/indexmanipulations.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Splitting is then the adjoint of the same map:
+
+```julia
+@tensor t_split[…, i, j, …] := f'[i, j, a] * t_fused[…, a, …]
+```
+
+Using `f'` as the splitter guarantees that the round-trip is the identity, i.e. `t_split == t`.
+Using a *different* isomorphism to split would give a physically equivalent but numerically different tensor, so it is important to keep `f` and its adjoint consistent throughout a calculation.
```
Jutho:
````suggestion
In the general case there is no canonical embedding of `V1 ⊗ V2` into the fused space `V = fuse(V1 ⊗ V2)`: any two such embeddings differ by a basis transform, i.e. there is a gauge freedom.
TensorKit resolves this by requiring the user to construct an explicit isomorphism — the *fuser* — and contract it with the tensor.
One particular isomorphism can be constructed using the [`unitary](@ref) function.
It preserves norms and inner products, and has an inverse given by its adjoint. 
For a plain tensor (`sectortype(t) == Trivial`), applying this particular `unitary` is equivalent to `reshape` on the underlying array.

Fusing index `i` and `j = i+1` of a tensor `t` is then accomplished as

```julia
f = unitary(fuse(space(t, i) ⊗ space(t, j)), space(t, i) ⊗ space(t, j))
@tensor t_fused[…, a, …] := f[a, i, j] * t[…, i, j, …]
```

while splitting 

```julia
@tensor t_split[…, i, j, …] := f'[i, j, a] * t_fused[…, a, …]
```

using `f'` as the splitter guarantees that the round-trip is the identity, i.e. `t_split == t`.
````

File: docs/src/man/indexmanipulations.md [hunk 10]
```diff
@@ -0,0 +1,114 @@
...
+
+Note that tensor factorizations (SVD, QR, etc.) can be applied directly to any index bipartition without needing to fuse indices first; see [Tensor factorizations](@ref ss_tensor_factorization).
```
Jutho:
```suggestion
The resulting `unitary` is a dense `TensorMap`, and this fusion and splitting approach is not optimized for maximal performance. However, because many tensor operations including tensor factorizations (SVD, QR, etc.) can be applied without needing any fusion, we do not expect fusion and splitting to be an essential part of performance critical parts of typical tensor algorithms. 
```

docs/src/man/linearalgebra.md
```diff
@@ -0,0 +1,85 @@
+# [Basic linear algebra](@id ss_tensor_linalg)
```
Jutho: I would also make some suggestions in this file, but maybe we need a short discussion first. Clearly, this was written in a pre-VectorInterface era, so we definitely want to mention that all of VectorInterface is supported, and in fact re-exported. Do we want to keep the equivalent and low-level LinearAlgebra functionality like `axpy!` etc also exported by default, or should we get rid of those? Is there any way to "deprecate" a change in export behavior?

docs/src/man/contractions.md
```diff
@@ -0,0 +1,106 @@
+# [Tensor contractions and tensor networks](@id ss_tensor_contraction)
```
Jutho: In this file, I could perhaps directly push some changes in an attempt to finally complete the documentation?

src/tensors/indexmanipulations.jl
```diff
@@ -46,33 +28,214 @@ function flip(t::AbstractTensorMap, I; inv::Bool = false)
...
+"""
+    insertleftunit(
+            tsrc::AbstractTensorMap, i = numind(t) + 1;
+            conj = false, dual = false, copy = false
+        ) -> tdst
+
+Insert a trivial vector space, isomorphic to the underlying field, at position `i`,
+which can be specified as an `Int` or as `Val(i)` for improved type stability.
+More specifically, adds a left monoidal unit or its dual.
+
```
Jutho:
```suggestion
Insert a trivial vector space, isomorphic to the underlying field, before position `i`,
which should satisfy `1 ≤ i ≤ numind(t) + 1`
and can be specified as an `Int` or as `Val(i)` for improved type stability,
More specifically, add a left monoidal unit (or its dual) of the space associated with index `i`.
The new index appears at position `i` in the new tensor,
namely in its codomain for `1 ≤ i ≤ numout(t)` and in its domain otherwise.
```

src/tensors/indexmanipulations.jl
```diff
@@ -46,33 +28,214 @@ function flip(t::AbstractTensorMap, I; inv::Bool = false)
...
+"""
+    insertrightunit(
+            tsrc::AbstractTensorMap, i = numind(t);
+            conj = false, dual = false, copy = false
+        ) -> tdst
+
+Insert a trivial vector space, isomorphic to the underlying field, after position `i`,
+which can be specified as an `Int` or as `Val(i)` for improved type stability.
+More specifically, adds a right monoidal unit or its dual.
```
Jutho:
```suggestion
Insert a trivial vector space, isomorphic to the underlying field, after position `i`,
which should satisfy `0 ≤ i ≤ numind(t)`
and can be specified as an `Int` or as `Val(i)` for improved type stability,
More specifically, add a right monoidal unit (or its dual) of the space associated with index `i`.
The new index appears at position `i+1` in the new tensor,
namely in its codomain for `0 ≤ i ≤ numout(t)` and in its domain otherwise.
```

src/tensors/indexmanipulations.jl
```diff
@@ -46,33 +28,214 @@ function flip(t::AbstractTensorMap, I; inv::Bool = false)
...
     )
-    return add_permute!(tdst, tsrc, p, One(), Zero())
+    @boundscheck spacecheck_transform(permute, tdst, tsrc, p)
+    levels = ntuple(identity, numind(tsrc))
+    return @inbounds braid!(tdst, tsrc, p, levels, α, β, backend, allocator)
 end
 
 """
-    permute(tsrc::AbstractTensorMap, (p₁, p₂)::Index2Tuple; copy::Bool = false) -> tdst::TensorMap
+    permute(tsrc, (p₁, p₂)::Index2Tuple; copy = false, [backend], [allocator]) -> tdst
```
Jutho: For analogy with `braid` 

```suggestion
    permute(
        tsrc, (p₁, p₂)::Index2Tuple; copy = false,
        backend = DefaultBackend(), allocator = DefaultAllocator()
    ) -> tdst::TensorMap    
```

src/tensors/indexmanipulations.jl
```diff
@@ -115,307 +279,216 @@ function has_shared_permute(t::AdjointTensorMap, (p₁, p₂)::Index2Tuple)
...
 end
 
 """
-    braid(tsrc::AbstractTensorMap, (p₁, p₂)::Index2Tuple, levels::IndexTuple;
-          copy::Bool = false)
-        -> tdst::TensorMap
+    braid(
+            tsrc, (p₁, p₂)::Index2Tuple, levels::IndexTuple; copy = false,
+            backend = DefaultBackend(), allocator = DefaultAllocator()
+        ) -> tdst::TensorMap
```
Jutho:
```suggestion
    braid(
        tsrc, (p₁, p₂)::Index2Tuple, levels::IndexTuple; copy = false,
        backend = DefaultBackend(), allocator = DefaultAllocator()
    ) -> tdst::TensorMap
```

src/tensors/indexmanipulations.jl
```diff
@@ -115,307 +279,216 @@ function has_shared_permute(t::AdjointTensorMap, (p₁, p₂)::Index2Tuple)
...
 end
 
 """
-    transpose(tsrc::AbstractTensorMap, (p₁, p₂)::Index2Tuple;
-              copy::Bool=false)
-        -> tdst::TensorMap
+    transpose(
+            tsrc, (p₁, p₂)::Index2Tuple; copy = false,
+            backend = DefaultBackend(), allocator = DefaultAllocator()
+        ) -> tdst::TensorMap
```
Jutho:
```suggestion
    transpose(
        tsrc, (p₁, p₂)::Index2Tuple; copy = false,
        backend = DefaultBackend(), allocator = DefaultAllocator()
    ) -> tdst::TensorMap
```

src/tensors/indexmanipulations.jl
```diff
@@ -115,307 +279,216 @@ function has_shared_permute(t::AdjointTensorMap, (p₁, p₂)::Index2Tuple)
...
+        tdst::AbstractTensorMap, tsrc::AbstractTensorMap,
+        α::Number = One(), β::Number = Zero(),
+        backend::AbstractBackend = TO.DefaultBackend(), allocator = TO.DefaultAllocator()
+    )
     check_spacetype(tdst, tsrc)
     numind(tsrc) == numind(tdst) ||
         throw(ArgumentError("tsrc and tdst should have an equal amount of indices"))
     all_inds = (codomainind(tsrc)..., reverse(domainind(tsrc))...)
     p₁ = ntuple(i -> all_inds[i], numout(tdst))
     p₂ = reverse(ntuple(i -> all_inds[i + numout(tdst)], numin(tdst)))
```
Jutho: Are these types of closures now type stable? I would be inclined to write something like


```suggestion
    p₁, p₂ = let all_inds = (codomainind(tsrc)..., reverse(domainind(tsrc))...)
        ntuple(i -> all_inds[i], numout(tdst)), reverse(ntuple(i -> all_inds[i + numout(tdst)], numin(tdst)))
    end
```

src/tensors/indexmanipulations.jl
```diff
@@ -115,307 +279,216 @@ function has_shared_permute(t::AdjointTensorMap, (p₁, p₂)::Index2Tuple)
...
 @constprop :aggressive function repartition(
-        t::AbstractTensorMap, N₁::Int, N₂::Int = numind(t) - N₁; copy::Bool = false
+        t::AbstractTensorMap, N₁::Int, N₂::Int = numind(t) - N₁;
+        copy::Bool = false, backend = TO.DefaultBackend(), allocator = TO.DefaultAllocator()
     )
     N₁ + N₂ == numind(t) ||
         throw(ArgumentError("Invalid repartition: $(numind(t)) to ($N₁, $N₂)"))
     all_inds = (codomainind(t)..., reverse(domainind(t))...)
     p₁ = ntuple(i -> all_inds[i], N₁)
     p₂ = reverse(ntuple(i -> all_inds[i + N₁], N₂))
```
Jutho: Same question / suggestion. Though I now see this is not really part of this PR as this code was unmodified:

```suggestion
    p₁, p₂ = let all_inds = (codomainind(t)..., reverse(domainind(t))...)
        ntuple(i -> all_inds[i], N₁), reverse(ntuple(i -> all_inds[i + N₁], N₂))
    end
```

src/tensors/indexmanipulations.jl
```diff
@@ -447,82 +520,46 @@ end
...
-@propagate_inbounds function add_transpose!(
-        tdst::AbstractTensorMap, tsrc::AbstractTensorMap, p::Index2Tuple,
-        α::Number, β::Number, backend::AbstractBackend...
-    )
-    @boundscheck spacecheck_transform(transpose, tdst, tsrc, p)
-    transformer = treetransposer(tdst, tsrc, p)
-    return @inbounds add_transform!(tdst, tsrc, p, transformer, α, β, backend...)
-end
+# Deprecated add_*! wrappers
+# --------------------------
```
Jutho, line 531: It does seem like the `add_braid!` and friends were never exported, though they were documented. So I guess it is nice to deprecate them indeed.

src/tensors/indexmanipulations.jl
```diff
@@ -537,267 +574,172 @@ function use_threaded_transform(t::AbstractTensorMap, transformer)
...
+"""
+    adapt_transformer(U::AbstractMatrix, ::Type{A})
 
-    return nothing
-end
+Return a version of the basis transformation `U` that is compatible for storage type `A`.
+Default is a no-op.
+Backends (e.g. CUDA, AMDGPU) should overload this for their vector types to ensure the recoupling matrix `U` is on the correct device.
+"""
+adapt_transformer(U::AbstractMatrix, ::Type{A}) where {A} = U
```
Jutho: Would it make sense to already do some scalar type conversion, such that e.g. a `U::Matrrix{Float64}` is converted to the same scalar type of `A`, and `mul!` can use BLAS?

docs/src/man/linearalgebra.md
lkdvos: I'm not sure how that typically works, I definitely agree that we want to just keep the implementation for convenience but not necessarily advocate this, and we should definitely not be re-exporting these functions. In any case we have a breaking version now, so I think we can just delete the exports and keep the method definitions?

Happy to discuss anywhere next week though!

src/tensors/indexmanipulations.jl
lkdvos: It does seem like they are, I at some point checked with the new tool for boxed variables in TensorKit and none of these were flagged, but I'm happy to rewrite this.

src/tensors/indexmanipulations.jl
lkdvos: I gave this a try by directly using `adapt(A, U)`.

There is another question of whether we want to cache the adapted versions as well, to avoid repeatedly having to convert this, but that of course runs into the issue that we probably don't want to do that for GPU arrays...

src/tensors/braidingtensor.jl
```diff
@@ -244,9 +252,9 @@ function planarcontract!(
         ntuple(Returns(3), N - 2)...,
     )
 
-    add_braid!(
+    braid!(
         C, B′, ((2, 1), ntuple(i -> i + 2, N - 2)),
-        levels, α, β, backend,
+        levels, α, β, backend, allocator
     )
```
Jutho, line 258: I do think there are many more cases where `braid!` with properly constructed `levels` can do this operation without the need for constructing `B′`. I think a direct `braid!` call would only be problematic if `cindB` includes both index `1` and index `N_1+1` (i.e. the leftmost codomain and leftmost domain index). But I would need to revisit the whole planaroperations framework with al lof its assumptions and conventions to be sure, so I am fine with this for now.

lkdvos:
I've separated out the docs changes in an attempt to not conflate the two distinct changes in this PR. I think right now everything here should pass, so it should be ready for a round of review, before it can be merged.
The associated docs changes will then be for a separate PR.

src/tensors/indexmanipulations.jl
```diff
@@ -46,33 +28,221 @@ function flip(t::AbstractTensorMap, I; inv::Bool = false)
     return t′
 end
 
+# ---------
+# twist(!)
+# ---------
+function has_shared_twist(t, inds)
```
kshyatt, line 34: Maybe a stylistic thing but could this `if-else` be a set of 3 functions that dispatch on the `BraidingStyle`? Looks a little nicer to me

lkdvos: I agree, but I think Jutho in general prefers the other way, so this is more consistent with the rest of the codebase ;)

Jutho: I just prefer the shortest code that leads to optimal compiled code, and possibly I don't like to have to manually lower things like

```julia
has_shared_twist(t, inds) = _has_shared_twist(t, inds, BraidingStyle(sectortype(t)))
```

and then writing three separate function definitions. But if the change is less lines of code, than I am all for it.

lkdvos: Will just keep this as-is for now :)

src/tensors/indexmanipulations.jl
```diff
@@ -115,307 +286,216 @@ function has_shared_permute(t::AdjointTensorMap, (p₁, p₂)::Index2Tuple)
...
+function braid(
+        t::AdjointTensorMap, (p₁, p₂)::Index2Tuple, levels::IndexTuple;
+        kwargs...
+    )
+    p₁′ = adjointtensorindices(t, p₂)
+    p₂′ = adjointtensorindices(t, p₁)
+    perm = adjointtensorindices(adjoint(t), ntuple(identity, numind(t)))
+    levels′ = TupleTools.getindices(levels, perm)
+    return adjoint(braid(adjoint(t), (p₁′, p₂′), levels′; kwargs...))
 end
```
Jutho, line 352: Is this a completely new definition. I am wondering about its correctness, in particular with respect to the definition of `levels′`. Given that the adjoint of an overbraid is an underbraid, it might be that we want to change levels to map(-, levels), in combination with applying the permuation `perm`?

Jutho: Ok in some simple example that I tested it with, the current implementation is correct, so never mind.

src/tensors/indexmanipulations.jl
```diff
@@ -537,267 +577,141 @@ function use_threaded_transform(t::AbstractTensorMap, transformer)
...
-                local_counter > nblocks && break
-                @inbounds subtransformer = transformer.data[local_counter]
-                if length(subtransformer[1]) == 1
-                    _add_transform_single!(tdst, tsrc, p, subtransformer, α, β, backend...)
-                else
-                    _add_transform_multi!(tdst, tsrc, p, subtransformer, buffers, α, β, backend...)
-                end
+            # 2. Recoupling: buffer_dst = buffer_src * U^T  (each output tree is a linear
+            #    combination of input trees weighted by the recoupling coefficients).
+            U′ = Adapt.adapt(typeof(data_dst), StridedView(U))
```
Jutho, line 701: What is the point of wrapping `U` in a `StridedView` here before the `adapt` call? `data_dst` is not a `StridedView` at this point, right? Is this equivalent to `StridedView(Adapt.adapt(typeof(data_dst), U))`?

src/tensors/indexmanipulations.jl
```diff
@@ -537,267 +577,141 @@ function use_threaded_transform(t::AbstractTensorMap, transformer)
...
+            buffer = @lock buffer_lock TO.tensoralloc(typeof(data_dst), blocksize * (rows + cols), Val(true), allocator)
+            ptriv = (ntuple(identity, length(sz_src)), ())
+            buffer_dst = StridedView(buffer, (blocksize, rows), (1, blocksize), 0)
+            buffer_src = StridedView(buffer, (blocksize, cols), (1, blocksize), blocksize * rows)
+
+            # 1. Extract: copy each source block into column i of buffer_src as a flat vector,
+            #    using a trivial permutation so the layout is canonical before the matmul.
+            @inbounds for (i, struct_src_i) in enumerate(structs_src)
+                TO.tensoradd!(
+                    sreshape(buffer_src[:, i], sz_src), StridedView(data_src, sz_src, struct_src_i...),
```
Jutho: This contains a `getindex` that relies on StridedView producing a view, so if we ever change this behavior, this is where we will have to be careful.

src/tensors/indexmanipulations.jl
```diff
@@ -537,267 +577,141 @@ function use_threaded_transform(t::AbstractTensorMap, transformer)
...
+            blocksize = prod(sz_src)
+            buffer = @lock buffer_lock TO.tensoralloc(typeof(data_dst), blocksize * (rows + cols), Val(true), allocator)
+            ptriv = (ntuple(identity, length(sz_src)), ())
+            buffer_dst = StridedView(buffer, (blocksize, rows), (1, blocksize), 0)
+            buffer_src = StridedView(buffer, (blocksize, cols), (1, blocksize), blocksize * rows)
+
+            # 1. Extract: copy each source block into column i of buffer_src as a flat vector,
+            #    using a trivial permutation so the layout is canonical before the matmul.
+            @inbounds for (i, struct_src_i) in enumerate(structs_src)
+                TO.tensoradd!(
```
Jutho, line 693: Does it make sense to simply call `copy!` for the `ptriv` case?

src/tensors/indexmanipulations.jl
```diff
@@ -537,267 +577,141 @@ function use_threaded_transform(t::AbstractTensorMap, transformer)
...
+
+        if length(U) == 1 # Degenerate block with a single tree: no matmul needed.
+            coeff = only(U)
+            TO.tensoradd!(
+                StridedView(data_dst, sz_dst, only(structs_dst)...),
+                StridedView(data_src, sz_src, only(structs_src)...),
+                p, false, α * coeff, β, backend, allocator
+            )
+        else # Multi-tree block: pack → recoupling matmul → unpack.
+            rows, cols = size(U)
```
Jutho, line 683: Is there ever a case where this is not square? If so, does it make sense to do the trivial permutation on the largest of the two (src vs dest), and the non-trivial permutation on the smallest?

src/tensors/indexmanipulations.jl
```diff
@@ -537,267 +577,141 @@ function use_threaded_transform(t::AbstractTensorMap, transformer)
...
+
+            # 3. Insert: scatter column i of buffer_dst into the destination, applying the
+            #    actual index permutation p in the same tensoradd! call.
+            @inbounds for (i, struct_dst_i) in enumerate(structs_dst)
+                TO.tensoradd!(
+                    StridedView(data_dst, sz_dst, struct_dst_i...), sreshape(buffer_dst[:, i], sz_src),
+                    p, false, α, β, backend, allocator
+                )
             end
+            @lock buffer_lock TO.tensorfree!(buffer, allocator)
```
Jutho, line 712: I have forgotten how the allocators work. It is now fine that we free buffers in on order that is not the exact reverse as the order in which they were allocated?

src/tensors/indexmanipulations.jl
```diff
@@ -537,267 +577,141 @@ function use_threaded_transform(t::AbstractTensorMap, transformer)
...
-        ::FusionStyle, tdst, tsrc, p, transformer::GenericTreeTransformer, α, β, backend...
-    )
-    # preallocate buffers
-    buffers = allocate_buffers(tdst, tsrc, transformer)
+    cp = TO.allocator_checkpoint!(allocator)
+
+    # buffers have to be created without race condition: err on the side of caution with a lock
+    buffer_lock = Threads.ReentrantLock()
+
+    OhMyThreads.@tasks for src in fusionblocks(tsrc)
```
Jutho, line 598: There is some code duplication between this generic implementation, and the one for `GenericTreeTransformer` below (hence my questions there also apply here). But I think the code duplication is unavoidable (and anyway quite limited).

src/tensors/treetransformers.jl
```diff
@@ -227,3 +203,9 @@ end
 function _transformer_weight((mat, structs_dst, structs_src)::GenericTransformerData)
     return length(mat) * prod(structs_dst[1])
 end
+
+function buffersize(transformer::GenericTreeTransformer)
+    return maximum(transformer.data; init = 0) do (basistransform, structures_dst, _)
+        return prod(structures_dst[1]) * size(basistransform, 1)
+    end
+end
```
Jutho: Is this function used anywhere? I couldn't find a single calling instance? The `add_transform_kernel!` computes the buffer size manually.

Jutho (approved):
Ok, this looks great. I've left a few questions, but mostly just to get better understanding.

src/tensors/indexmanipulations.jl
lkdvos: I did end up adding some test for this since indeed it is a new implementation, I think I've convinced myself that while indeed you want underbraids on the adjoint, this ends up being the case because the index order gets reversed

Jutho: What index order? I am still confused. The index mappings are such that you do the corresponding braid on the adjoint tensor? In what sense is that reversing index order? 

As a side note (since I like derailing PRs 😄 ), one thing I noticed in trying to test this, is that `copy(::AdjointTensor)` produces a new `::AdjointTensor`. This is different from `Matrix`, where `copy` of an adjoint produces a regular `Matrix`. Also `TensorMap(::AdjointTensorMap)` doesn't work, so it took me a while to find a good way to reinstantiate an `AdjointTensorMap` as its corresponding `TensorMap`.

src/tensors/indexmanipulations.jl
lkdvos: It is in spirit, the reason for the change in order is that storagetype yields something with ndims = 1, while U is a Matrix, so we have the freedom to be slightly more liberal with the strided implementation but the standard CuArray doesn't capture that

src/tensors/indexmanipulations.jl
lkdvos:
```julia
julia> t'
2←2 adjoint(::TensorMap{Float64, ComplexSpace, 1, 1, Vector{Float64}}):
 codomain: ⊗(ℂ^2)
 domain: ⊗(ℂ^2)
 blocks:
 * Trivial() => 2×2 adjoint(reshape(view(::Vector{Float64}, 1:4), 2, 2)) with eltype Float64:
 0.53278   0.225126
 0.582244  0.550174

julia> copy(t')
2←2 TensorMap{Float64, ComplexSpace, 1, 1, Vector{Float64}}:
 codomain: ⊗(ℂ^2)
 domain: ⊗(ℂ^2)
 blocks:
 * Trivial() => 2×2 reshape(view(::Vector{Float64}, 1:4), 2, 2) with eltype Float64:
 0.53278   0.225126
 0.582244  0.550174
```

I can't seem to reproduce that, as I think for me this (correctly?) returns a `TensorMap`? The tests depend on this behavior so this is kind of important:

```julia
t1 = copy(braid(t', p, levels))
t2 = braid(copy(t'), p, levels)
@test isapprox(t1, t2)
```

src/tensors/indexmanipulations.jl
lkdvos: I don't think there is actually, these are always unitary (basis transformations) so they have to be square, otherwise would have been a good idea!

src/tensors/indexmanipulations.jl
lkdvos: It does, I think the main reason I kept it is to be slightly more uniform with handling backends, e.g. if we want to play with HPTT or cuTENSOR it's more convenient to keep this uniform I think.

src/tensors/indexmanipulations.jl
lkdvos: The `tensorfree!` calls are inserted as soon as the temporary object is no longer required, so they are never in that order. The `allocator_reset!` calls on the other hand do have that restriction, but that should be fine here I think

PR #418, opened 2026-04-26
Braidingtensor `planarcontract!` fixes

lkdvos:
This was overlooked in the release, and unfortunately now it is a bit annoying to backport this to 0.16 (although we may discuss if this could be worth it, as a true bugfix).

To do:
- [x] Add tests that actually cover this code.


src/tensors/braidingtensor.jl
```diff
@@ -229,91 +227,57 @@ end
...
-            TO.tensoradd!(
-                C[f₁′, f₂′], B[f₁, f₂], (reverse(cindB), oindB), false,
-                α * coeff, One(), backend, allocator
-            )
-        end
-    end
+    τ_level_1 = τ_levels[cindA[1]]
+    τ_level_2 = τ_levels[cindA[2]]
+    levels_B = ntuple(i -> i == cindB[1] ? τ_level_1 : i == cindB[2] ? τ_level_2 : 3, numind(B))
+    add_braid!(C, B, (reverse(cindB)..., oindB), levels_B, α, β, backend)
```
Jutho: I am confused by the splatting of `reverse(cindB)...`. Don't we want an `Index2Tuple`. See also old line 281 `TO.tensoradd!(C[f₁′, f₂′], B[f₁, f₂], (reverse(cindB), oindB), false, α * coeff, One(), backend, allocator)`.

src/tensors/braidingtensor.jl
```diff
@@ -229,91 +227,57 @@ end
...
-
-    p = (oindA, reverse(cindA))
-    N = length(oindA)
-    levels = (ntuple(identity, N)..., (B.adjoint ? (N + 1, N + 2) : (N + 2, N + 1))...)
-    return add_braid!(C, A, p, levels, α, β, backend)
+    τ_levels = B.adjoint ? (1, 2, 2, 1) : (2, 1, 1, 2)
+    τ_level_1 = τ_levels[cindB[1]]
+    τ_level_2 = τ_levels[cindB[2]]
+    levels_A = ntuple(i -> i == cindA[1] ? τ_level_1 : i == cindA[2] ? τ_level_2 : 3, numind(A))
+    add_braid!(C, A, (oindA, reverse(cindA)), levels_A, α, β, backend)
```
Jutho: In relation to the comment above, there is no `...` here. Does this indicate that the previous `...` is a bug that is not detected by the tests?

src/tensors/braidingtensor.jl
lkdvos: You are right, this is a bug :)

src/tensors/braidingtensor.jl
lkdvos: it is actually detected, the tests were failing.

Jutho: I am very confused why this would not be caught by the tests, as the tests should yield `planarcontract!` calls where the BraidingTensor is the first argument.

Jutho: Oh, I didn't see your comment yet.

lkdvos: it **was** caught by the tests, see e.g. https://github.com/QuantumKitHub/TensorKit.jl/actions/runs/25078533126/job/73513970689#step:6:1535, I just hadn't realized this yet because it was only introduced in my final clean up where I started using `pA`, `pB` and `pAB`.

lkdvos: Anyways, hopefully should be fixed now!

Jutho (dismissed):

lkdvos:
So, it turns out that I made a mistake in simply replacing the entire call with `add_braid!`, since that is not actually equivalent to `transpose` + `artin_braid`, which is what the `BraidingTensor` is supposed to do.
This caused me a lot of painful debugging, but I think I now figured it out :).

I know that we could still consider merging this by writing a new fusiontree manipulation that does this, but unless this is really showing up in the profiler I would argue that it is fine to leave this as is for now.

lkdvos:
Failures are unrelated (still the MatrixAlgebraKit issues), so will merge this as-is.

Jutho:
> So, it turns out that I made a mistake in simply replacing the entire call with `add_braid!`, since that is not actually equivalent to `transpose` + `artin_braid`, which is what the `BraidingTensor` is supposed to do. This caused me a lot of painful debugging, but I think I now figured it out :).

I should have known this. I had some hunch while reviewing that something might be off, and was looking for the effect of the final `pAB` permutation, but then I stopped worrying about it and assumed that it would be fine, with the permutation really only doing the swap and therefore `add_braid!` being the right primitive. But indeed, it is not.

lkdvos:
Yeah, it is very confusing since the `@planar` macro actually only generates calls where the `pAB` turns out to be trivial, which is why I somehow assumed it would be fine.
I at least put in some comments to prevent me (and anyone else in the future) from making the same mistake again :)

PR #419, opened 2026-04-27
`remove_gauge_dependence!` overloads

lkdvos:
After MatrixAlgebraKit v0.6.6, these functions are now simply overloaded and the block implementation can be re-used


lkdvos:
~~Needs https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/221~~
Bypassed this by using default arguments

src/factorizations/pullbacks.jl
```diff
@@ -51,8 +50,61 @@ for pullback_trunc! in (:svd_trunc_pullback!, :eig_trunc_pullback!, :eigh_trunc_
...
+    @eval function MAK.$remove_f_gauge_dependence!(ΔV::AbstractTensorMap, D, V; kwargs...)
+        foreachblock(ΔV, D, V) do c, (Δv, d, v)
+            MAK.$remove_f_gauge_dependence!(Δv, d, v; kwargs...)
+            return nothing
+        end
+        return ΔV
+    end
+    @eval function MAK.$remove_f_gauge_dependence!(ΔV::AbstractTensorMap, D, V, inds; kwargs...)
+        foreachblock(ΔV, D, V) do c, (Δv, d, v)
+            haskey(inds, c) || return nothing
```
Jutho: I assume that if `!haskey(inds, c)`, then indeed `ΔV` doesn't have such a block either and `Δv` is just an empty matrix? But due to the fact that a `c` block can exist in `V` and `D` and the union semantics of `foreachblock`, such `c` would still be generated by the iterator.

lkdvos: Yes, although actually in the current implementation this never happens since all sectors are always present in the `inds` that come from the truncation implementations, possibly with `inds[c]` empty.

lkdvos:
Fun fact, apparently the wording in the  MatrixAlgebraKit linked PR auto closed this one ??

Jutho:
Ok, the argument composition of the `remove_x_gauge_dependence` functions is apparently quite inconsistent, which I did not realise. That is not very nice, but will unfortunately require yet another breaking update of MAK. Anyway, independent of this PR which looks good.

Jutho (approved):

PR #420, opened 2026-04-28
Stabilize `cond` tests by using `randn`

### LINKED ISSUE #413: `cond` test fails depending on whether `--fast` is used in the testsuite

lkdvos:
Not sure if this makes a large difference, but this perhaps could give better conditioned matrices?

### DIFF
--- a/test/factorizations/svd.jl
+++ b/test/factorizations/svd.jl
@@ -21,9 +21,9 @@ for V in spacelist
         @testset "Condition number and rank" begin
             for T in eltypes,
                     t in (
-                        rand(T, W, W), rand(T, W, W)',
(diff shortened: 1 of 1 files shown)


lkdvos:
Victory 🎉

kshyatt (approved):
hoorah

Jutho:
Approved until the next change in the Julia RNGs that requires us to switch `randn -> rand` again 😄 .

PR #422, opened 2026-04-28
Implement `DefaultAlgorithm` support

### LINKED ISSUE #405: Issue with `DefaultAlgorithm` for `left_orth` on `TensorMap`
Calling `left_orth` on a `TensorMap` using `DefaultAlgorithm` to pass the algorithm settings doesn't seem to work right now:

```julia
using MatrixAlgebraKit
using TensorKit

a = randn(ComplexF64, ComplexSpace(3), ComplexSpace(3))
Q, R = left_orth(a, DefaultAlgorithm(; positive = true))
```
throws
```
ERROR: Unknown or invalid `left_orth` algorithm type `DefaultAlgorithm{@NamedTuple{positive::Bool}}`.
To register the algorithm type for `left_orth`, define

    MatrixAlgebraKit.left_orth_alg(alg::CustomAlgorithm) = LeftOrthAlgorithm{kind}(alg)

where `kind` selects the factorization type that will be used.
By default, this is either `:qr`, `:polar` or `:svd`, to select [`qr_compact!`](@ref),
[`left_polar!`](@ref), [`svd_compact!`](@ref) or [`svd_trunc!`](@ref) respectively.
```

Doing the same on an `AbstractMatrix` works fine on the other hand, indicating that the `DefaultAlgorithm` intercepts for `TensorMap`s still have some issues. Not sure what the problem is, since other decompositions such as
```julia
U, S, V = svd_compact(a, DefaultAlgorithm())
```
work fine.

Wasn't sure whether to open this here or in MatrixAlgebraKit.jl, I can move it if that would be more suitable.

lkdvos:
Fixes #405

### DIFF
--- a/src/factorizations/matrixalgebrakit.jl
+++ b/src/factorizations/matrixalgebrakit.jl
@@ -64,6 +64,27 @@ end
 
 MAK.zero!(t::AbstractTensorMap) = zerovector!(t)
 
+# Default algorithm
+# -----------------
+for f in [
+        :lq_full, :lq_compact, :lq_null,
+        :qr_full, :qr_compact, :qr_null,
+        :schur_full, :schur_vals,
+        :eig_full, :eig_vals, :eig_trunc, :eig_trunc_no_error,
+        :eigh_full, :eigh_vals, :eigh_trunc, :eigh_trunc_no_error,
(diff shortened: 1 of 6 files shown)


lkdvos:
Test failures are due to the MatrixAlgebraKit issues, should be unrelated.

Jutho (approved):

PR #426, opened 2026-05-11
Fix for SUNRepresentations type parameter change

leburgel:
Small fix in the symmetric tensor tutorial for SUNRepresentations v0.4, where `SUNIrrep{3}` has become an abstract type. It anyway made no sense to mix `SU3Irrep` and `SUNIrrep{3}` in the same line, looks needlessly confusing.

If we want, I can also bump the SUNRepresentations compat entry to v0.4 in the docs project, so we can use `SUNRepresentations.casimir` instead of having to define our own `casimir` method. It's quite pedagogical the way it is so I didn't do this right away, but if this would be better I'll change it.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -16,6 +16,6 @@ TensorKit = {path = ".."}
 [compat]
 Documenter = "1"
 DocumenterInterLinks = "1"
-SUNRepresentations = "0.3"
+SUNRepresentations = "0.4"
 Test = "1"
 WignerSymbols = "1,2"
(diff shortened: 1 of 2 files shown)


lkdvos:
I would actually be in favor of using `SUNRepresentations.casimir` as that helps visibility of that method, and I think the pedagogical part is more about using it than defining it?

Btw, could you then also update the [compat] in the docs Project.toml?

lkdvos (approved):

PR #429, opened 2026-05-13
`isdual` for AbstractTensorMaps

Yue-Zhengyuan:
This PR adds a utility function `isdual(x::AbstractTensorMap)` that returns a Boolean vector indicating which spaces of a tensor map are dual spaces. For example:
```julia
x = randn(ℂ^2 ⊗ (ℂ^3)' ← (ℂ^1)' ⊗ ℂ^4)
isdual(x) # returns [0, 1, 0, 1] as a Boolean vector
```
It is mainly used to assert that arrows in a network are preserved, which I frequently need to do.

The name of the function can be changed if you feel that `isdual` must return a single `Bool`. I don't know how to meaningfully test it, either.

### DIFF
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -308,6 +308,15 @@ function adjointtensorindices(t, p::Index2Tuple)
     return (adjointtensorindices(t, p[1]), adjointtensorindices(t, p[2]))
 end
 
+"""
+    isdual(x::AbstractTensorMap) -> Vector{Bool}
+
+Returns a Boolean vector indicating which spaces of `x` are dual spaces.
+"""
+function isdual(x::AbstractTensorMap)
+    return [isdual(space(x, i)) for i in 1:numind(x)]
+end
+
 # tensor characteristics: work on instances and pass to type
 #------------------------------------------------------------
 InnerProductStyle(t::AbstractTensorMap) = InnerProductStyle(typeof(t))
--- a/test/tensors/construction.jl
+++ b/test/tensors/construction.jl
@@ -24,6 +24,7 @@ for V in spacelist
                 @test space(t) == (W ← one(W))
                 @test domain(t) == one(W)
                 @test typeof(t) == TensorMap{T, spacetype(t), 5, 0, Vector{T}}
+                @constinferred isdual(t)
                 # Array type input
                 t = @constinferred zeros(Vector{T}, W)
                 @test @constinferred(hash(t)) == hash(deepcopy(t))


Jutho:
I am not a great fan of calling `isdual` on a tensor, since that might have other connotations as well. One way forward that I could see is the following, but I would certainly like the opinion of @lkdvos .

Firstly, we could have a method:

```julia
spaces(t::AbstractTensorMap) = (codomain(t)..., map(dual, domain(t))...)
```

and then you can simply use broadcasting to write `isdual.(spaces(t))`. Would that be a sufficient shorthand?

Jutho:
Also, in relation to this and #430 , while `map( f, ::ProductSpace)` now indeed returns a tuple, broadcasting `f.(::ProductSpace)` still returns a vector, so we probably should also fix the broadcasting behavior of `ProductSpace` objects.

Yue-Zhengyuan:
> you can simply use broadcasting to write `isdual.(spaces(t))`. Would that be a sufficient shorthand?

This will be convenient enough for me 👍. Or maybe even better, instead of write a new `spaces`, we can define how to iterate over `TensorMapSpace`.

> broadcasting f.(::ProductSpace) still returns a vector

It appears to have a simple solution: explicitly define
```julia
Base.Broadcast.broadcastable(P::ProductSpace) = P.spaces
```
Otherwise it defaults to `collect(x)`, resulting in a vector.

lkdvos:
Summarizing my earlier comment, I would like something like this: `isdual.(codomain(t)), isdual.(domain(t))` (with appropriate negations), but I don't have strong opinions on adding a utility function to replace that if this is really desired (if we can come up with a good name).

The following also would make sense to me: in Base we often have the pattern `size(A, i) = size(A)[i]`, so if we have `space(t, i) = space(t)[i]` (which I think we already do?), I would be fine making `isdual.(space(t))` broadcast over the `codomain, dual(domain)` spaces in a linear way.

Yue-Zhengyuan:
> if we have `space(t, i) = space(t)[i]` (which I think we already do?), I would be fine making `isdual.(space(t))` broadcast over the `codomain, dual(domain)` spaces in a linear way.

We do have `space(t, i) = space(t)[i]`. The only thing preventing `isdual.(space(t))` is the lack of `Base.iterate` for `HomSpace`. Also be careful that `dual(domain)` will reverse the leg order in the domain, which I don't want in `isdual.(space(t))`.

Jutho:
Given that we already support `getindex` on `HomSpace`, I would be supportive of both broadcasting and iteration support for `ProductSpace` and `HomSpace`.

Can you implement the necessary functionality for this in the current PR?

Issue #428, opened 2026-05-12
AMDGPU tests are OOMing the runners


kshyatt:
Thanks @luraess for surfacing this

We're being bad neighbors in the AMD GPU CI, our tests are tooooooo memory heavy and it's OOMing some of the Docker runners.

> Seems it allocs about 60 GB if not wrong. Runners 1 has "only" 32 GB. Maybe that's one of the reason. Dunno tho as in CUDA tests memory usage spikes at 80 GB. But maybe the runner can accommodate this for CUDA

Quick fix: restrict to target `gfx1101` in our runners.

Could be a good forcing function to make our tests more modular and do what the GPUArrays testsuite does.

luraess:
I see you are using ParallelTestRunner.jl to orchestrate the tests. A first straight forward fix could be to just split the unique testfile for AMDGPU into a few smaller ones which would then be able to run in parallel on various runners an benefit from the implicit resource management built into the runner orchestration.

kshyatt:
This seems to have resolved itself for now?

luraess:
After some rework of the testing and runner setup we may have found a more stable config.

PR #433, opened 2026-05-18
Initial steps to disentangle tensor tests

kshyatt:
The idea here is to use something like what we have in MAK to reduce the duplication across test files... if people like this approach I will continue on to `indexmanipulation` tests and others


kshyatt:
This should also hopefully help with https://github.com/QuantumKitHub/TensorKit.jl/issues/428

lkdvos (commented):
Definitely in favor of having a more modular and structured test set, but I was wondering if for the GPU tests we really want to repeat the tests in the same way we do them on CPU, and not just check that they give the same results as on CPU.

For example, a pattern I quite like is something like this:
https://github.com/QuantumKitHub/Strided.jl/blob/2a2b380e3718b2875c5125672fc9e58140c68e3b/test/gpu.jl#L1-L12

To make this more concrete, while I know that for CPU we test that `norm` is well-defined, i.e. `norm(2t) = 2norm(t)`, ... I would probably just check in the GPU case that `norm(t) = norm(adapt(CuVector, t))` and be done with it?

kshyatt:
>  I was wondering if for the GPU tests we really want to repeat the tests in the same way we do them on CPU, and not just check that they give the same results as on CPU.

It's a fair question. For now, I'm keeping them mirrored because we ran into weirdness where `norm` specifically of an `AdjointTensorMap{<:CuTensorMap}` was doing something dumb... there were enough edge cases that I was biased towards keeping the bigger suite of tests. But maybe for CI it would make sense to have a slimmer version which just directly compares with the CPU in a more restricted set of cases?

test/testsuite/tensors/construction.jl
```diff
@@ -0,0 +1,77 @@
...
+        @test scalartype(t) == T
+        @test norm(t) == 0
+        @test codomain(t) == W
+        @test space(t) == (W ← one(W))
+        @test domain(t) == one(W)
+        @test typeof(t) == TensorMap{T, spacetype(t), 5, 0, AT}
+    end
+end
+
+function basic_blocks_properties(t, W)
```
Jutho, line 13: Is it intentional that this one is not following the format

```julia
return @testset "Basic `blocks` properties" begin

end

kshyatt: Ah, no 😓

lkdvos:
> where `norm` specifically of an `AdjointTensorMap{<:CuTensorMap}` was doing something dumb...

I don't think this is really an argument in favor or against the different setups, you could also just do `compare(t -> norm(t'), ...)`.
I'm not saying to cut down on the things we test, I'm mostly advocating for not testing them by comparison to analytical results and just comparing with CPU results where possible.

PR #434, opened 2026-05-19
Mooncake forward rules for linalg

kshyatt:
Also added tests and a primitive marker for the projection methods.

### DIFF
--- a/ext/TensorKitMooncakeExt/linalg.jl
+++ b/ext/TensorKitMooncakeExt/linalg.jl
@@ -3,7 +3,7 @@
 pullback_dC!(ΔC, β) = (scale!(ΔC, conj(β)); return NoRData())
 pullback_dβ(ΔC, C, β) = _needs_tangent(β) ? project_scalar(β, inner(C, ΔC)) : NoRData()
 
-@is_primitive DefaultCtx ReverseMode Tuple{typeof(mul!), AbstractTensorMap, AbstractTensorMap, AbstractTensorMap, Number, Number}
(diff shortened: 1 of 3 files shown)


ext/TensorKitMooncakeExt/linalg.jl
```diff
@@ -86,13 +117,21 @@ function Mooncake.rrule!!(::CoDual{typeof(inv)}, A_ΔA::CoDual{<:AbstractTensorM
 
     return Ainv_ΔAinv, inv_pullback
 end
+function Mooncake.frule!!(::Dual{typeof(inv)}, A_ΔA::Dual{<:AbstractTensorMap})
+    A, ΔA = arrayify(A_ΔA)
+    Ainv = inv(A)
+    ΔAinv = -Ainv * ΔA * Ainv
```
lkdvos:
```suggestion
    ΔAinv = scale!(Ainv * ΔA * Ainv, -1)
```

I was looking at this and hoping to avoid another intermediate allocation, but this might be the best I can come up with

lkdvos (approved):
Overall looks good to me! Left one small comment but otherwise I'd say good to go

ext/TensorKitMooncakeExt/linalg.jl
kshyatt: Thanks! I was definitely not as focused on performance when setting this up so optimizations are most welcome

lkdvos (approved):

kshyatt:
All these failures look unrelated (AMDGPU and macos runners being sad) so I'm going to merge

PR #436, opened 2026-05-19
Forward AD tests and cleanup for vector interface

kshyatt:


lkdvos (commented):
Cool! Do you have any idea if this would make sense to implement at the VectorInterface level? Alternatively, to not necessarily mark everything as `@is_primitive` there but to already provide the implementation?

kshyatt:
We could do it there if it's generic enough? I don't have a strong opinion tbh

lkdvos:
It looks like all rules are implemented solely in terms of Mooncake or VectorInterface functions, so that might just work?

kshyatt:
Nice, I can move these over tomorrow then

lkdvos:
I think you still need the `@is_primitive` calls, since the VI rules are only defined for `AbstractArray`

kshyatt:
Addressed by another PR over there 😉

ext/TensorKitMooncakeExt/vectorinterface.jl
lkdvos, line 1: Do you know if something like this works:
```julia
@static if pkgversion(VectorInterface) < v"0.6.0"
    # previous implementation
end
```

This would allow us to continue supporting old + new version of VectorInterface.

For the current implementation, I guess this might break the `rrule` implementations that were already there if the correct version of VectorInterface is not loaded. In that case, I would maybe just bite the bullet and actually restrict to VectorInterface v0.6 and just deal with that?

kshyatt: Not sure, I can try it tomorrow!

test/mooncake/vectorinterface.jl
```diff
@@ -17,20 +15,19 @@ eltypes = (Float64, ComplexF64)
...
+        Mooncake.TestUtils.test_rule(rng, scale!, C, α; atol, rtol)
+        Mooncake.TestUtils.test_rule(rng, scale!, C', α; atol, rtol)
+        Mooncake.TestUtils.test_rule(rng, scale!, C, A, α; atol, rtol)
+        Mooncake.TestUtils.test_rule(rng, scale!, C', A', α; atol, rtol)
+        Mooncake.TestUtils.test_rule(rng, scale!, copy(C'), A', α; atol, rtol)
+        Mooncake.TestUtils.test_rule(rng, scale!, C', copy(A'), α; atol, rtol)
+
+        Mooncake.TestUtils.test_rule(rng, add!, C, A; atol, rtol, is_primitive = false)
+        Mooncake.TestUtils.test_rule(rng, add!, C, A, α; atol, rtol, is_primitive = false)
+        Mooncake.TestUtils.test_rule(rng, add!, C, A, α, β; atol, rtol)
```
Jutho, line 28: no `AdjointTensorMap` instances for `add!` ?

kshyatt: That was already the case previously but I'll add some

ext/TensorKitMooncakeExt/vectorinterface.jl
kshyatt: Having to support the old versions again leads to having duplicated rules all over the place, though, one advantage of doing it the current way is being able to cut quite a few lines from the extension.

lkdvos (approved):
Alright, I think I've been annoying for long enough about the VectorInterface compat, so I think the main thing here is to just bump the VI version and drop the versions that don't have these rules, and go from here. Otherwise this PR looks ready for merging to me, thanks @kshyatt !

lkdvos (approved):

Jutho (approved):

kshyatt:
OK the Enzyme fail is unrelated, something that intermittently didn't show up previously, I'll fix separately in the Enzyme VI PR

PR #437, opened 2026-05-19
Forward rules for TensorOperations calls

kshyatt:


ext/TensorKitMooncakeExt/tensoroperations.jl
```diff
@@ -70,6 +69,35 @@ function Mooncake.rrule!!(
...
+    (C, ΔC), (A, ΔA), (B, ΔB) = arrayify.((C_ΔC, A_ΔA, B_ΔB))
+    pA, pB, pAB = primal.((pA_ΔpA, pB_ΔpB, pAB_ΔpAB))
+    α, Δα = Mooncake.extract(α_Δα)
+    β, Δβ = Mooncake.extract(β_Δβ)
+    backend, allocator = primal.((backend_Δbackend, allocator_Δallocator))
+    # ΔC′ = ΔC*β + C*Δβ + A*B*Δα + ΔA*B*α + A*ΔB*α
+    scale!(ΔC, β)
+    if !isa(Δβ, Mooncake.NoTangent)
+        add!(ΔC, C, Δβ)
+    end
```
lkdvos:
```suggestion
    if isa(Δβ, Mooncake.NoTangent)
        scale!(ΔC, β)
    else
        add!(ΔC, C, Δβ, β)
    end
```

lkdvos: Same pattern holds below :)

kshyatt: Applied there too

ext/TensorKitMooncakeExt/tensoroperations.jl
```diff
@@ -70,6 +69,36 @@ function Mooncake.rrule!!(
     return C_ΔC, blas_contract_pullback
 end
 
+function Mooncake.frule!!(
```
lkdvos, line 72: I am less familiar with Mooncake's `frule!!` support, but I have one comment about this implementation that mostly has to do with efficiency:
By intercepting the rule at this level, we will have to permute the input arrays multiple times, once for each `blas_contract` call. It seems like it would be better to try and overload one level down, since we could just get the `tensoradd` and `mul` calls separately?

I know this is also not entirely implemented for the `rrule`, and there are definitely efficiency questions for that too.

kshyatt: Yeah it's a fair point. I did this here mostly to match the `rrule`, but maybe we should instead add a rule for `tensoradd` since we anyway have `mul!`. We could also merge this for now but open an issue to look at handling the lower level stuff for both fwd and rvs modes?

ext/TensorKitMooncakeExt/tensoroperations.jl
```diff
@@ -70,6 +69,36 @@ function Mooncake.rrule!!(
...
+        A_ΔA::Dual{<:AbstractTensorMap}, pA_ΔpA::Dual{<:Index2Tuple},
+        B_ΔB::Dual{<:AbstractTensorMap}, pB_ΔpB::Dual{<:Index2Tuple},
+        pAB_ΔpAB::Dual{<:Index2Tuple},
+        α_Δα::Dual{<:Number}, β_Δβ::Dual{<:Number},
+        backend_Δbackend::Dual, allocator_Δallocator::Dual
+    )
+    # prepare arguments
+    (C, ΔC), (A, ΔA), (B, ΔB) = arrayify.((C_ΔC, A_ΔA, B_ΔB))
+    pA, pB, pAB = primal.((pA_ΔpA, pB_ΔpB, pAB_ΔpAB))
+    α, Δα = Mooncake.extract(α_Δα)
```
Jutho: Is `Mooncake.extract` a new thing? This doesn't seem so familiar as `primal` and `arrayify`. It seems like there are too many different methods to destructure these `Dual` variables into their primal and (co)tangent part.

Jutho: Maybe we want to add `extract` to the list of names we are `using` from Mooncake, as otherwise it seems a bit inhomogeneous to only qualify this one.

ext/TensorKitMooncakeExt/tensoroperations.jl
```diff
@@ -70,6 +69,36 @@ function Mooncake.rrule!!(
...
+    β, Δβ = Mooncake.extract(β_Δβ)
+    backend, allocator = primal.((backend_Δbackend, allocator_Δallocator))
+    # ΔC′ = ΔC*β + C*Δβ + A*B*Δα + ΔA*B*α + A*ΔB*α
+    if isa(Δβ, Mooncake.NoTangent)
```
Jutho: While looking at the list of `using Mooncake: `, `NoTangent` is also in there so it doesn't need to be qualified in principle.

Jutho (approved):
Two stylistic comments, but otherwise looks good to me.

Jutho (approved):
Up to minor stylistic comments, looks good to me.

ext/TensorKitMooncakeExt/tensoroperations.jl
Jutho: What would be the multiple permutation of `A` appearing here? I thought I understood this remark in the context of the pullback, but here I don't really see how this would arise.

lkdvos: - on line 94, 97 and 98 we have `blas_contract!(..., A, pA, ...)`, which will each effectively call `permute(A, pA)`.
- on line 94, 96 and 98 the same argument is true for `blas_contract!(..., B, pB, ...)`

Also, the `AB` that is first added to `C`, and then to `DeltaC` could in principle be reused, similar to the `rrule` case. (line 94 and line 98)

The problem is that if we simply don't overload this, this will automatically be resolved. Schematically (ignoring alpha and beta for simplicity):

```julia
Ap = permute(A, pA)
dA = permute(dA, pA) # frule for permute

Bp = permute(B, pB)
dB = permute(dB, pB) # frule for permute

AB = Ap * Bp
dAB = Ap * dB + dA * Ap # frule for *

C = permute(AB, pAB)
dC = permute(dAB, pAB) # frule for permute
```

Jutho: Well, certainly, I would have hoped that with Moonzyme, we needed to write much fewer rules, so if we can do without certain rules, I think that would be great, also from a maintainer's perspective.

kshyatt: I was sort of hoping to do all this in a "tick-tock" fashion -- first, write rules relatively 1:1 with what already exists for ChainRules, so we can verify correctness and have "something to go off of". Then find lower level commonalities, write rules for those, and lean more on the fancy compilers to support us. All this is to say I'll open an issue and link to these comments.

PR #439, opened 2026-05-21
Fix `isapprox` for TensorMaps with different spaces

Yue-Zhengyuan:
This PR improves `isapprox` for TensorMaps: when two TensorMaps have different spaces, `isapprox` should directly return `false`. Previously such a check is missing and a `SpaceMismatch` error will be raised.

### DIFF
--- a/src/tensors/abstracttensor.jl
+++ b/src/tensors/abstracttensor.jl
@@ -683,6 +683,7 @@ function Base.isapprox(
         t1::AbstractTensorMap, t2::AbstractTensorMap;
         atol::Real = 0, rtol::Real = Base.rtoldefault(scalartype(t1), scalartype(t2), atol)
     )
+    (codomain(t1) == codomain(t2) && domain(t1) == domain(t2)) || return false
     d = norm(t1 - t2)
     if isfinite(d)
         return d <= max(atol, rtol * max(norm(t1), norm(t2)))


Jutho:
I think the current behavior is in line with how `isapprox` works on matrices:

```julia-repl
julia> isapprox(rand(3,4), rand(5,6))
ERROR: DimensionMismatch: a has size (3, 4), b has size (5, 6), mismatch at dim 1
Stacktrace:
 [1] throw_promote_shape_mismatch(a::Tuple{Base.OneTo{…}, Base.OneTo{…}}, b::Tuple{Base.OneTo{…}, Base.OneTo{…}}, i::Int64)
   @ Base ./indices.jl:135
 [2] promote_shape
   @ ./indices.jl:196 [inlined]
 [3] promote_shape
   @ ./indices.jl:188 [inlined]
 [4] -(A::Matrix{Float64}, B::Matrix{Float64})
   @ Base ./arraymath.jl:7
 [5] isapprox(x::Matrix{…}, y::Matrix{…}; atol::Int64, rtol::Float64, nans::Bool, norm::typeof(LinearAlgebra.norm))
   @ LinearAlgebra ~/.julia/juliaup/julia-1.12.5+0.aarch64.apple.darwin14/Julia-1.12.app/Contents/Resources/julia/share/julia/stdlib/v1.12/LinearAlgebra/src/generic.jl:1985
 [6] isapprox(x::Matrix{Float64}, y::Matrix{Float64})
   @ LinearAlgebra ~/.julia/juliaup/julia-1.12.5+0.aarch64.apple.darwin14/Julia-1.12.app/Contents/Resources/julia/share/julia/stdlib/v1.12/LinearAlgebra/src/generic.jl:1981
 [7] top-level scope
   @ REPL[1]:1

```

Yue-Zhengyuan:
Let me ask if LinearAlgebra.jl does this intentionally: https://github.com/JuliaLang/LinearAlgebra.jl/issues/1624

lkdvos:
As I mentioned in that issue there, I am not a huge fan of this, and would prefer if this just throws an error. My main objection is that the definition of `isapprox(x, y)` implies `norm(x, y) > tol` if `!isapprox` is `true`, which would no longer be the case here.

PR #440, opened 2026-05-22
Add Enzyme forward and reverse rules and tests for VectorInterface

kshyatt:
Building on top of the rules at VI in order to start breaking up mega PR https://github.com/QuantumKitHub/TensorKit.jl/pull/380

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,9 @@ uuid = "07d1fe3e-3e46-537d-9eac-e9e13d0d4cec"
 version = "0.17.0"
 authors = ["Jutho Haegeman, Lukas Devos"]
 
+[workspace]
+projects = ["test", "docs"]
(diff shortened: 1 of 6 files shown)


lkdvos:
Here I think we should wait a bit, I would prefer not to immediately drop support for the older VectorInterface versions before the v0.17 release if you don't mind?

kshyatt:
Yeah for sure!

kshyatt:
This anyway needs something merged at Enzyme to work

kshyatt:
Once https://github.com/QuantumKitHub/TensorKit.jl/pull/444 is in I think this is ready, if others agree

kshyatt:
Or not... this was working locally 😭

lkdvos (approved):

PR #441, opened 2026-05-29
Correct artin braid image

borisdevos:
Typo in the artin braid image in the docs is corrected here. Credits to @Jutho who made the image.

Background is no longer alpha, not sure if that's an issue. Let's see how the preview looks like.


lkdvos:
The background is kind of important for the dark themes though, is there any chance you can simply edit the original svg?

lkdvos:
I do think the image still got stretched somehow, it seems like the width/height got severely altered? I'm not sure if this really matters, but it does look slightly stretched on my phone

Jutho:
I made a new attempt; still not identical but almost the same now.

borisdevos (commented):
Can't approve, but 4 pixels taller shouldn't be an issue ;)

Jutho (approved):

Jutho:
Good, I'll approve my own 😄

PR #442, opened 2026-06-01
Fix foldright cache miss dual flag

yitan1:
This fixes the cache miss path in `foldright(::FusionTreeBlock)` by passing the same `!isduala` argument to `multi_Fmove_inv` as the cache initialization path.

Without this argument, the miss path can construct intermediate trees with `isdual` metadata inconsistent with the destination block. The current matrix result is typically unaffected because `treeindex_data` ignores block-level `isdual`, but the cached tree metadata is inconsistent and differs from the `foldleft` implementation.

Local checks:
- `git diff --check`
- focused Julia check covering `foldright`/`foldleft` round trip for SU2Irrep, FibonacciAnyon, and IsingAnyon


Jutho:
Thanks. Nice find! This was a subtle bug indeed.

Jutho (approved):

PR #443, opened 2026-06-01
Try to use vector multiplication for truncation error computation

kshyatt:
I'll add benchmarks to this after I have dinner. I think this should work ok (at least for trivial `SectorVector`s?) because the boolean inds are just `0` or `1` and we can multiply by their `!` to set stuff to zero. That kernel seems to be a bit faster.

### DIFF
--- a/ext/TensorKitGPUArraysExt.jl
+++ b/ext/TensorKitGPUArraysExt.jl
@@ -96,6 +96,14 @@ function MatrixAlgebraKit.findtruncated_svd(values::GPUSectorVector, strategy::M
     return SectorDict(c => Adapt.adapt(Vector, MatrixAlgebraKit.findtruncated_svd(d, strategy′)) for (c, d) in pairs(values))
 end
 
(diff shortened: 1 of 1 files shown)


lkdvos (commented):
I guess this only works for logical indices though, so probably have to specialize on that? Additionally, if we write custom code for that, you might be able to combine the norm and truncation pass with broadcasting (you can take a norm of a broadcast expression), since the norm also needs to spawn multiple kernels to include the quantum dimensions?

kshyatt:
> you can take a norm of a broadcast expression

For GPU arrays? I'm not sure about this tbh, it wasn't working with `norm(x, Inf)` but perhaps for 2-norm it's ok, I'll try

kshyatt:
Also I don't think we can bcast over `values` itself because it's a subtype of `AbstractDict`

lkdvos (approved):
Looks good to me! Do you know if this is covered in the tests? If it is, good to go for me!

kshyatt:
It seems to be based on codecov

PR #446, opened 2026-06-05
Turn off ReadOnlyMemoryError triggering AMD tests

kshyatt:


test/amd/tensors.jl
```diff
@@ -97,7 +97,7 @@ for V in spacelist
             for T in (Int, Float32, ComplexF64)
                 t = @constinferred AMDGPU.rand(T, W)
                 d = convert(Dict, t)
-                @test adapt(Array, t) == convert(TensorMap, d)
+                @test convert(Dict, adapt(Array, t)) == d
```
Jutho, line 100: This is a subtle change, as it stops testing the restoration of the tensor. But since that is already tested at different places, I think this change is ok.

kshyatt: It's also now mirroring what the CUDA tests are doing

lkdvos (approved):

PR #448, opened 2026-06-10
A couple missing zero-derivs and an inplace rrule test for svd_trunc

kshyatt:


ext/TensorKitMooncakeExt/factorizations.jl
lkdvos, line 1: I forgot we have the `svd_trunc` rules here, and don't actually remember why we needed that. (Did I do this? Does my memory still work? Are we even real?)
 Is this not the same as the MatrixAlgebraKit implementation?

kshyatt: You added them lmao

kshyatt: But yeah I think it should be, let me try removing this and just falling into MAKtopia

test/mooncake/factorizations.jl
```diff
@@ -88,7 +88,7 @@ eltypes = (Float64, ComplexF64)
 
     @timedtestset "Singular value decomposition" begin
         for t in (randn(T, V[1] ← V[1]), randn(T, V[1] ⊗ V[2] ← (V[3] ⊗ V[4] ⊗ V[5])'))
-            USVᴴ = svd_compact(t)
+            #=USVᴴ = svd_compact(t)
```
lkdvos: Do we know why the non-truncated version is not working?

kshyatt: Oh whoops I think I turned this off to fix other tests, I can restore it.

lkdvos: It looks like this was disabled before as well, I just can't remember anything apparently :p

kshyatt:
OK well apparently this just works without the rules in TensorKit, pretty darn cool in my view.

lkdvos (approved):

PR #449, opened 2026-06-10
Forward and reverse Enzyme tests and rules for linalg

kshyatt:
Trying to make these a little more manageable and pick up the fwd rules where possible


ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,262 @@
...
+        ::Type{RT},
+        C::Annotation{<:AbstractTensorMap},
+        A::Annotation{<:AbstractTensorMap},
+        B::Annotation{<:AbstractTensorMap},
+        α::Annotation,
+        β::Annotation,
+    ) where {RT}
+    cacheC = !isa(β, Const) && copy(C.val)
+    cacheA = !isa(B, Const) && EnzymeRules.overwritten(config)[3] ? copy(A.val) : nothing
+    cacheB = !isa(A, Const) && EnzymeRules.overwritten(config)[4] ? copy(B.val) : nothing
```
Jutho, line 16: Since my Enzyme knowledge has fallen back to zero, I was comparing this with some `mul!` rule in Enzyme.jl ( https://github.com/EnzymeAD/Enzyme.jl/blob/6d9c0cb7fa1ab4a4ce347ba506ea9715761365a8/src/internal_rules/linalg.jl#L304 ), and while there are clearly some similarities, there are also some differences, e.g. in which `overwritten(config)` positions are checked to decide on `cacheA` and `cacheB` (they use 5 and 6, compared to 3 and 4 here). Is there a good explanation for that?

kshyatt: It's probably that they are correct and I made a mistake. This part is very difficult to test because it only arises in a longer set of operations. I wish they had written their `mul!` rule to be a little more generic

kshyatt: https://enzymead.github.io/Enzyme.jl/dev/generated/custom_rule/#Defining-a-reverse-mode-rule what's confusing is it doesn't match what is shown here. I will ask the devs.

kshyatt: https://github.com/EnzymeAD/Enzyme.jl/blob/6d9c0cb7fa1ab4a4ce347ba506ea9715761365a8/src/internal_rules/linalg.jl#L35 Seems to follow what I have here.

Jutho: I also think your counting is consistent with the doc string of `overwritten`.

kshyatt: I posted about it on Slack to hopefully get this figured out but now I'm so confused about who is right 😹

kshyatt: > Since my Enzyme knowledge has fallen back to zero,

Never give up hope, mental healing *is* possible

kshyatt: OK, I checked with Billy on slack and made a PR to correct the method you linked (https://github.com/EnzymeAD/Enzyme.jl/pull/3176)

kshyatt:
The test on 1.12 is passing locally for me! I assume it's getting OOMed or something...

kshyatt:
OK, everything looks happy now except the GPU stuff which is unrelated. Are we good to go?

ext/TensorKitEnzymeExt/utility.jl
```diff
@@ -0,0 +1,80 @@
...
+    project_scalar(x::Number, dx::Number)
+
+Project a computed tangent `dx` onto the correct tangent type for `x`.
+For example, we might compute a complex `dx` but only require the real part.
+"""
+project_scalar(x::Number, dx::Number) = oftype(x, dx)
+project_scalar(x::Real, dx::Complex) = project_scalar(x, real(dx))
+
+# in-place multiplication and accumulation which might project to (real)
+# TODO: this could probably be done without allocating
```
Jutho: Probably not, because BLAS doesn't support matrix multiplication with matrices that do not have stride one, which is what you will get if you try to isolate the real and imaginary part of a complex matrix, in order to do something like
`C = Ar * Br - Ai * Bi`.

We might be able to make the final `real` as a view instead of an extra allocation.

ext/TensorKitEnzymeExt/utility.jl
```diff
@@ -0,0 +1,80 @@
...
+end
+Base.@constprop :aggressive function _repartition(p::Index2Tuple, N₁::Int)
+    return _repartition(linearize(p), N₁)
+end
+function _repartition(p::Union{IndexTuple, Index2Tuple}, ::Index2Tuple{N₁}) where {N₁}
+    return _repartition(p, N₁)
+end
+function _repartition(p::Union{IndexTuple, Index2Tuple}, t::AbstractTensorMap)
+    return _repartition(p, TensorKit.numout(t))
+end
```
Jutho: It feels like much of this must already exist in the main TensorKit module, or in TensorOperations. If not, it is possibly useful to add it in the main module. (`TensorOperations.jl/src/indices.jl` has some stuff).

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,262 @@
...
+        α::Annotation{<:Number},
+        β::Annotation{<:Number},
+    ) where {RT}
+    # ΔC′ = ΔC*β + C*Δβ + A*B*Δα + ΔA*B*α + A*ΔB*α
+    if !isa(C, Const)
+        scale!(C.dval, β.val)
+        !isa(β, Const) && add!(C.dval, C.val, β.dval)
+        !isa(α, Const) && project_mul!(C.dval, A.val, B.val, α.dval)
+        !isa(A, Const) && project_mul!(C.dval, A.dval, B.val, α.val)
+        !isa(B, Const) && project_mul!(C.dval, A.val, B.dval, α.val)
```
Jutho: These are probably all regular `mul!` calls with `beta = 1`. Maybe it is a bit confusing that `project_mul!` does admit a beta and implicitly has beta=1 (I understand this is all you need for the tangents, but still), whereas `mul!` has default `beta=0`.

ext/TensorKitEnzymeExt/utility.jl
kshyatt: That one is a comment of @lkdvos (I ported these over from the Mooncake extension) so I'll defer to him here 😉

ext/TensorKitEnzymeExt/utility.jl
kshyatt: Yeah, especially since this is now duplicated across the two AD extensions. I'll see if I can't move it into the main TK module.

ext/TensorKitEnzymeExt/utility.jl
lkdvos: I thought there was a trick to make use of BLAS anyways, especially if one of the arrays is real, but I am now no longer really convinced... In any case, this is probably trying to over-optimize anyways, so I'm definitely not saying this is something to tackle here...

(the trick I was thinking of probably is this:)

If you have something like `A' * B`, you can turn this into a `(m x 2n) * (2n x k)` multiplication simply by reinterpreting as real arrays, in which case this will compute exactly `transpose(Ar) * Br - transpose(Ai) * Bi`, without allocating and with BLAS. The problem being that this is not super useful here, since we don't actually get to choose the layout of the matrices 😁

ext/TensorKitEnzymeExt/utility.jl
lkdvos: I actually think that these are now in the later versions of TensorOperations, so it's probably better to use that? E.g.: https://github.com/QuantumKitHub/TensorOperations.jl/blob/1cf29e29e43efac41f713292d99a9d58b7078c57/src/indices.jl#L35-L44

ext/TensorKitEnzymeExt/linalg.jl
lkdvos: I agree that it might be nicer to simply have the default `beta = 0` and be explicit about `beta = 1` here, also since it makes it more clear that this is accumulating when reading

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,262 @@
...
+    ) where {RT}
+    # ΔC′ = ΔC*β + C*Δβ + A*B*Δα + ΔA*B*α + A*ΔB*α
+    if !isa(C, Const)
+        scale!(C.dval, β.val)
+        !isa(β, Const) && add!(C.dval, C.val, β.dval)
+        !isa(α, Const) && TK.project_mul!(C.dval, A.val, B.val, α.dval, One())
+        !isa(A, Const) && TK.project_mul!(C.dval, A.dval, B.val, α.val, One())
+        !isa(B, Const) && TK.project_mul!(C.dval, A.val, B.dval, α.val, One())
+    end
+    mul!(C.val, A.val, B.val, α.val, β.val)
```
Jutho: I guess another case could be made here for computing `AB = A.val * B.val` separately, in the case where `!isa(α, Const)`. Also, I don't think we need `project_mul!` here, since `C` should be complex as soon as any of `A`, `B`, `α` or `β` is complex.

Maybe something like this:


```suggestion
    if !isa(C, Const)
        scale!(C.dval, β.val)
        !isa(β, Const) && add!(C.dval, C.val, β.dval)
        !isa(A, Const) && mul!(C.dval, A.dval, B.val, α.val, One())
        !isa(B, Const) && mul!(C.dval, A.val, B.dval, α.val, One())
    end
    if !isa(α, Const) && !isa(C, Const)
        if iszero(β.val) && !iszero(α.val)
            # this is probably quite a common case, so maybe worth specializing
            mul!(C.val, A.val, B.val, α.val, β.val)
            add!(C.dval, C.val, α.dval / α.val)
        else
            AB = A.val * B.val
            add!(C.val, AB, α.val, β.val)
            add!(C.dval, AB, α.dval)
        end
    else
        mul!(C.val, A.val, B.val, α.val, β.val)
    end
```

Jutho: Something went wrong, this is a suggestion for the whole function body above.

kshyatt: > Also, I don't think we need project_mul! here, since C should be complex as soon as any of A, B, α or β is complex.

Can't you pass in real-valued `C` and complex valued other arguments if you are using the direct in-place method?

kshyatt: You're only guaranteed `C` has complex elements if you're using `mul` (no bang)

There's no forcing of all complex [here](https://github.com/QuantumKitHub/TensorKit.jl/blob/20ec62847376cce9cbfb92ec9f7c0ae839dbe50f/src/tensors/diagonal.jl#L333) or [here](https://github.com/QuantumKitHub/TensorKit.jl/blob/20ec62847376cce9cbfb92ec9f7c0ae839dbe50f/src/tensors/linalg.jl#L237)?

kshyatt: Tried the suggestion locally and it worked, thanks!

lkdvos: There's no forcing, but the function will throw `InexactError` because `setindex!` on `C` will have to convert for any non-real value.
I guess you could in principle come up with an evil case where the eltype of one of the inputs is complex but the values are purely real, which then might work and give issues here?

kshyatt: Yeah exactly, suppose I generate `C`, `A`, `B` all real then get `Ac = complex(A)` which is of course all "real" valued. But this doesn't need to be solved in the forward rules, this PR has been expanded enough

lkdvos:
Do we think the test failure is a problem with how LRU interacts with Enzyme?

From the stacktrace, I seem to read this as not finding a key, even though being in an if-clause that explicitly checks this: https://github.com/JuliaCollections/LRUCache.jl/blob/1dad9fef75fef51ea1b7e984e5850ad4e374a7e0/src/LRUCache.jl#L172-L175

The really confusing part to me is that it seems to originate from a forward call, which should just be a regular function call, so I'm not sure what is really going on there. I also don't think this can really be a race condition, since 1) I don't think we are multithreading, 2) LRU protects against this?

kshyatt:
It also seems to only happen in the CompatCheck tests, not the main ones

kshyatt:
Let me just see if bumping the LRUCache compat helps at all...

kshyatt:
OK that makes CompatCheck pass and the min test fail. It seems the failures only happen on 1.10 regardless, but they are intermittent. Really annoying.

kshyatt:
Also locally I can see it happen in reverse calls so I think it's not to do with fwd mode really

kshyatt:
Removing the `@cached` in front of the definition of `degeneracystructure` does seem to fix this. Maybe there's a way to fill the cache before the Enzyme tests run? I'll take a look.

kshyatt:
OK so after more digging, it looks like the problem here is 1.10 + Enzyme + the `@cached` macro. I'm ok with disabling this set of tests on 1.10 for now while I try to work with the Enzyme people to figure out what's going on. Does that sound reasonable?

ext/TensorKitEnzymeExt/linalg.jl
Jutho: A) That is evil
B) I think this is just plain wrong. The situation is really different from the reverse rule, where real A times complex B makes sense, and the pullback with respect to A needs to be projected back. If you do something like real A times complex B which is actually real, but has some complex derivative, then maybe `mul!(C, A, B)` works with a real C, but it doesn't make sense to say that `dC` is the projected version of `dA * B + A * dB`. If `dB` is complex, so does `dC` need to be.

Jutho: So I'd rather have an error than a silent incorrect result.

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,273 @@
...
+        config::EnzymeRules.RevConfigWidth{1},
+        func::Const{typeof(norm)},
+        dret::Active,
+        cache,
+        A::Annotation{<:AbstractTensorMap},
+        p::Const{<:Real},
+    )
+    n, cacheA = cache
+    Δn = dret.val
+    p.val == 2 || error("currently only implemented for p = 2")
```
Jutho, line 187: Do we need the error twice? It also appears in `augmented_primal`. Can we only have it here, when `reverse` is actually called?

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,273 @@
...
+    Aval = something(cacheA, A.val)
+    if !isa(A, Const)
+        x = (Δn' + Δn) / 2 / hypot(n, eps(one(n)))
```
Jutho: Is this `hypothetical(n, eps(one(n)))` some regularization for the case where `n = norm(A.val)` is very small?

Jutho: Also, is `(Δn' + Δn) / 2` just `real(Δn)`. And is this ever not real, as we anyway have `n::Real`?

kshyatt: I believe so but we can also ask its [original progenitor](https://github.com/QuantumKitHub/TensorKit.jl/tree/main/ext/TensorKitChainRulesCoreExt#L114) @lkdvos

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,273 @@
...
+end
+function EnzymeRules.forward(
+        config::EnzymeRules.FwdConfigWidth{1},
+        func::Const{typeof(norm)},
+        ::Type{RT},
+        A::Annotation{<:AbstractTensorMap},
+        p::Const{<:Real},
+    ) where {RT}
+    y = norm(A.val, p.val)
+    Δy = if EnzymeRules.needs_shadow(config) && !isa(A, Const)
```
Jutho: Does this need a `p==2` check?

ext/TensorKitEnzymeExt/linalg.jl
kshyatt: In other AD systems (where Δn might be some weird Dual number type) I don't know if real is implemented? But here it should be fine.

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,273 @@
...
+function EnzymeRules.augmented_primal(
+        config::EnzymeRules.RevConfigWidth{1},
+        func::Const{typeof(inv)},
+        ::Type{RT},
+        A::Annotation{<:AbstractTensorMap},
+    ) where {RT}
+    ret = inv(A.val)
+    primal = EnzymeRules.needs_primal(config) ? ret : nothing
+    shadow = EnzymeRules.needs_shadow(config) ? make_zero(ret) : nothing
+    cache = (ret, shadow)
```
Jutho: I am still very confused by how Enzyme works. If I understand correctly, then if any of the input arguments are needed for the `reverse` rule, there is `EnzymeRules.overwritten(config)` that records whether these variables may be overwritten after calling `augmented_primal` and before calling `reverse`, so you know you have to copy them and put them in the cache. But if you need the return value in order to implement the `reverse` rule, then you can just put it in the cache without copy. What happens is `ret` is changed before the `augmented_primal` call?

kshyatt: You mean if `ret` is changed after?

kshyatt: I'm also extremely confused by this but [this line](https://github.com/EnzymeAD/Enzyme.jl/blob/main/src/internal_rules/linalg.jl#L60) implies we might need to copy `res`... but the previous issue we had with the SparseArrays `mul!` makes me unsure and the docs don't seem to address this.

Jutho: Yes sorry, after the `augmented_primal` but before the `reverse` call. I have already forgotten about `mul!` for SparseArrays. Was I involved in that? 😄

kshyatt: You pointed out that the overwritten indexing wasn't consistent and we correct a bug at Enzyme as a result! I think, having brooded on this, we should indeed copy the result if we'll need it in the reverse pass. Until now we haven't encountered this situation much...

ext/TensorKitMooncakeExt/indexmanipulations.jl
```diff
@@ -57,7 +57,7 @@ for transform in (:permute, :transpose)
             end
             ΔAr = NoRData()
 
-            Δαr = isnothing(Ap) ? NoRData() : project_scalar(α, inner(Ap, ΔC))
+            Δαr = isnothing(Ap) ? NoRData() : TO.project_scalar(α, inner(Ap, ΔC))
```
Jutho: I lost track of where our new `project_` methods live. Is TO a logical place for `project_scalar`? We probably also have it in VI? Or is TO just importing the VI version? Where is `project_mul!`, is that in TO? Or TK?

kshyatt: Lukas suggested pointing to TO. We have duplicated versions of these in both Mooncake and Enzyme extensions over at VI 🫠

kshyatt: `project_mul` lives here at TK

kshyatt: @Jutho did you ever read those books "Where's Waldo?"? We need that for ad helpers

lkdvos: I think I originally just wrote these as small helper functions, but then it turned out we kind of need them everywhere, and then they just kind of proliferated

Jutho: Can you *read* "Where's Waldo"? Maybe we can just create one super package VIMAKTOTK.jl, and this solves all of our problems. And then we can keep up with the QMC guys in terms of acronyms 😄 .

kshyatt: All depends on your definition of "reading" 😉

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,274 @@
...
+        cache,
+        A::Annotation{<:AbstractTensorMap},
+        p::Const{<:Real},
+    )
+    n, cacheA = cache
+    Δn = dret.val
+    p.val == 2 || error("currently only implemented for p = 2")
+    Aval = something(cacheA, A.val)
+    if !isa(A, Const)
+        x = real(Δn) / hypot(n, eps(one(n)))
```
Jutho: In analogy to the forward rule, I would use `pinv(n)` here.


```suggestion
        x = real(Δn) * pinv(n)
```

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,274 @@
...
+end
+function EnzymeRules.forward(
+        config::EnzymeRules.FwdConfigWidth{1},
+        func::Const{typeof(norm)},
+        ::Type{RT},
+        A::Annotation{<:AbstractTensorMap},
+        p::Const{<:Real},
+    ) where {RT}
+    p.val == 2 || error("currently only implemented for p = 2")
+    y = norm(A.val, p.val)
```
Jutho: In analogy with the reverse rule, I would use `n` and `Δn` here instead of `y` and `Δy`.

kshyatt:
Some more strangeness: the missing key error only ever occurs with `Vtr`, I cannot trigger it with `VRepU₁` or any of the other AD spaces. This makes me wonder if the `@cached` stuff is kind of a red herring -- I'll dig into this more today.

ext/TensorKitMooncakeExt/indexmanipulations.jl
kshyatt: This line got moved into TK proper as `_pullback_dα`, which the Mooncake and Enzyme extensions call as appropriate once they check in their own personalized way whether the tangent of `α` is needed

kshyatt:
We're still getting this strange cache miss for reasons I don't fully understand. One option is to simply disable these specific tests on 1.10 (as we did for Enzyme + TO for a while), another is to wait for Enzyme experts to have a look and help us figure out where the problem is. Anyone have any thoughts?

lkdvos:
I would be fine disabling this for now and leaving an issue open, it's a bit silly to block ourselves for this

kshyatt:
FWIW I have noticed that if I force TensorKit to empty the global caches after each `test_reverse`/`test_forward` in `test/enzyme-linalg/mul.jl`, it appears to """fix""" things on 1.10. That would also be a way to keep testing this there without completely disabling it?

kshyatt:
Actually on 1.10, we could just test with a slightly more "slender" domain which has only 2 complex spaces in the product space, no cache emptying needed. I think that works best and I will leave comments in the file explaining as well.

kshyatt:
Anyone have more comments here?

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -0,0 +1,274 @@
...
+    ) where {RT}
+    # ΔC′ = ΔC*β + C*Δβ + A*B*Δα + ΔA*B*α + A*ΔB*α
+    if !isa(C, Const)
+        scale!(C.dval, β.val)
+        !isa(β, Const) && add!(C.dval, C.val, β.dval)
+        !isa(A, Const) && mul!(C.dval, A.dval, B.val, α.val, One())
+        !isa(B, Const) && mul!(C.dval, A.val, B.dval, α.val, One())
+    end
+    if !isa(α, Const) && !isa(C, Const)
+        if iszero(β.val) && !iszero(α.val)
```
lkdvos, line 79: I think this branch is not tested by the values in the testsuite because alpha and beta are generic random values there.

kshyatt: Shall I add another test case for the non-CI case, then? I'm worried about timing out the CI 😅

kshyatt: Added something very similar to what we have in TensorOperations, but I'm not running most of it on CI (I did run it locally) in order to avoid over-burdening it

lkdvos (approved):

PR #455, opened 2026-06-22
Remove explicit dependence on cuTENSOR for CUDA ext

### LINKED ISSUE #454: CUDA functionality is hidden behind cuTENSOR dependency
@kshyatt I bumped into this recently, and this is probably one of my mistakes that we might try and fix:
Most of the functionality of the CUDA extension does not actually require cuTENSOR, and this was mostly there because TensorOperations only had a cuTENSOR backend, while now Strided *should just work* (🤞).
We could try and separating that out, and only require cuTENSOR for specifically accessing the faster kernels.

kshyatt:
Addresses https://github.com/QuantumKitHub/TensorKit.jl/issues/454

With the generic Strided.jl GPU support, we no longer *need* cuTENSOR to support TensorOperations on `CuTensorMap` -- but if the user does load `cuTENSOR`, TO can still use the happy path it has for this library.


lkdvos (approved):

Jutho (approved):

kshyatt:
Fails are unrelated, I'm going to merge

PR #461, opened 2026-06-24
Try running Mooncake linalg tests with CuTensorMaps

kshyatt:


lkdvos (commented):
For runtime purposes, do you think it makes sense to instead of calling the full `test_rule`, we could just compare the values with the CPU versions? Or do we care a bit more about the interface tests as well (and can we check only those?)

kshyatt:
I think it can make sense to check that the memory is unchanged considering sometimes stuff allocs a new array on CUDA that doesn't on the CPU. Let me see if that is handled in the interface or correctness part of the Mooncake test utils.

kshyatt:
Looks like the memory address checking happens in the interface tests for reverse rules. Let's then restrict to interface-only on CI, we can keep the full suite when running locally.

lkdvos (approved):

PR #464, opened 2026-06-26
Add Enzyme rules for factorizations

kshyatt:
For some reason `svd_compact` and `svd_trunc_no_error` don't play nicely with Enzyme here using the MAK rules, I think because of the `DiagonalTensorMap` output. I can try to investigate further if preferred. I also added some additional logic in pullbacks to match what the MAK pullbacks actually kick back.


kshyatt:
Oh crap lol I was working on top of a `dev`-ed MAK

kshyatt:
Weirdly the eig failures seem to go away on `MatrixAlgebraKit` master, let me try with some `[sources]` and see if it's reproducible outside of my laptop

kshyatt:
This `map` business with `block` is not the nicest but I thought it was better than letting `block(nothing, c)` go through silently in other cases

PR #465, opened 2026-06-26
Exponential

sanderdemeyer:
This changes the current way of calculating the exponential (using `LinearAlgebra`) to the one recently implemented in `MatrixAlgebraKit` (see QuantumKitHub/MatrixAlgebraKit.jl#94). This will not yet work, since there is no new version of MatrixAlgebraKit tagged yet. I just wanted to start the discussion here already.

I added both `exp!(t)` and `exp!(tau, t)`. I added the special case where t is real and tau is complex, when `exp!` cannot overwrite the original tensor and instead a new one is allocated. Suggestions on how to best do this are definitely appreciated.

I still need to add some tests. We can definitely add some tests on whether the output tensor has the expected `scalartype`, and whether `exp!(copy(a)) == exp!((1.0, copy(a))`, but suggestions on more nontrivial tests are also welcome.


lkdvos (commented):
We should probably discuss together how to best approach this, because I guess the question is if we want to implement the MatrixAlgebraKit functionality and interface directly on the tensors first, like we did for the factorizations.

I didn't even remember we have `exp!`, which is as far as I can tell defined in TensorKit itself, which means we could in principle indeed decide to use that for `exp!((tau, t))`, although my initial feeling seems to gravitate more towards just overloading the MatrixAlgebraKit versions.
That would then also at least expose the algorithm selection and in-place versions a little better?

sanderdemeyer:
I can definitely get behind just overloading the `MatrixAlgebraKit` version. Maybe we want to keep `exp!` as a (deprecated) version just to avoid breaking changes?

src/factorizations/matrixalgebrakit.jl
```diff
@@ -62,6 +64,31 @@ for f! in (
...
+        Nblock′ = exponential!((τ, tblock), Nblock, alg)
+        # deal with the case where the output is not the same as the input
+        Nblock === Nblock′ || copy!(Nblock, Nblock′)
+        return nothing
+    end
+    return N
+end
+
+# Default algorithm for exponential with Tuple
+MAK.exponential!((τ, t)::Tuple{E, T}) where {E <: Number, T <: DiagonalTensorMap} = MAK.exponential!((τ, t), DefaultAlgorithm())
```
sanderdemeyer: This is a line that I'm a bit sceptical about. I don't know why this has to be included, since I think it should be covered by the functions above, but if I don't include this, the algorithm that is selected for `DiagonalTensorMap`s is the default (`LinearAlgebra.exp!`) instead of the desired `DiagonalAlgorithm`.

sanderdemeyer:
The diagonal case needed some extra care (see also the comment above). So if there's any suggestions on tests, that would be greatly appreciated. That would also illuminate whether there are any additional type restrictions needed in MatrixAlgebraKit (preferably before tagging the new version).

src/factorizations/matrixalgebrakit.jl
```diff
@@ -62,6 +64,31 @@ for f! in (
...
+MAK.exponential!(t::Tuple{E, T}, ::DefaultAlgorithm) where {E <: Number, T <: Diagonal} = MAK.exponential!(t, DiagonalAlgorithm())
+MAK.exponential!(t::Tuple{E, T1}, out::T2, ::DefaultAlgorithm) where {E <: Number, T1 <: Diagonal, T2 <: Diagonal} = MAK.exponential!(t, out, DiagonalAlgorithm())
```
lkdvos: These definitions should probably go into MatrixAlgebraKit, see also e.g. https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/11096c1667ca0c08cb15eaa7e8388a55a0abe24a/src/implementations/lq.jl#L89-L98

src/factorizations/matrixalgebrakit.jl
```diff
@@ -62,6 +64,31 @@ for f! in (
...
+
+function MAK.default_algorithm(::typeof(exponential!), ::Type{Tuple{E, T}}; kwargs...) where {E <: Number, T}
```
lkdvos:
```suggestion
function MAK.default_algorithm(::typeof(exponential!), ::Type{Tuple{E, T}}; kwargs...) where {E <: Number, T <: AbstractTensorMap}
```

Although I think it might be better to replace this in MAK: https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/11096c1667ca0c08cb15eaa7e8388a55a0abe24a/src/interface/exponential.jl#L37-L39
should just forward to the non-tuple case, which will then select the default.

src/factorizations/matrixalgebrakit.jl
```diff
@@ -221,3 +249,9 @@ MAK.initialize_output(::typeof(project_antihermitian!), tsrc::AbstractTensorMap,
     tsrc
 MAK.initialize_output(::typeof(project_isometric!), tsrc::AbstractTensorMap, ::AbstractAlgorithm) =
     similar(tsrc)
+
+# Exponential
+# ----------------
+MAK.initialize_output(::typeof(exponential!), t::AbstractTensorMap, ::AbstractAlgorithm) = t
+MAK.initialize_output(::typeof(exponential!), (τ, t)::Tuple{Number, AbstractTensorMap}, ::AbstractAlgorithm) = t
+MAK.initialize_output(::typeof(exponential!), (τ, t)::Tuple{T1, AbstractTensorMap{T2}}, ::AbstractAlgorithm) where {T1 <: Complex, T2 <: Real} = similar(complex(t))
```
lkdvos:
```suggestion
MAK.initialize_output(::typeof(exponential!), (τ, t)::Tuple{T1, AbstractTensorMap{T2}}, ::AbstractAlgorithm) where {T1 <: Complex, T2 <: Real} = similar(t, complex(eltype(t)))
```

Otherwise this allocates twice, once for the `complex` call and another time for `similar` 😉

src/tensors/linalg.jl
```diff
@@ -421,7 +421,23 @@ function exp!(t::TensorMap)
     domain(t) == codomain(t) ||
         error("Exponential of a tensor only exist when domain == codomain.")
     for (c, b) in blocks(t)
-        copy!(b, LinearAlgebra.exp!(b))
+        MatrixAlgebraKit.exponential!(b, b)
```
lkdvos: I think here we should probably replace the **entire** function, not just the blockwise one

src/tensors/linalg.jl
```diff
@@ -421,7 +421,23 @@ function exp!(t::TensorMap)
...
+    end
+    return t
+end
+
+function exp!((τ, t)::Tuple{Number, TensorMap})
```
lkdvos: I wouldn't really introduce this new functionality into `exp!`, and just call `exponential!` directly if the scaled version is required.

src/factorizations/matrixalgebrakit.jl
sanderdemeyer: I pushed this in QuantumKitHub/MatrixAlgebraKit.jl#253.

src/factorizations/matrixalgebrakit.jl
sanderdemeyer: What is then your suggested change in MAK? Because there we don't specify anything (in both `exponential!(t)` as `exponential!((tau,t))` ). Do you want to restrict it to `A <: AbstractMatrix` there?

src/tensors/linalg.jl
sanderdemeyer: This is now outdated, since we deprecate `exp!` and just use `exponential!` instead.

src/factorizations/matrixalgebrakit.jl
lkdvos:
```julia
function default_algorithm(::typeof(exponential!), ::Type{A}; kwargs...) where {A}
    return default_exponential_algorithm(A; kwargs...)
end

function default_algorithm(::typeof(exponential!), ::Tuple{A, B}; kwargs...) where {A, B}
    return default_algorithm(exponential!, B; kwargs...)
end
function default_algorithm(::typeof(exponential!), ::Type{Tuple{A, B}}; kwargs...) where {A, B}
    return default_algorithm(exponential!, B; kwargs...)
end
```

Something like this?

lkdvos:
Confirms this builds fine on top of https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/255

test/tensors/exponential.jl
```diff
@@ -0,0 +1,64 @@
...
+    if st1 <: Real && st2 <: Complex
+        @test objectid(A2) != objectid(A)
+    else
+        @test objectid(A2) == objectid(A)
+    end
+
+    expτA = exponential!((τ, copy(A)))
+    expminτA = exponential!((-τ, copy(A)))
+    @test expτA * expminτA ≈ id(scalartype(expτA), space)
+    @test expτA ≈ inv(expminτA)
```
Jutho: I think this is needlessly testing the same thing twice.

sanderdemeyer: I removed the second test.

lkdvos (approved):
As MAK v0.6.9 is now released, I think this one is ready for review.

For me this looks great, thanks @sanderdemeyer for the work (as well as the work that is less visible here making sure that the MAK interface and implementations is compatible)!

PR #466, opened 2026-06-26
Enzyme reverse mode rules and test for index manipulations

kshyatt:
Also consolidated the pullbacks for Enzyme and Mooncake


kshyatt:
The `permute` fail I really cannot reproduce locally on Ubuntu and 1.10 at all... how frustrating!

Jutho:
General question: separating all these tests over different folders now causes a large number of CI jobs, some of which only take in the order of one minute. If I open one of them, the actual tests took 0.7 seconds, so that is one minute of overhead I guess to launch the CI job and set up the environment etc. Is that still efficient?

It seems useful to bundle CI jobs such that the actual launching overhead is perhaps in the order of 5 - 10 %, so jobs of about 10 - 20 minutes? Or am I overlooking something?

kshyatt:
I think now that we fixed the "reinstalling and precompiling all the packages" issue by properly using the caches, these can indeed be consolidated (previously they were each taking 30mins+). I'll do that today.

ext/TensorKitEnzymeExt/indexmanipulations.jl
```diff
@@ -0,0 +1,266 @@
...
+        ::Type{RT},
+        C::Annotation{<:AbstractTensorMap},
+        A::Annotation{<:AbstractTensorMap},
+        p::Const{<:Index2Tuple},
+        levels::Const{<:IndexTuple},
+        α::Annotation{<:Number},
+        β::Annotation{<:Number},
+        ba::Const...
+    ) where {RT}
+    C_cache = !isa(β, Const) ? deepcopy(C.val) : nothing
```
Jutho: Any reason for `deepcopy` here? For `permute/transpose` above you use `copy`, which I think should be sufficient.

ext/TensorKitEnzymeExt/indexmanipulations.jl
```diff
@@ -0,0 +1,266 @@
...
+    A_cache = EnzymeRules.overwritten(config)[3] ? deepcopy(A.val) : nothing
```
Jutho: Same.

ext/TensorKitEnzymeExt/indexmanipulations.jl
kshyatt: Failed debugging attempt :(

Jutho:
> I think now that we fixed the "reinstalling and precompiling all the packages" issue by properly using the caches, these can indeed be consolidated (previously they were each taking 30mins+). I'll do that today.

But I guess that now the overhead is indeed "only 1 minute"; before that I assume it would have been much worse even to split up all these jobs.

Jutho:
This looks good, but I am still surprised that we even need to define rules for `flip` and `insert/remove unit`, which are not difficult but also not entirely straightforward with the `copy` kwarg etc. I would think this is exactly where Mooncake and Enzyme would have made our life easier.

kshyatt:
I think Enzyme must have made some internal fixes too. Previously they had to be split up because even after the package compilation was done, the tests themselves took 45+ mins each which exceeded the timeout.

kshyatt:
> This looks good, but I am still surprised that we even need to define rules for `flip` and `insert/remove unit`, which are not difficult but also not entirely straightforward with the `copy` kwarg etc. I would think this is exactly where Mooncake and Enzyme would have made our life easier.

I agree... I would like to try pruning these down after this goes in and seeing what minimal rule set we can get away with.

ext/TensorKitEnzymeExt/indexmanipulations.jl
```diff
@@ -0,0 +1,266 @@
...
+                config::EnzymeRules.RevConfigWidth{1},
+                func::Const{typeof($insertunit)},
+                ::Type{RT},
+                cache,
+                tsrc::Annotation{<:AbstractTensorMap},
+                ival::Const{<:Val};
+                kwargs...
+            ) where {RT}
+            tsrc_cache, tdst, Δtdst = cache
+            # note: since data is already shared for <:TensorMap, don't have to do anything here!
```
lkdvos, line 213: If we are not doing anything, I don't really understand why we are copying a `tsrc_cache` (line 190), can we remove that one or do we need it anywhere?

ext/TensorKitEnzymeExt/indexmanipulations.jl
```diff
@@ -0,0 +1,266 @@
...
+        func::Const{typeof(removeunit)},
+        ::Type{RT},
+        tsrc::Annotation{<:AbstractTensorMap},
+        ival::Const{<:Val};
+        kwargs...
+    ) where {RT}
+    # tdst shares data with tsrc if <:TensorMap & copy=false, in this case we have to deal with correctly
+    # sharing address spaces
+    if tsrc.val isa TensorMap && !get(kwargs, :copy, false) && !isa(tsrc, Const)
+        tsrc_cache = copy(tsrc.val)
```
lkdvos: same comment here

ext/TensorKitEnzymeExt/indexmanipulations.jl
kshyatt: It's a bit difficult to reason about these chains of conditions but I think we shouldn't need it since we don't have to restore the state of `tsrc` in Enzyme, good catch

kshyatt: I think we can just set it to `tsrc_cache = tsrc.val` just to make sure the logic in the reverse pass vis a vis `nothing` works

kshyatt:
As discussed on Zulip, I'm going to merge this once CI passes and then take another run at getting Enzyme and Mooncake to generate good rules themselves.

lkdvos (approved):

kshyatt:
AMD fail can't be related, going to merge!

PR #469, opened 2026-06-30
Consolidate Enzyme tests

kshyatt:
Fixes to the caching for CI runs and some Enzyme improvements have cut the runtime of Enzyme tests substantially, so we can probably fold these back into smaller files to reduce the amount of CI runs.


test/enzyme/linalg.jl
```diff
@@ -0,0 +1,145 @@
...
+        αβs = !is_ci ? vcat(zero_αβs..., (randn(T), randn(T))) : ((randn(T), randn(T)),)
+        for TC in (Duplicated,), TA in (Duplicated,), TB in (Duplicated,)
+            for (α, β) in αβs
+                rTαs = if α === Zero()
+                    (Const,)
+                elseif !is_ci
+                    (Active, Const)
+                else
+                    (Active,)
+                end
```
lkdvos:
```suggestion
                rTαs = []
                α === Zero() || push!(rTαs, Active)
                (α === Zero() || !is_ci) && push!(rTαs, Const)
```

Just a small suggestion for code style, I tend to find it a bit easier to see what is going on if there aren't too many lines

Jutho: I don't quite like that this is now changed just here in this one instance, while the test file is full of these constructs. I don't mind the three case separation and keeping `rTαs` a tuple (though clearly that doesn't have any real advantages here). I would suggest to revert it to the previous state, unless @lkdvos feels strongly, but then they should probably be changed everywhere

lkdvos: I kind of meant to change it everywhere but got pulled into a meeting midway through my suggestions 😁

PR #471, opened 2026-07-01
Fix Enzyme index manipulation spaces to match Mooncake

kshyatt:


kshyatt:
Even after a bunch of fixes for these they are still failing, and of course in a way that I cannot reproduce locally on my Ubuntu machine or on Mac M4. Maybe we should just turn these off for now, or `allow-fail` them, until I can figure out the cause? Sorry for messing up CI 😓

kshyatt:
After some Zulip discussion with @lkdvos , I think it makes sense to merge this with the large runner count just to unblock CI for everyone, then make another PR where we stop using `test_reverse` for stuff because it's extremely unstable and instead try directly using `autodiff` with a randomly projected gradient.

lkdvos (approved):
Definitely unfortunate to have to deal with this, but in any case I do agree to just deal with the added compute cost for now, while we work on getting this sorted in the longer term.

PR #474, opened 2026-07-06
Reverse/forward Enzyme rules, refactoring, and tests for planar ops

kshyatt:


kshyatt:
Marking `FusionTreeDict` inactive seems to have helped, now the errors are all in the situation of the more complex braidings. Time to investigate more 🤓

kshyatt:
OK so some digging seems to show the error for `VfHubbard` and `VRepA4Twistedℤ₄` is arising from the coefficient * alpha going into `tensortrace!`, a path that is only reached for these more complicated spaces. I've tried a bunch of things -- caching doesn't seem to be the culprit -- but still pretty mystified. Basically, it seems one subblock is getting the wrong coefficient on the reverse pass. If anyone has any ideas, chime in!

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -11,7 +11,7 @@ function EnzymeRules.augmented_primal(
         α::Annotation,
         β::Annotation,
     ) where {RT}
-    cacheC = !isa(β, Const) && copy(C.val)
+    cacheC = !isa(β, Const) ? copy(C.val) : C.val
     cacheA = !isa(B, Const) && EnzymeRules.overwritten(config)[3] ? copy(A.val) : nothing
     cacheB = !isa(A, Const) && EnzymeRules.overwritten(config)[4] ? copy(B.val) : nothing
```
lkdvos:
```suggestion
cacheA = !isa(B, Const) && EnzymeRules.overwritten(config)[3] ? copy(A.val) : A.val
cacheB = !isa(A, Const) && EnzymeRules.overwritten(config)[4] ? copy(B.val) : B.val
```

I'm not really sure I am understanding everything here, but would this not follow the same strategy?

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -39,13 +39,8 @@ function EnzymeRules.reverse(
...
-    if RT <: Const
-        Δα = isa(α, Const) ? nothing : zero(α.val)
-        Δβ = isa(β, Const) ? nothing : zero(β.val)
-        return (nothing, nothing, nothing, Δα, Δβ)
-    end
     cacheC, cacheA, cacheB, AB = cache
-    Cval = something(cacheC, C.val)
+    Cval = cacheC
     Aval = something(cacheA, A.val)
     Bval = something(cacheB, B.val)
```
lkdvos: In case the above is accepted, I can't suggest this but:

```julia
Cval, Aval, Bval, AB = cache
```

ext/TensorKitEnzymeExt/utility.jl
```diff
@@ -15,9 +15,18 @@ pullback_dC!(ΔC, β::Number) = scale!(ΔC, conj(β))
...
 
+@inline EnzymeRules.inactive_type(::Type{TensorKit.GenericFusion}) = true
+@inline EnzymeRules.inactive_type(::Type{TensorKit.UniqueFusion}) = true
+@inline EnzymeRules.inactive_type(::Type{TensorKit.MultipleFusion}) = true
+@inline EnzymeRules.inactive_type(::Type{TensorKit.SimpleFusion}) = true
+@inline EnzymeRules.inactive_type(::Type{TensorKit.MultiplicityFreeFusion}) = true
 @inline EnzymeRules.inactive_type(::Type{<:TensorKit.FusionTree}) = true
+@inline EnzymeRules.inactive_type(::Type{<:TensorKit.FusionTreeDict}) = true
+@inline EnzymeRules.inactive_type(::Type{<:TensorKit.FusionTreeBlock}) = true
+@inline EnzymeRules.inactive_type(::Type{<:TensorKit.FusionTreePair}) = true
```
lkdvos: This one might be type piracy, fusiontreepair is just a const for `Tuple{FusionTree, FusionTree}`.

ext/TensorKitEnzymeExt/utility.jl
```diff
@@ -15,9 +15,18 @@ pullback_dC!(ΔC, β::Number) = scale!(ΔC, conj(β))
...
 @inline EnzymeRules.inactive_type(::Type{<:TensorKit.GenericTreeTransformer}) = true
 @inline EnzymeRules.inactive_type(::Type{<:TensorKit.VectorSpace}) = true
+@inline EnzymeRules.inactive_type(::Type{<:TensorKit.LRU}) = true
```
lkdvos, line 28: Are we allowed to do this, or does this also count as type piracy?

lkdvos (commented):
Some minor comments, but otherwise this looks llike a nice cleanup!

ext/TensorKitEnzymeExt/utility.jl
kshyatt: I think we're allowed to? Probably the LRUCache.jl people don't want an Enzyme specific extension, but I had issues with Enzyme trying to differentiate through LRU's `get!` which was pretty silly

ext/TensorKitEnzymeExt/utility.jl
kshyatt: Hm, but then isn't it also bad form to mark `Index2Tuple` as inactive?

kshyatt: Let's see if we can avoid marking it

ext/TensorKitEnzymeExt/utility.jl
lkdvos: It might be reasonable to try and upstream this, but I'm definitely okay with not blocking this PR for that.

kshyatt: Is this otherwise good to merge, then?

ext/TensorKitEnzymeExt/linalg.jl
```diff
@@ -39,16 +39,7 @@ function EnzymeRules.reverse(
...
-        Δα = isa(α, Const) ? nothing : zero(α.val)
-        Δβ = isa(β, Const) ? nothing : zero(β.val)
-        return (nothing, nothing, nothing, Δα, Δβ)
-    end
-    cacheC, cacheA, cacheB, AB = cache
-    Cval = something(cacheC, C.val)
-    Aval = something(cacheA, A.val)
-    Bval = something(cacheB, B.val)
-
+    Cval, Aval, Bval, AB = cache
```
lkdvos, line 42: I think you now also removed the special case for constant C but non-constant alpha/beta, was this intentional?

lkdvos (approved):
Will already approve, but have 1 final comment left

ext/TensorKitEnzymeExt/linalg.jl
kshyatt: Yes it was, the presence of that case (`RT <: Const`) (which was wrong) was why the `planarcontract` test was failing

Issue #475, opened 2026-07-08
Unnecessary Set allocation in sectorequal/issetequal for Trivial sectors


VinceNeede:
# `sectorequal`/`issetequal` for Trivial-sector spaces materializes a full hash Set to compare "0 or 1 elements"

## Summary

`TensorKit.sectorequal`, for `Trivial`-sector spaces, falls through to
`Base.issetequal`'s fully generic fallback. That fallback materializes a
full hash `Set` (a `Dict` plus 3 backing `Memory` arrays) on each side,
just to compare two `TensorKit.OneOrNoneIterator{Trivial}` objects — each
of which can hold **at most one element**. This shows up just from
constructing ordinary `TensorMap`s (no contraction, no `TensorOperations`
even required — see reproducer below), and in our own downstream package
accounted for roughly 35-40% of total allocations in two structurally
different algorithms built on TensorKit.

## Minimal reproducer

Bisected down to the simplest trigger found: **no contraction is
needed at all**. Repeatedly constructing two structurally-distinct
(transposed-shape) `TensorMap`s — nothing done with them beyond that —
is enough, with fixed (not even varying) dimensions:

```julia
using TensorKit

function build_pair()
    A = randn(ComplexF64, ℂ^3 ← ℂ^4)
    B = randn(ComplexF64, ℂ^4 ← ℂ^3)
    return A, B
end

for _ in 1:100
    build_pair()
end
```

Allocation-by-type breakdown over these 100 iterations (full script,
including the `Profile.Allocs` setup, in the collapsible section below):

```
800 allocs  (62.5 KiB)  <-  Dict{Trivial, Nothing}
800 allocs  (25.0 KiB)  <-  Memory{UInt8}
800 allocs  (25.0 KiB)  <-  Vector{Tuple{Any, Any}}
800 allocs  (12.5 KiB)  <-  Memory{Trivial}
800 allocs  (12.5 KiB)  <-  Memory{Nothing}
600 allocs  (14.1 KiB)  <-  TensorKit.DegeneracyStructure{2}
200 allocs  (9.4 KiB)   <-  TensorKit.Hashed{TensorMapSpace{ComplexSpace, 1, 1}, typeof(TensorKit.sectorhash), typeof(TensorKit.sectorequal)}
```

(That last line is worth noting on its own — `sectorequal` appears
directly in the type name of another frequently-allocated object here,
independent confirmation alongside the stack trace below.)

<details>
[...]

PR #476, opened 2026-07-08
Bypass some overhead in TensorMap constructor, bypass overhead in `sectorequal`/`sectorhash` for trivial symmetry

lkdvos:
As discussed with @Jutho and @kshyatt before, this PR consists of two changes:

The first (set) is:
- The inner constructors for `TensorMap` are now no longer doing the checks, bypassing a small amount of overhead for looking up some of the cached data, as well as a small amount of overhead for the warnings with `maxlog = 1`.
- The outer constructors now have these warnings/checks to keep the external behavior more or less the same

The second is:
- `sectorequal` and `sectorhash` can be specialized for `Trivial` symmetry type, as brought up in #475.

Note that this is only a partial fix for #475, and to thoroughly fix some of this we would probably rather try to avoid caching the symmetry structure, which requires slightly more plumbing which I would like to hold off on for now, if this turns out to not be required it's a bunch of specialized code to maintain and check.

### DIFF
--- a/src/spaces/cartesianspace.jl
+++ b/src/spaces/cartesianspace.jl
@@ -74,5 +74,5 @@ sectortype(::Type{CartesianSpace}) = Trivial
 
 Base.show(io::IO, V::CartesianSpace) = print(io, "ℝ^$(V.d)")
 
-sectorhash(::CartesianSpace, h::UInt) = h
-sectorequal(::CartesianSpace, ::CartesianSpace) = true
+sectorhash(V::CartesianSpace, h::UInt) = hash(dim(V) == 0, h)
+sectorequal(V₁::CartesianSpace, V₂::CartesianSpace) = iszero(dim(V₁)) == iszero(dim(V₂))
--- a/src/spaces/cartesianspace.jl
+++ b/src/spaces/cartesianspace.jl
@@ -73,3 +73,6 @@ sectors(V::CartesianSpace) = OneOrNoneIterator(dim(V) != 0, Trivial())
 sectortype(::Type{CartesianSpace}) = Trivial
 
 Base.show(io::IO, V::CartesianSpace) = print(io, "ℝ^$(V.d)")
+
+sectorhash(::CartesianSpace, h::UInt) = h
+sectorequal(::CartesianSpace, ::CartesianSpace) = true
--- a/src/spaces/complexspace.jl
+++ b/src/spaces/complexspace.jl
@@ -83,5 +83,9 @@ sectortype(::Type{ComplexSpace}) = Trivial
 
 Base.show(io::IO, V::ComplexSpace) = print(io, isdual(V) ? "(ℂ^$(V.d))'" : "ℂ^$(V.d)")
 
-sectorhash(V::ComplexSpace, h::UInt) = hash(isdual(V), h)
-sectorequal(V₁::ComplexSpace, V₂::ComplexSpace) = isdual(V₁) == isdual(V₂)
+function sectorhash(V::ComplexSpace, h::UInt)
+    return hash(dim(V) == 0, hash(isdual(V), h))
+end
+function sectorequal(V₁::ComplexSpace, V₂::ComplexSpace)
+    return isdual(V₁) == isdual(V₂) && iszero(dim(V₁)) == iszero(dim(V₂))
+end
--- a/src/spaces/complexspace.jl
+++ b/src/spaces/complexspace.jl
@@ -82,3 +82,6 @@ sectors(V::ComplexSpace) = OneOrNoneIterator(dim(V) != 0, Trivial())
 sectortype(::Type{ComplexSpace}) = Trivial
 
 Base.show(io::IO, V::ComplexSpace) = print(io, isdual(V) ? "(ℂ^$(V.d))'" : "ℂ^$(V.d)")
+
+sectorhash(V::ComplexSpace, h::UInt) = hash(isdual(V), h)
+sectorequal(V₁::ComplexSpace, V₂::ComplexSpace) = isdual(V₁) == isdual(V₂)
--- a/src/tensors/tensor.jl
+++ b/src/tensors/tensor.jl
@@ -10,6 +10,13 @@ a tensor category), where the data is stored in a dense vector.
 struct TensorMap{T, S <: IndexSpace, N₁, N₂, A <: DenseVector{T}} <: AbstractTensorMap{T, S, N₁, N₂}
     data::A
     space::TensorMapSpace{S, N₁, N₂}
+
+    # explicit inner constructor to prevent auto-generating TensorMap(data, space)
+    function TensorMap{T, S, N₁, N₂, A}(
+            data::A, space::TensorMapSpace{S, N₁, N₂}
+        ) where {T, S <: IndexSpace, N₁, N₂, A <: DenseVector{T}}
+        return new{T, S, N₁, N₂, A}(data, space)
+    end
 end
 
 """
--- a/src/tensors/tensor.jl
+++ b/src/tensors/tensor.jl
@@ -10,30 +10,7 @@ a tensor category), where the data is stored in a dense vector.
 struct TensorMap{T, S <: IndexSpace, N₁, N₂, A <: DenseVector{T}} <: AbstractTensorMap{T, S, N₁, N₂}
     data::A
     space::TensorMapSpace{S, N₁, N₂}
-
-    # uninitialized constructors
-    function TensorMap{T, S, N₁, N₂, A}(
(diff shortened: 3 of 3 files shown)


src/spaces/cartesianspace.jl
```diff
@@ -73,3 +73,6 @@ sectors(V::CartesianSpace) = OneOrNoneIterator(dim(V) != 0, Trivial())
 sectortype(::Type{CartesianSpace}) = Trivial
 
 Base.show(io::IO, V::CartesianSpace) = print(io, "ℝ^$(V.d)")
+
+sectorhash(::CartesianSpace, h::UInt) = h
+sectorequal(::CartesianSpace, ::CartesianSpace) = true
```
Jutho: I think we technically do not return a sector if the dimension is zero, so I'd rather had something like:


```suggestion
sectorhash(V::CartesianSpace, h::UInt) = hash(dim(V) == 0, h)
sectorequal(V₁::CartesianSpace, V₂::CartesianSpace) = iszero(dim(V₁)) == iszero(dim(V₂))
```

src/spaces/complexspace.jl
```diff
@@ -82,3 +82,6 @@ sectors(V::ComplexSpace) = OneOrNoneIterator(dim(V) != 0, Trivial())
 sectortype(::Type{ComplexSpace}) = Trivial
 
 Base.show(io::IO, V::ComplexSpace) = print(io, isdual(V) ? "(ℂ^$(V.d))'" : "ℂ^$(V.d)")
+
+sectorhash(V::ComplexSpace, h::UInt) = hash(isdual(V), h)
+sectorequal(V₁::ComplexSpace, V₂::ComplexSpace) = isdual(V₁) == isdual(V₂)
```
Jutho: Similar reasoning:

```suggestion
function sectorhash(V::ComplexSpace, h::UInt)
    return hash(dim(V) == 0, hash(isdual(V), h))
end
function sectorequal(V₁::ComplexSpace, V₂::ComplexSpace)
    return isdual(V₁) == isdual(V₂) && iszero(dim(V₁)) == iszero(dim(V₂))
end
```

VinceNeede:
<html><head></head><body><p>I tested an earlier version of this branch (commit <code>6f6ea02</code>, before the recent test-related changes) against a downstream MPS/MPO tensor-network package I maintain (built on TensorKit) — its own real benchmark suite, not just the earlier synthetic MWE: <code>PkgBenchmark.judge</code> comparing this branch against the currently released TensorKit, across MPO-MPS application (three algorithms) and DMRG, at multiple system sizes and bond dimensions.</p>
<p>Allocation counts for DMRG dropped ~25-30% consistently:</p>

DMRG benchmark | branch allocs | released allocs | ratio
-- | -- | -- | --
dmrg_tfim_L20_open, nsite=1 | 604,339 | 857,127 | 0.71
dmrg_tfim_L20_open, nsite=2 | 635,210 | 840,741 | 0.76
dmrg_tfim_L20_periodic, nsite=1, maxdim=10 | 695,778 | 957,995 | 0.73
dmrg_tfim_L20_periodic, nsite=2, maxdim=10 | 732,462 | 948,522 | 0.77


<p>(<code>judge</code>'s own comparison doesn't report allocation counts directly — these are computed from the raw per-side results.)</p>
<p>More broadly, essentially every MPO-MPS-application benchmark across all three algorithms and both tested sizes shows a real improvement — time ratios mostly 0.67–0.99, memory ratios 0.39–1.00, with zero memory regressions anywhere in the suite. DMRG's own time ratios land in a similar 0.88–0.99 range.</p>
<p>Happy to share the full <code>judge</code> markdown output if useful, or re-run once the current changes land.</p></body></html>

Jutho:
That is great to hear. Can I ask why you want to create a new MPS/MPO package, instead of using MPSKit, or, if you find that functionality is missing, contributing to it? I see in the description of your package that you think that MPSKit is somehow "VUMPS first", but that doesn't make any sense. MPSKit has full support for standard finite MPS, including DMRG, TDVP and a bunch of other functionality (like finite size excitations).

Jutho (approved):

Jutho:
Also, @VinceNeede , if you are interested in a "finite MPS first" package, there is also https://github.com/Qiaoyi-Li/FiniteMPS.jl .

lkdvos:
Happy to hear the changes are helping in any case. I'm running our own benchmark suite locally as well and will get back with the results.

A couple remarks I still have though: looking at just the allocation number is often a misleading metric for the actual performance - the allocations we are saving here are of the order of bytes, not the actual kilo or megabytes for the data of the tensors themselves. While it is true that this puts some pressure on the garbage collector, realistically tthis should be subdominant in most tensor network applications.

The fact that your timings are affected by this change so drastically might be something worth looking into - I of course don't have much detail about what you are running, but it seems like if you are getting speedups of ~30%, that means more than ~30% of your runtime is spent in constructing tensors.
For example in MPSKit, almost all of the time is spent in the tensor contractions and factorizations, which is more along the lines of what you would expect, and why we haven't investigated in a lot of detail to iron out these remaining performance bits.
Probably you would want to have a look and see if you are not creating too many tensors in hot loops.

Of course, a lot of time, effort and expertise has already gone in these packages that Jutho mentioned, so I would also strongly advise to always consider contributing first, before feeding all this effort of other people into an LLM and simply building your own copy.

VinceNeede:
Hi @Jutho, thanks for reaching out. While this PR might not be the best place for a long architectural discussion, I'd love to clear up the context.

This honestly started as a personal passion project to learn the ropes, even before I was fully aware of TensorKit. When I briefly looked at MPSKit, I noticed the absence of direct MPO-MPS contraction algorithms (like zip-up or density matrix methods) and found its handling of orthogonality quite complex compared to what I was used to in frameworks like ITensor. This led me to assume the package simply had a different scope from what I wanted to build.

Because I was largely driven by the personal satisfaction of building something from scratch, my evaluation of MPSKit was admittedly superficial, and I completely take your point regarding its robust finite-size support. My "VUMPS-first" characterization was definitely off-base.

That said, I'd much rather collaborate than duplicate effort. If you think any of the features I'm working on—like zip-up/density matrix contractions, successive randomized compression, DMRG3S for single-site subspace expansion, or future open quantum system algorithms—would be valuable additions to MPSKit, I would be incredibly glad to contribute them!

VinceNeede:
Hi @lkdvos, Thanks for the feedback.

​To clarify the timings, the significant speedup I observed was specifically at lower bond dimensions, while higher bond dimensions were almost unaffected. The high allocation count was what prompted me to do a finer profiling in the first place, as I was worried something was off.

​Regarding your other point, my intention was never to disrespect or copy anyone's hard work. While I have experience with MPS from my master's thesis, I didn't feel confident enough to contribute directly to an advanced ecosystem like TensorKit/MPSKit initially, which is why I started a personal project to learn. I use an LLM purely as an interactive pair-programmer executing under my own direction—for brainstorming architectural ideas, debugging specific edge-case failures, and handling maintenance tasks like writing tests and benchmarks—not to copy existing codebases; in fact, since some algorithms I am focusing on (like DMRG3S and Successive Randomized Compression) aren't currently available out-of-the-box in the major packages, I am implementing them directly from the papers.

​I'm glad the issue I opened helped improve things here. As I mentioned to Jutho, I would much rather collaborate than duplicate effort, and I'd be very glad to contribute any of the algorithms I'm working on to MPSKit if they fit your roadmap.

lkdvos:
### Contraction benchmarks: `ld-constructor` vs `main`

| Group  | Cases | Median (dirty/main) |
|:-------|:-----:|:-------------------:|
| mera   |  24   |        0.99         |
| mpo    |  32   |        1.00         |
| pepo   |  32   |        0.99         |

In general I think this confirms the claim that this does not show up for most larger tests :)

---

Concerning the MPSKit features, I am always happy to have more contributors and to include more of the missing functionality. The goal of that package and this ecosystem in general has always been to be a collaborative experience, so if you feel up for it I will gladly invest time discussing and going over ideas, features and PRs, as well as answer any questions you might have.

VinceNeede:
That matches what I'd been seeing too @lkdvos  — reassuring to see it hold up on a broader test suite.

Really appreciate the offer to invest time in this — I'd like to take you up on it. I have a couple of specific things I want to work through (one about CBE, one about whether/how SRC could fit given its current limitations) — where would be the best place to continue, a GitHub Discussion, an issue, or somewhere else?

lkdvos:
My email is on my github profile, if you'd like to schedule an online meeting I'm happy to discuss! (I'm in ET time zone)

VinceNeede:
Thanks, @lkdvos  — I'll email you shortly to set up a time!

PR #478, opened 2026-07-09
Mark more error strings lazy

kshyatt:
Makes JET happier


lkdvos (commented):
For the debug messages, is this necessary to make JET happy? It does not necessarily make that much sense there since that is anyways only evaluated if the debug messages are enabled, in which case we have to instantiate that string anyways

kshyatt:
Yes, Jet was complaining about them. I agree it's goofy.

lkdvos (approved):
🤷

PR #479, opened 2026-07-12
perf: Trivial symmetry performance improvements

lkdvos:
This is a slight follow-up on #476 which reduces a bit more of the overhead for Trivial sectortypes. Additionally a slight refactor to avoid some unnecessary respecialization for `contract!` implementations.

### DIFF
--- a/src/tensors/tensoroperations.jl
+++ b/src/tensors/tensoroperations.jl
@@ -324,31 +324,22 @@ function contract!(
     )
     length(pA[2]) == length(pB[1]) ||
         throw(IndexError("number of contracted indices does not match"))
-    N₁, N₂ = length(pA[1]), length(pB[2])
 
     # find optimal contraction scheme by checking the following options:
     # - sorting the contracted inds of A or B to avoid permutations
     # - contracting B with A instead to avoid permutations
+    pA′, pB′, pA″, pB″, pAB′ = _contract_candidates(pA, pB, pAB)
 
-    qA = TupleTools.sortperm(pA[2])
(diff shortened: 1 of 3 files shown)


lkdvos:
@kshyatt do you know if these AMD failures are the ones from the `conj` thing in Strided?

kshyatt:
I see these intermittently on 1.10 and 1.12, I think they are unrelated to this PR. We should open an issue and I'll see if I can figure out what is going wrong.

src/tensors/tensoroperations.jl
```diff
@@ -361,23 +356,49 @@ function contract!(
...
         B::AbstractTensorMap, pB::Index2Tuple,
         pAB::Index2Tuple
     )
     ipAB = TO.oindABinC(pAB, pA, pB)
-    return dim(A) * (!TO.isblascontractable(A, pA) || eltype(A) !== eltype(C)) +
-        dim(B) * (!TO.isblascontractable(B, pB) || eltype(B) !== eltype(C)) +
+    return dim(A) * (!TO.isblascontractable(A, pA) || scalartype(A) !== scalartype(C)) +
+        dim(B) * (!TO.isblascontractable(B, pB) || scalartype(B) !== scalartype(C)) +
         dim(C) * !TO.isblasdestination(C, ipAB)
 end
```
Jutho: Do we still need this one, if TensorKit now calls its own `_contract_memcost`?

Jutho (approved):
One remark, but otherwise looks good.

PR #480, opened 2026-07-13
Enzyme doesn't need a custom rule for trace_permute

kshyatt:
Noticed that this seemed to be working sans custom rule locally. Let's see how it does for all the test cases.


kshyatt:
Apparently this works now with a custom set of rules for `subblock`!

lkdvos (commented):
🥳

lkdvos (approved):

kshyatt:
It seems to have helped with compile times too 🧐

Jutho (approved):

PR #482, opened 2026-07-14
Try to get SVD tests to chill out

kshyatt:


kshyatt:
Think this got resolved and it turned out to be an issue with AMDGPU CI itself

PR #483, opened 2026-07-14
No need for custom twist rule

kshyatt:
Also seems to work fine letting Enzyme build a rule on its own


Jutho (approved):
Less Enzyme code hooray 🥳 !

kshyatt:
I'm working on other methods as I go but trying to prune things back with a minimum of fuss

PR #485, opened 2026-07-14
docs: Add docstrings for catdomain and catcodomain

### LINKED ISSUE #484: `catdomain` and `catcodomain` are exported but undocumented
`catdomain` and `catcodomain` are both exported from TensorKit, but `?catdomain` / `?catcodomain` return nothing — there's no docstring attached to either.

The explanation does exist, just not attached to the functions — it's prose on the [linear algebra docs page](https://quantumkithub.github.io/TensorKit.jl/dev/man/linearalgebra/#ss_tensor_linalg):

> If we have two `TensorMap{T, S, N, 1}` instances `t1` and `t2` with the same codomain, we can combine them in a way that is analogous to `hcat`, i.e. we stack them such that the new tensor `catdomain(t1, t2)` has also the same codomain, but has a domain which is `domain(t1) ⊕ domain(t2)`. Similarly, if `t1` and `t2` are of type `TensorMap{T, S, 1, N}` and have the same domain, the operation `catcodomain(t1, t2)` results in a new tensor with the same domain and a codomain given by `codomain(t1) ⊕ codomain(t2)`, which is the analogy of `vcat`. Note that the direct sum only makes sense between `ElementarySpace` objects, i.e. there is no way to give a tensor product meaning to a direct sum of tensor product spaces.

It might be nice to have this (or a shorter version of it) directly on the functions as a docstring, so `?catdomain` works without needing to already know where to look for it.

Happy to open a PR adding docstrings based on the existing explanation above, if that's welcome.

VinceNeede:
Closes #484.

Adds docstrings for `catdomain` and `catcodomain`, based on the existing explanation in the [linear algebra docs page](https://quantumkithub.github.io/TensorKit.jl/dev/man/linearalgebra/#ss_tensor_linalg), including the codomain/domain-duality requirements enforced by the implementation and a runnable example for each.

### DIFF
--- a/src/tensors/linalg.jl
+++ b/src/tensors/linalg.jl
@@ -467,7 +467,32 @@ for f in (:sqrt, :log, :asin, :acos, :acosh, :atanh, :acoth)
     end
 end
 
-# concatenate tensors
+"""
+    catdomain(t1::AbstractTensorMap{<:Any, S, N₁, 1}, t2::AbstractTensorMap{<:Any, S, N₁, 1}) where {S, N₁}
+
+Given two tensors that share the same codomain, and whose domain is a single
+`ElementarySpace` (rank 1) with matching duality, return a new tensor `t` with that
+same codomain and domain `V = domain(t1) ⊕ domain(t2)` — the tensor-map analogue of
+`hcat`.
+
+Throws a `SpaceMismatch` if the codomains don't match, or if `domain(t1)` and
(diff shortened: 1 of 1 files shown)


lkdvos (approved):
Thanks! Will merge this in a bit without the full test suite, as we still have some other PRs running and I want to avoid restarting their full CI.

PR #486, opened 2026-07-14
Boil down indexmanip rules to add_transform

kshyatt:
Enzyme can't reverse diff through `add_transform` (yet) without some rules for the `OhMyThreads.jl` functions, which I was a little lazy about adding. But it can make a nice forward rule, it seems.

### DIFF
--- a/ext/TensorKitEnzymeExt/indexmanipulations.jl
+++ b/ext/TensorKitEnzymeExt/indexmanipulations.jl
@@ -1,87 +1,27 @@
-for transform in (:permute, :transpose)
-    transform! = Symbol(transform, :!)
-    transform_pb = Symbol(transform, :_pullback_dA)
-    @eval function EnzymeRules.augmented_primal(
-            config::EnzymeRules.RevConfigWidth{1},
-            func::Const{typeof(TK.$transform!)},
-            ::Type{RT},
-            C::Annotation{<:AbstractTensorMap},
-            A::Annotation{<:AbstractTensorMap},
-            p::Const{<:Index2Tuple},
-            α::Annotation{<:Number},
-            β::Annotation{<:Number},
-            ba::Const...
-        ) where {RT}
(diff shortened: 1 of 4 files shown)


ext/TensorKitEnzymeExt/indexmanipulations.jl
```diff
@@ -91,24 +31,33 @@ function EnzymeRules.augmented_primal(
...
     bavs = map(a -> a.val, ba)
     # ΔA
     if !isa(A, Const) && !isa(C, Const)
-        TK.braid_pb(A.dval, Aval, C.dval, C.val, p.val, levels.val, α.val, bavs...)
+        ip = invperm(linearize(p.val))
+        pΔA = TO.repartition(ip, numout(Aval))
+        TC = VectorInterface.promote_scale(Cval, α.val)
+        if scalartype(A.dval) <: Real && !(TC <: Real)
+            ΔAc = TO.tensoralloc_add(TC, C.dval, pΔA, false, Val(false))
+            TK.add_transform!(ΔAc, C.dval, pΔA, transformer.val, conj(α.val), Zero(), bavs...)
```
lkdvos, line 56: I'm really surprised this works though - the transformer is the metadata needed to efficiently permute and recouple, but this depends on the value of `p` so unless this is accidentally a self-inverse permutation where `C` and `A` have the same structure, this should probably give wrong results?

kshyatt: I was testing this on spaces that are "too easy" maybe, for the A4 ones it doesn't work :(

lkdvos: Yeah, the problem is that at this point we no longer have access to what kind of transformer it is, so it is not possible to construct the reverse one from that.

kshyatt: But my negative lines diff!!!!!!

PR #487, opened 2026-07-14
Precompilation

lkdvos:
This PR adds a precompilation suite to TensorKit.
It takes about 300 seconds on my machine, and is enabled by default.
Given the general migration of the ecosystem towards more heavy precompilation workflows, I think this is warranted, and I'm hoping this improves our CI times as well as the general user experiences.

I've also structured it in such a way that you could conceivably reuse this locally for projects with specific symmetry needs, to ensure faster startup times.
As in general TensorKit is not updated all that often, I hope this is not something that has to be retriggered too much, therefore I felt the ~5mins of precompilation time to be reasonable.


docs/src/man/precompilation.md
```diff
@@ -0,0 +1,44 @@
...
+
+Changing a preference triggers recompilation of TensorKit the next time it is loaded.
+
+Downstream packages or startup files that define or heavily use their own symmetry can precompile TensorKit's operations for it by
+calling the `precompile_*` helpers inside their own `PrecompileTools.@compile_workload`:
+
+```julia
+using TensorKit, PrecompileTools
+@compile_workload begin
+    TensorKit.precompile_contract(Vect[MyAnyon](s₀ => 1, s₁ => 1))
```
borisdevos, line 36: Since we're defaulting to unit spaces for the sectors from TensorKit, maybe it's better to say that that suffices in this docstring?

Also, I prefer `MySector` or `CustomSector` to `MyAnyon` as a name.

On another note, shouldn't you call `TensorKit.Precompilation.precompile_x`, or otherwise put `using .Precompilation` underneath its `include` in `TensorKit.jl`?

docs/src/man/precompilation.md
```diff
@@ -0,0 +1,44 @@
...
+
+Because the numerically-heavy kernels are mostly *sector-agnostic* (they depend only on the element type, the arity, and `sectorscalartype`),
+this workload also speeds up the first operation on symmetries that are *not* in the workload — including user-defined ones.
+
+The workload can be tuned or disabled through [Preferences](https://github.com/JuliaPackaging/Preferences.jl):
+
+```julia
+using TensorKit, Preferences, PrecompileTools
+# disable the workload entirely (fastest precompilation, no TTFX benefit)
+PrecompileTools.set_preferences!(TensorKit, "precompile_workloads" => false; force=true)
```
borisdevos, line 15: I think `TensorKit` here needs to be replaced by `PrecompileTools`, at least from what I understand [here](https://julialang.github.io/PrecompileTools.jl/stable/#Package-developers:-reducing-the-cost-of-precompilation-during-development). The other option is "precompile_workload" in singular form, which I think is more what you want to achieve here.

src/precompile.jl
```diff
@@ -0,0 +1,71 @@
...
+    (n isa Int && n ≥ 1) ||
+        throw(ArgumentError("`precompile_ndims` should be a positive `Int`, got `$n`"))
+    n
+end
+
+function _precompile_space(name::AbstractString)
+    I = Base.eval(TensorKit, Meta.parse(name))
+    (I isa DataType && I <: Sector) ||
+        error("invalid `precompile_sectors` entry `$name`; expected a `Sector` type")
+    return oneunit(Vect[I])
```
borisdevos, line 39:
```suggestion
    return unitspace(Vect[I])
```

src/tensors/indexmanipulations.jl
```diff
@@ -560,15 +560,16 @@ Base.@deprecate(
     if p[1] === codomainind(tsrc) && p[2] === domainind(tsrc)
         add!(tdst, tsrc, α, β)
     else
+        p2 = (linearize(p), ())
```
borisdevos, line 563: Sorry for my naive question here (I'm not too familiar with this part of the code), but I guess passing the same type of tuple, one containing the indices and the other always being empty, is beneficial for precompilation, and somehow doesn't matter within TensorOperations thatindices are grouped up like this now? So why would you not make the same changes to `transpose!` and `braid` in the cases that they don't pass `add_transform!` first?

lkdvos: The main point is that for anything that is just raw data (i.e. stridedview), this does not matter and therefore it is beneficial to pick a single canonical option that is then only compiled once. For the tensormaps themselves, this is actually more difficult to get to because the spaces as well as the recoupling data need to be partitioned correctly

lkdvos:
@borisdevos I'd be very interested in any timing results you have for the MPSKit testsuite on main TensorKit vs on this branch, in case you were already investigating

borisdevos:
https://quantumkithub.github.io/TensorKit.jl/previews/PR487/man/precompilation/ for the relevant docs preview. Completely unrelated to this PR, but I noticed in the docs that the contents of tensors only show the first page, namely "Constructing tensors and the `TensorMap` type", and its subsections.

Besides that, with my one data point on the MPSKit tests, it seems this precompilation is useful, but my machine is so garbage that there's lots of noise, so I'm testing some of the heavier compilation files more.

src/precompile/contract.jl
```diff
@@ -0,0 +1,59 @@
...
+Run a small, representative set of contraction, trace, and permutation operations on tensors built
+from the space `V = unitspace(S)`, for each element type in `eltypes` and each tensor arity (number of
+legs) in `1:ndims`.
+
+Note that it can be beneficial to put a more comprehensive set of relevant symmetries in a startup file,
+for example by adding:
+
+```julia
+@compile_workload begin
+    TensorKit.precompile_contract(Vect[MySector])
```
borisdevos, line 13:
```suggestion
    TensorKit.Precompilation.precompile_contract(Vect[MySector])
```
I think? Same for the other 2 docstrings.

lkdvos: they are actually exported from the precompilation module, so this is not required.

lkdvos:
Test failures are unrelated, @Jutho ok if I merge this?

src/precompile/precompile.jl
```diff
@@ -0,0 +1,71 @@
...
+
+function _precompile_spacetype(name::AbstractString)
+    I = Base.eval(TensorKit, Meta.parse(name))
+    (I isa DataType && I <: Sector) ||
+        throw(ArgumentError(lazy"invalid `precompile_sectors` entry `$name`; expected a `Sector` type"))
+    return Vect[I]
+end
+
+const PRECOMPILE_SECTORS = let s = @load_preference(
+        "precompile_sectors", map(x -> repr(x; context = :module => TensorKit), [Trivial, Z2Irrep, SU2Irrep, FermionParity])
```
Jutho: No U1Irrep?

src/precompile/factorizations.jl
```diff
@@ -0,0 +1,62 @@
...
+"""
+function precompile_factorizations(::Type{S}; eltypes = PRECOMPILE_ELTYPES) where {S <: IndexSpace}
+    V = unitspace(S)
+    for T in eltypes
+        # Three representative shapes: a square tensor for the bulk of the decompositions, a
+        # hermitian one for the `eigh` path, and rectangular ones so the null-space kernels
+        # operate on non-empty blocks.
+        W = V^2   # V ⊗ V
+        t = randn(T, W ← W)            # square
+        tr = randn(T, W ← V)           # tall (codomain larger) -> non-empty left null space
```
Jutho, line 27: What is the `r` in `tr` for?

Jutho (approved):

src/precompile/factorizations.jl
lkdvos: rectangular, but 🤷 because I ran out of letters with the wide one 😆

Jutho: `tt` for tall? 🙃

PR #488, opened 2026-07-15
Try letting Enzyme make rules for insert/remove unit

kshyatt:


ext/TensorKitEnzymeExt/utility.jl
```diff
@@ -52,6 +52,39 @@ function EnzymeRules.reverse(
...
 end
 
+function EnzymeRules.augmented_primal(
+        config::EnzymeRules.RevConfigWidth{1},
+        ::Const{typeof(block)},
+        ::Type{RT},
+        t::Annotation{<:AbstractTensorMap},
+        c::Annotation{<:Sector},
+    ) where {RT}
+    ret = EnzymeRules.needs_primal(config) ? block(t.val, c.val) : nothing
```
Jutho, line 58: Just a question out of curiosity: if `needs_primal(config) == false`, is it an error to still return the value anyways, or are you required to return `nothing` for the first argument to `AugmentedReturn`.

ext/TensorKitEnzymeExt/utility.jl
```diff
@@ -52,6 +52,39 @@ function EnzymeRules.reverse(
...
+        ::Const{typeof(block)},
+        ::Type{RT},
+        cache,
+        t::Annotation{<:AbstractTensorMap},
+        c::Annotation{<:Sector},
+    ) where {RT}
+    dret = cache
+    if !isnothing(dret) && !isa(t, Const)
+        block(t.dval, c.val) .= dret
+    end
```
Jutho: Is this actually doing something? As far as I understand, `cache` contains `dret` which already corresponds to `block(t.dval, c.val)`. So is this just copying this variable into itself?

Jutho: Probably the same question applies to the `subblock` rule from yesterday's PR.

ext/TensorKitEnzymeExt/utility.jl
kshyatt: I believe you're supposed to return `nothing` in that case

ext/TensorKitEnzymeExt/utility.jl
kshyatt: Yes, I was trying to be extra cautious for the case where the block has been reassigned between the forward step and reverse pass (which I guess could happen for `BlockTensorMap` or something), but perhaps it's not needed...

kshyatt: Tried locally and it seems we're ok without it for now!

ext/TensorKitEnzymeExt/utility.jl
Jutho: Ok, maybe let's see if @lkdvos  has any remarks regarding the BlockTensor case. Also, if we don't need this, maybe it makes sense to also remove this for `subblock`?

kshyatt: Already done :)

lkdvos: I'm really not sure how all of this will interact with blocktensorkit, but it seems reasonable to leave that up to BlockTensorKit. I can't immediately spot glaring issues, and it's a bit hard to see how it will all pan out without just doing it

ext/TensorKitEnzymeExt/utility.jl
```diff
@@ -45,10 +45,35 @@ function EnzymeRules.reverse(
...
+end
+
+function EnzymeRules.reverse(
+        config::EnzymeRules.RevConfigWidth{1},
+        ::Const{typeof(block)},
+        ::Type{RT},
+        cache,
+        t::Annotation{<:AbstractTensorMap},
+        c::Annotation{<:Sector},
+    ) where {RT}
```
lkdvos, line 76: Is this `nothing` because `block` always gives a view?

kshyatt: I think so? You mean why the `reverse` doesn't do anything?

lkdvos: yes, as in you'd expect that you would have to accumulate the incoming block into the full tensor, but since they share data there is nothing left to do, but I was staring at this and that seems counterintuitive at first 🙃

lkdvos (approved):

PR #493, opened 2026-07-27
change blocktype of TensorMap to `StridedView`

kshyatt:
This is https://github.com/QuantumKitHub/TensorKit.jl/pull/342 rebased on top of `main` but I opened a new branch to avoid overwriting that one


kshyatt:
I'm still working on the GPU stuff but the others look p good, nice

PR #498, opened 2026-07-29
Fix bugs in `GradedSpace` constructor + `similar` of `SectorVector`

borisdevos:
Two small bugs I encountered.

1) Keyword arguments were splatted when forwarded to `Vect[I]`, so `dual` was read as `:dual`.
2) One of the `similar`s just `MethodError`ed.


lkdvos (commented):
Good catch! Any chance you can add a test for this to catch this in the future?

test/tensors/sectorvector.jl
```diff
@@ -0,0 +1,142 @@
...
+
+        @timedtestset "AbstractVector interface ($T)" begin
+            v = SectorVector{T}(undef, V)
+            rand!(parent(v))
+            @test @testinferred(eltype(v)) == T
+            @test length(v) == length(parent(v))
+            oldval = v[1]
+            v[1] = one(T)
+            @test v[1] == one(T) == parent(v)[1]
+            v[1] = oldval
```
Jutho: Is there any use or meaning to doing this last line?

borisdevos: No, this is a remnant of a mess I made 🙃

lkdvos (approved):

PR #500, opened 2026-07-30
Try combining the mul/planarcontract tests

kshyatt:
For more complicated braidings, `planarcontract` calls `mul!` so I would like to see if combining these cuts down test times and memory usage at all.


lkdvos (approved):
Seems to be working, ok for me if tests pass

lkdvos (approved):

PR #502, opened 2026-08-04
Pass allocator through to transpose! in planar operations

liuzhaochen:
planaradd! and planarcontract! call transpose! without the allocator argument, so internal transform buffers fall back to DefaultAllocator (GC) instead of the caller's allocator.


lkdvos (approved):
Thank you, good catch!

PR #503, opened 2026-08-04
Do not allow duplicate `levels` in anyonic `braid`

borisdevos:
Previously, equal `levels` were allowed to be passed to `braid`, which would then silently overbraid. The docstring was not clear about this being allowed, so I changed some things to just simply forbid it. There were also some inconsistencies in docs/docstrings on how to interpret levels, so now I put "depths" everywhere as the synonym to them.

### DIFF
--- a/docs/src/man/indexmanipulations.md
+++ b/docs/src/man/indexmanipulations.md
@@ -67,7 +67,7 @@ removeunit(::AbstractTensorMap, ::Val{i}) where {i}
 These operations reorder indices and/or move them between domain and codomain by applying the transposing or braiding isomorphisms of the underlying category.
 They form a hierarchy from most general to most restricted:
 
-- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of heights, one per index — that determines whether each index crosses over or under the others it has to pass.
+- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of depths, one per index — that determines whether each index crosses over or under the others it has to pass.
 - [`permute`](@ref) is a simpler interface for sector types with a symmetric braiding (`BraidingStyle(I) isa SymmetricBraiding`), where over- and under-crossings are equivalent and `levels` is therefore not needed.
 - [`transpose`](@ref) is restricted to *cyclic* permutations (indices do not cross).
 - [`repartition`](@ref) only moves the codomain/domain boundary without reordering the indices at all.
--- a/docs/src/man/indexmanipulations.md
+++ b/docs/src/man/indexmanipulations.md
@@ -67,7 +67,7 @@ removeunit(::AbstractTensorMap, ::Val{i}) where {i}
 These operations reorder indices and/or move them between domain and codomain by applying the transposing or braiding isomorphisms of the underlying category.
 They form a hierarchy from most general to most restricted:
 
-- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of depths, one per index — that determines whether each index crosses over or under the others it has to pass. In particular, duplicate levels are not allowed, and the length of `levels` must match the number of indices in the tensor.
+- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of depths, one per index — that determines whether each index crosses over or under the others it has to pass. In particular, duplicate levels are not allowed when the braiding is anyonic (`BraidingStyle(I) isa Anyonic`), and the length of `levels` must match the number of indices in the tensor.
 - [`permute`](@ref) is a simpler interface for sector types with a symmetric braiding (`BraidingStyle(I) isa SymmetricBraiding`), where over- and under-crossings are equivalent and `levels` is therefore not needed.
 - [`transpose`](@ref) is restricted to *cyclic* permutations (indices do not cross).
 - [`repartition`](@ref) only moves the codomain/domain boundary without reordering the indices at all.
--- a/docs/src/man/indexmanipulations.md
+++ b/docs/src/man/indexmanipulations.md
@@ -67,7 +67,7 @@ removeunit(::AbstractTensorMap, ::Val{i}) where {i}
 These operations reorder indices and/or move them between domain and codomain by applying the transposing or braiding isomorphisms of the underlying category.
 They form a hierarchy from most general to most restricted:
 
-- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of depths, one per index — that determines whether each index crosses over or under the others it has to pass.
+- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of depths, one per index — that determines whether each index crosses over or under the others it has to pass. In particular, duplicate levels are not allowed, and the length of `levels` must match the number of indices in the tensor.
 - [`permute`](@ref) is a simpler interface for sector types with a symmetric braiding (`BraidingStyle(I) isa SymmetricBraiding`), where over- and under-crossings are equivalent and `levels` is therefore not needed.
 - [`transpose`](@ref) is restricted to *cyclic* permutations (indices do not cross).
 - [`repartition`](@ref) only moves the codomain/domain boundary without reordering the indices at all.
--- a/src/auxiliary/auxiliary.jl
+++ b/src/auxiliary/auxiliary.jl
@@ -84,7 +84,7 @@ _alldistinct(t::Tuple) = !in(t[1], tail(t)) && _alldistinct(tail(t))
     _check_levels(levels::Tuple) -> Nothing
 
 Check whether all elements of a tuple are distinct, throwing an `ArgumentError` if not.
-This is used to validate the `levels` of `braid`.
+This is used to validate the `levels` of [`braid`](@ref).
 """
 _check_levels(levels::Tuple) = _alldistinct(levels) || _throw_levels(levels)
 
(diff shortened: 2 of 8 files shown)


docs/src/man/indexmanipulations.md
```diff
@@ -67,7 +67,7 @@ removeunit(::AbstractTensorMap, ::Val{i}) where {i}
 These operations reorder indices and/or move them between domain and codomain by applying the transposing or braiding isomorphisms of the underlying category.
 They form a hierarchy from most general to most restricted:
 
-- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of heights, one per index — that determines whether each index crosses over or under the others it has to pass.
+- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of depths, one per index — that determines whether each index crosses over or under the others it has to pass.
```
kshyatt: should this explicitly mention that duplicated depths are forbidden?

src/auxiliary/auxiliary.jl
```diff
@@ -76,3 +76,18 @@ end
...
 end
+
+_alldistinct(::Tuple{}) = true
+_alldistinct(t::Tuple) = !in(t[1], tail(t)) && _alldistinct(tail(t))
+
+"""
+    _check_levels(levels::Tuple) -> Nothing
+
+Check whether all elements of a tuple are distinct, throwing an `ArgumentError` if not.
+This is used to validate the `levels` of `braid`.
```
kshyatt: should `braid` get an `@ref` link here?

src/tensors/indexmanipulations.jl
```diff
@@ -293,8 +293,10 @@ end
 
 Compute `tdst = β * tdst + α * braid(tsrc, (p₁, p₂), levels)`, writing the result into `tdst`.
 The codomain and domain of `tdst` correspond to the indices in `p₁` and `p₂` of `tsrc` respectively.
-Here, `levels` is a tuple of length `numind(tsrc)` that assigns a level or height to the indices of `tsrc`,
```
kshyatt, line 296: another chance here to mention that duplicated levels are BANNED

test/symmetries/doubletree.jl
```diff
@@ -174,6 +174,12 @@ using TensorKitSectors
     end
 
     BraidingStyle(I) isa HasBraiding && @testset "Double fusion tree: permutation and braiding" begin
+        n = rand(0:(2N))
+        p1, p2 = p[1:n], p[(n + 1):(2N)]
+        levels = ntuple(identity, 2N)
+        l1, l2 = levels[1:N], levels[(N + 1):(2N)]
+        @test_throws ArgumentError TensorKit.braid(src, (p1, p2), (TupleTools.setindex(l1, l2[1], 1), l2))
```
kshyatt: can we check the whole argument error string here?

test/tensors/indexmanipulations.jl
```diff
@@ -140,3 +140,13 @@ for V in spacelist
     end
     TensorKit.empty_globalcaches!()
 end
+
+@timedtestset "braid: invalid levels" begin
+    t = rand(ComplexF64, ℂ^2 ⊗ ℂ^3 ← ℂ^4)
+    p = ((2, 1), (3,))
+    @test braid(t, p, (1, 3, 2)) isa TensorMap
+    @test_throws ArgumentError braid(t, p, (1, 2, 2)) # duplicate levels
```
kshyatt: similar to above comment here

lkdvos:
As you can tell in the braidingtensor implementation, I was actually making use of the fact that you only need distinct levels for legs that will actually cross. This might actually be a useful feature at some point, since in principle we could envision designing an ITensor-like system that holds on to space,level and and id, and at that point it might be annoying to ensure uniqueness.

As an alternative, how about within the actual braiding code where we check which level is larger we add a check that they are not equal whenever the braiding is not symmetric?

borisdevos:
The only downsides here are that the error will be raised much deeper, and that you can do bad homspace braids, but I can live with that :)
I believe with this that `NoBraiding` also still works fine, but that's untested.

docs/src/man/indexmanipulations.md
```diff
@@ -67,7 +67,7 @@ removeunit(::AbstractTensorMap, ::Val{i}) where {i}
 These operations reorder indices and/or move them between domain and codomain by applying the transposing or braiding isomorphisms of the underlying category.
 They form a hierarchy from most general to most restricted:
 
-- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of heights, one per index — that determines whether each index crosses over or under the others it has to pass.
+- [`braid`](@ref) is the most general: it accepts any permutation and requires a `levels` argument — a tuple of depths, one per index — that determines whether each index crosses over or under the others it has to pass. In particular, duplicate levels are not allowed when the braiding is anyonic (`BraidingStyle(I) isa Anyonic`), and the length of `levels` must match the number of indices in the tensor.
```
lkdvos, line 70: Do you know if we are consistently using heights or depths in the rest of the docs/code? I'm kind of impartial for either, but would definitely prefer to stick to one. I'm also not so sure if this is the right place to add the technical information about what the `levels` argument is (and it is quite easy to become outdated, as it currently is 😄 )

lkdvos: [edit] Is the depth because larger level values are braided under lower level values? That does not sound like a great terminology, and if the reverse is true I would actually insist on heights

lkdvos (commented):
So reading this in a bit more detail, I think the goal has always been to have `levels` refer to a height, so larger values braid **over** lower values, and I guess the implementation was wrong (I'm also not entirely sure if we have anything that can test this, as of course this is just the same as using the timereversed category)

test/symmetries/doubletree.jl
```diff
@@ -174,8 +174,15 @@ using TensorKitSectors
...
+        if !(BraidingStyle(I) isa SymmetricBraiding)
+            eq = ntuple(Returns(1), N) # potential problematic level 1
+            err_str = "ambiguous braid: two indices with equal level 1 have to cross"
+
+            @test TensorKit.braid(src, (ntuple(identity, N), ntuple(i -> i + N, N)), (eq, eq)) isa Pair # equal levels are fine when the indices never meet
+            @test_throws ArgumentError(err_str) TensorKit.braid(src, ((2, 1, ntuple(i -> i + 2, N - 2)...), ntuple(i -> i + N, N)), (eq, eq)) # and rejected when they do
+        end
+
         for n in 0:(2N)
-            p = (randperm(2 * N)...,)
```
lkdvos, line 178: probably an accident here

borisdevos:
> So reading this in a bit more detail, I think the goal has always been to have `levels` refer to a height, so larger values braid **over** lower values, and I guess the implementation was wrong (I'm also not entirely sure if we have anything that can test this, as of course this is just the same as using the timereversed category)

This is in contrast with the one example in the [docs](https://quantumkithub.github.io/TensorKit.jl/stable/man/fusiontrees/#Manipulations-on-a-fusion-tree), where level = depth (it's mentioned literally in the text below the image). So that's why I changed everything to be consistent with depth.

The current implemenation checks `inv = levels[s] < levels[s+1]` at fusion tree level, which I think doesn't get changed from the pass at tensor level. As far as I can tell, `braid` -> `braid!` where the tuple gets split up into 2 tuples for the codomain and domain (pure regrouping), and within `fsbraid` the domain levels get reversed for the all outgoing convention. So `levels` is always depths.

docs/src/man/indexmanipulations.md
borisdevos: I answered below, but yes the use of depths is consistent in the code, but it wasn't in the docs. And yes, levels should really be seen from top-down, so it's really depths. We could just change `<` to `>` somewhere, but that seems unnecessarily breaking.

lkdvos:
Well, in that case maybe it's just my brain that always translated it to heights. Good catch, in that case I absolutely agree the language should become depths ;)

src/tensors/indexmanipulations.jl
```diff
@@ -293,8 +293,11 @@ end
...
+Here, `levels` is a tuple of length `numind(tsrc)` that assigns a level or depth to the indices of `tsrc`,
 which determines whether they will braid over or under any other index with which they have to change places.
+In other words, a smaller value in `levels` means that the corresponding index will be braided over any other index with a higher value.
+`levels` can thus not contain duplicate values when the braiding is anyonic, as overbraiding and underbraiding are then not equivalent.
```
```suggestion
For anyonic braiding styles, equal levels can therefore not unambiguously cross.
```

I think this is otherwise not true, you could do `p = ((2, 1), (3, 4, 5, ...)); levels = (1, 2, 3, 3, 3, ...))`, which seems reasonable to me.
I'm also okay with dropping this sentence entirely, I feel like that is kind of implicit in the definition of levels, and while I strongly agree with throwing an explicit error I'm not sure we need the additional mention in the docstring

src/tensors/indexmanipulations.jl
```diff
@@ -323,8 +326,10 @@ end
 
 Return tensor `tdst` obtained by braiding the indices of `tsrc`.
 The codomain and domain of `tdst` correspond to the indices in `p₁` and `p₂` of `tsrc` respectively.
-Here, `levels` is a tuple of length `numind(tsrc)` that assigns a level or height to the indices of `tsrc`,
+Here, `levels` is a tuple of length `numind(tsrc)` that assigns a level or depth to the indices of `tsrc`,
 which determines whether they will braid over or under any other index with which they have to change places.
+In other words, a smaller value in `levels` means that the corresponding index will be braided over any other index with a higher value.
+`levels` can thus not contain duplicate values when the braiding is anyonic, as overbraiding and underbraiding are then not equivalent.
```
lkdvos: Same comment here

lkdvos (commented):
Left some minor comments about wording, otherwise good to go for me!

PR #505, opened 2026-08-12
Fix `@planar` backend and allocator insertion

lkdvos:
By the time the postprocessors run, `_insert_planar_operations` has rewritten the tensor calls to `GlobalRef(TensorKit, :planar*!)`. `TO.insertargument` only matches `GlobalRef`s into `TensorOperations`, so `@planar backend=...` and `@planar allocator=...` silently dropped both arguments from the planar calls.

Adds a `TensorKit`-matching `_insertargument`, routes the `@planar` parser through `insertplanarbackend`/`insertplanarallocator`, and tests the expansion.

### DIFF
--- a/test/tensors/planar.jl
+++ b/test/tensors/planar.jl
@@ -101,6 +101,33 @@ end
 @testset "@planar" verbose = true begin
     T = ComplexF64
 
+    @testset "backend and allocator insertion" begin
+        # trailing arguments of every `planar*!` call in `ex`
+        function planartrailing(ex, out = Any[])
+            ex isa Expr || return out
+            if Meta.isexpr(ex, :call) && ex.args[1] isa GlobalRef &&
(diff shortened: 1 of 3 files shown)


test/tensors/planar.jl
```diff
@@ -101,6 +101,33 @@ end
 @testset "@planar" verbose = true begin
     T = ComplexF64
 
+    @testset "backend and allocator insertion" begin
```
kshyatt, line 104: should we add something similar for the CUDA and AMD planar tests

lkdvos: This is really just testing the macro, the `MarkerBackend` and `MarkerAllocator` don't exist and this is really just about expressions, probably don't need to test the GPU ones explicitly

kshyatt (approved):

PR #506, opened 2026-08-12
Fix `--modules=` parsing in the benchmark entry point

lkdvos:
Two bugs in `benchmark/benchmarks.jl`:

- the pattern captured `(\w+)`, which stops at the first comma, so `--modules=linalg,tensornetworks` silently loaded only `linalg`;
- `only(...)` was applied to the capture rather than the match, throwing for any module name longer than one character.


kshyatt (approved):
It's really abusive to ask me to review a regex

PR #509, opened 2026-08-18
Redo `add_transform!` for GPU backed Tensors to be much, much faster 

### LINKED ISSUE #508: GPU-backed `TensorMap` `permute!` `braid` is extremely inefficient
tl;dr We are doing a separate kernel launch for every element of the fusion tree, which is SUPER wasteful especially for otherwise small tensors.

Here's a sample script:

```Julia
using TensorKit, AMDGPU, Adapt, LinearAlgebra, Printf

function warmup(seconds = 3.0)
    X = AMDGPU.rand(Float32, 2048, 2048)
    t0 = time()
    while time() - t0 < seconds
        X * X
    end
    AMDGPU.synchronize()
    return nothing
end

function bench(f, n; reps = 7)
    f(); AMDGPU.synchronize()
    best = Inf
    for _ in 1:reps
        AMDGPU.synchronize()
        e = @elapsed begin
            for _ in 1:n; f(); end
            AMDGPU.synchronize()
        end
        best = min(best, e / n)
    end
    return best * 1e3
end

nfusiontrees(t) = length(collect(TensorKit.fusiontrees(t)))
nstored(t)      = sum(prod(size(b)) for (_, b) in blocks(t))

function report(name, t_cpu, perm)
    t_gpu = adapt(ROCArray, t_cpu)
    nb, nf, ne = length(collect(blocks(t_cpu))), nfusiontrees(t_cpu), nstored(t_cpu)

    c = bench(() -> permute(t_cpu, perm; copy = true), 200)
    g = bench(() -> permute(t_gpu, perm; copy = true), 200)

    # device allocations issued by ONE permute
    AMDGPU.synchronize()
    s0 = copy(AMDGPU.alloc_stats)
    permute(t_gpu, perm; copy = true)
    AMDGPU.synchronize()
    d = AMDGPU.alloc_stats - s0

    @printf("%-18s trees=%-5d elems=%-6d %-10s  CPU %7.3f ms  GPU %8.3f ms  GPU/CPU %6.2fx  allocs/permute=%d\n",
            name, nf, ne, Base.format_bytes(ne * 8), c, g, g / c, d.alloc_count)
    return nothing
end

println("=== permute cost vs. fusion-tree count (AMDGPU / MI210) ===")
println("(best-of-7 batches of 200; GPU warmed up first)\n")
warmup()

Vp = Vect[SU2Irrep](1//2 => 1)
Vv = Vect[SU2Irrep](0 => 1, 1 => 2)
A  = randn(Float64, Vp ⊗ Vp', Vv ⊗ Vv ⊗ Vv' ⊗ Vv')
perm6 = ((2, 3, 5, 6), (1, 4))
report("SU2 iPEPO tensor", A, perm6)

Wp, Wv = ℂ^2, ℂ^7
B = randn(ComplexF64, Wp ⊗ Wp', Wv ⊗ Wv ⊗ Wv' ⊗ Wv')
report("trivial symmetry", B, perm6)

println()
for nmax in 1:3
    V = Vect[SU2Irrep](j => 2 for j in 0:nmax)
    T = randn(Float64, Vp ⊗ Vp', V ⊗ V ⊗ V' ⊗ V')
    report("SU2 jmax=$nmax", T, perm6)
end

println()
Up = Vect[U1Irrep](1//2 => 1, -1//2 => 1)
Uv = Vect[U1Irrep](-1 => 2, 0 => 3, 1 => 2)
C = randn(Float64, Up ⊗ Up', Uv ⊗ Uv ⊗ Uv' ⊗ Uv')
report("U1", C, perm6)
```

Here are the results I see on an MI210:

```
SU2 iPEPO tensor   trees=42    elems=329    2.570 KiB   CPU   0.123 ms  GPU    2.144 ms  GPU/CPU  17.43x  allocs/permute=23
trivial symmetry   trees=1     elems=9604   75.031 KiB  CPU   0.045 ms  GPU    0.022 ms  GPU/CPU   0.50x  allocs/permute=1

SU2 jmax=1         trees=42    elems=672    5.250 KiB   CPU   0.095 ms  GPU    2.107 ms  GPU/CPU  22.11x  allocs/permute=23
SU2 jmax=2         trees=323   elems=5168   40.375 KiB  CPU   0.726 ms  GPU   15.136 ms  GPU/CPU  20.84x  allocs/permute=121
SU2 jmax=3         trees=1364  elems=21824  170.500 KiB  CPU   2.910 ms  GPU   57.840 ms  GPU/CPU  19.88x  allocs/permute=391

U1                 trees=70    elems=2226   17.391 KiB  CPU   0.053 ms  GPU    1.168 ms  GPU/CPU  21.95x  allocs/permute=1
```

The GPU is way way slower due to all the kernel launches. The `allocs/permute` number tells you how many allocations occurred **per permute**.

I think a good first step would be to try to fuse these into one launch if we can,

kshyatt:
Should close https://github.com/QuantumKitHub/TensorKit.jl/issues/508

The short version of what I've done here is rework `add_transform!` into two kernels with associated helper structures, one for the Abelian case (which is quite simple) and one for the more generic case. Instead of parallelizing over the blocks, which are of pretty different sizes, I parallelize over the output data in `data_dst`. This means a lot more annoying bookkeeping but also a much more uniform workload across the GPU and better usage of the large number of available threads. There are still probably some performance optimizations to be done but I thought this was a) already hard enough to understand and b) pretty compelling!

Here are the results for the sample script I linked in the issue above, on an AMD MI210:

### GPU permute time (ms)

| case | trees | elems | bytes | `main` | `ksh/add_transform` | speedup |
|---|---|---|---|---|---|---|
| SU2 iPEPO tensor | 42 | 329 | 2.6 KiB | 1.915 | **0.033** | 58× |
| trivial symmetry | 1 | 9604 | 75.0 KiB | 0.022 | 0.020 | 1.1× |
| SU2 jmax=1 | 42 | 672 | 5.3 KiB | 1.964 | **0.033** | 60× |
| SU2 jmax=2 | 323 | 5168 | 40.4 KiB | 13.818 | **0.034** | 406× |
| SU2 jmax=3 | 1364 | 21824 | 170.5 KiB | 53.537 | **0.046** | 1164× |
| U1 | 70 | 2226 | 17.4 KiB | 1.064 | **0.020** | 53× |

### Cost per fusion tree (GPU, µs/tree)

| case | trees | `main` | `ksh/add_transform` |
|---|---|---|---|
| SU2 iPEPO tensor | 42 | 45.6 | 0.79 | 
| SU2 jmax=1 | 42 | 46.8 | 0.79 |
| SU2 jmax=2 | 323 | 42.8 | 0.105 |
| SU2 jmax=3 | 1364 | 39.2 | 0.034 |
| U1 | 70 | 15.2 | 0.29 |

So this should hopefully finally let people doing complicated stuff (e.g. anything involving SU(2)) really benefit from the GPU. All the tests passed for me locally.

I still have some lingering questions about where stuff should live. I added some new caches of GPU objects as well to avoid sending things back to the GPU that don't need to go, but maybe those should live in the main package?

### DIFF
--- a/ext/TensorKitGPUArraysExt.jl
+++ b/ext/TensorKitGPUArraysExt.jl
@@ -3,11 +3,11 @@ module TensorKitGPUArraysExt
 using GPUArrays
 using GPUArrays: @allowscalar
 using GPUArrays.KernelAbstractions: @kernel, @index, get_backend
-
+using Adapt
 using Strided: StridedViews
 using MatrixAlgebraKit, Adapt
 using TensorKit
-using TensorKit.TensorOperations: linearize
+using TensorKit.TensorOperations: linearize, DefaultAllocator
 using TensorKit.Factorizations
 using TensorKit.Factorizations: AbstractAlgorithm
 using TensorKit: SectorDict, tensormaptype, scalar, similarstoragetype, AdjointTensorMap, scalartype, project_symmetric_and_check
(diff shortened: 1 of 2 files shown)


kshyatt:
I'm also running the GPU branch of PEPSKit.jl against this to see how much/if it helps

kshyatt:
Well, the short story for now is it doesn't help much because nearly all the time in our PEPSKit runs is spent on the CPU (GPU utilization is quite low). I'll look into that separately but I still think this is a nice bit of progress :)

kshyatt:
OK I've got one more thing to push here to make it play nicely with PEPSKit (cache-related) but most of this is ready to go I think

kshyatt:
OK!!! I modified stuff a bit for the `DEVICE_TRANSFORMER_CACHE` because we need the GPUArrays [caching allocator](https://juliagpu.github.io/GPUArrays.jl/dev/interface/#Caching-Allocator) to really see the benefits of this on PEPSKit (otherwise we choke to death on allocations). The `@uncached` here is to protect the cached GPU-side transformer infos from getting reaped while we still expect them to be live.

ext/TensorKitGPUArraysExt.jl
```diff
@@ -121,5 +122,346 @@ function TensorKit.scalar(t::TensorMap{T, S, 0, 0, <:AnyGPUArray}) where {T, S}
...
+
+_device_transformer(t::TensorKit.AbelianTreeTransformer, p) = DeviceAbelianTreeTransformer(t, p)
+_device_transformer(t::TensorKit.GenericTreeTransformer, p) = DeviceGenericTreeTransformer(t, p)
+
+# COV_EXCL_START
+# kernels are not reachable by coverage
+
+# largest `i` with `offsets[i] <= w`. This corresponds to the
+# block which  this kernel thread will work on.
+@inline function _searchblock(offsets, w)
```
lkdvos, line 336: Is this `Base.searchsortedlast` or `Base.searchsortedfirst`?

ext/TensorKitGPUArraysExt.jl
```diff
@@ -121,5 +122,346 @@ function TensorKit.scalar(t::TensorMap{T, S, 0, 0, <:AnyGPUArray}) where {T, S}
...
+    # two kernels touch disjoint subblocks so the launch order does not matter
+    _launch_abelian!(data_dst, data_src, device.degenerate, α, β, Val(N))
+    _launch_generic!(data_dst, data_src, device, α, β, Val(N))
+    return nothing
+end
+
+function TensorKit.TensorOperations.tensorfree!(C::AnyGPUArray, ::DefaultAllocator)
+    GPUArrays.unsafe_free!(C)
+    return nothing
+end
```
lkdvos: Probably has to go in TensorOperations :)

ext/TensorKitGPUArraysExt.jl
lkdvos: edit: I see now that this is used inside the kernel, which I assume is why that has to be like this

ext/TensorKitGPUArraysExt.jl
```diff
@@ -121,5 +122,346 @@ function TensorKit.scalar(t::TensorMap{T, S, 0, 0, <:AnyGPUArray}) where {T, S}
...
+        else
+            hi = mid - 1
+        end
+    end
+    return lo
+end
+
+# Cartesian coordinates of the `w`-th (0-based) entry of a dense subblock of shape `sz`.
+# Computed once per thread and then reused for every strided view of that subblock.
+# This avoids `StridedView` redoing these divisions on every single element access.
```
lkdvos, line 351: This is slightly confusing to me, isn't this still computed in each kernel call below (line 371)? Keep in mind I might be completely misreading kernels here.

lkdvos (commented):
have only looked at the abelian part for now, looks really cool! I didn't really realize that indeed we can just spawn a thread per element. I am definitely wondering how this affects the case where the tensors are a bit larger and the number of blocks not so severe, for example what would effectively be the case of more MPS-like contractions, for which it might be reasonable to try and generalize/run the benchmarks that are in this repository on GPU as well to get a sense about overall performance implications?

Given that every element is read/written exactly once, it could also be cool to just express these numbers in terms of actual read/write bandwidth percentages, for which 🤖 might be able to help? Not that I'm expecting or requiring any huge results, but I'd be very interested to know how much is left on the table

kshyatt:
> I am definitely wondering how this affects the case where the tensors are a bit larger and the number of blocks not so severe, for example what would effectively be the case of more MPS-like contractions, for which it might be reasonable to try and generalize/run the benchmarks that are in this repository on GPU as well to get a sense about overall performance implications?

Yeah, it's a really good question. I focused on the cases I looked at in the linked issue, but if we have some others I can run I will certainly do that to get a clearer picture.

ext/TensorKitGPUArraysExt.jl
kshyatt: Should I add a comment about this to make it clearer?

ext/TensorKitGPUArraysExt.jl
kshyatt: It's computed just once per kernel thread, though, which becomes more important for the generic case

ext/TensorKitGPUArraysExt.jl
kshyatt: One more `[sources]` entry can't hurt...

kshyatt:
> I didn't really realize that indeed we can just spawn a thread per element.

I did do a bit of testing with these kernels on CPU arrays and there's no real benefit there, this approach is really GPU specific I guess.

kshyatt:
OK, we need a patch version bump at TO. Then, are there any objections to merging this?

ext/TensorKitGPUArraysExt.jl
lkdvos: but aren't we spawning a kernel thread per element?

ext/TensorKitGPUArraysExt.jl
lkdvos: probably not a bad idea!

ext/TensorKitGPUArraysExt.jl
kshyatt: Yeah but each kernel thread is doing multiple individual elements, it first computes this (expensive) bit then does (cheap) additions on top

ext/TensorKitGPUArraysExt.jl
```diff
@@ -121,5 +122,341 @@ function TensorKit.scalar(t::TensorMap{T, S, 0, 0, <:AnyGPUArray}) where {T, S}
...
+        coords = _coordinates(local_w % blocksize, blk.sz, blk.densestrides)
+
+        st_dst, offs_dst = @inbounds structs_dst[blk.dst_offset + i + 1]
+        i_dst = _offset(coords, st_dst, offs_dst)
+
+        # dst_i = β * dst_i + α * Σ_j U[i, j] * permute(src_j, p): each output tree is a
+        # linear combination of the input trees weighted by the recoupling coefficients.
+        # The permutation of src_j was already done by permuting its strides before the
+        # kernel launched.
+        acc = zero(promote_type(eltype(data_src), eltype(coeffs)))
```
lkdvos, line 405: Would this not lead to a very uneven load balance on the kernel threads, as there can be cols = 1 and cols = 100 cases?

ext/TensorKitGPUArraysExt.jl
kshyatt: (In the generic case). In the abelian case it's one element per thread indeed

lkdvos (commented):
Overall definitely a cool set of improvements! I'm still trying to fully wrap my head around this, left some small comments but also maybe a request: could we split this into 2 PRs and treat the Abelian case first to make review simpler?

The main motivation here is that I think mostly the caching could use some work. It seems like we might want to include a `storagetype` parameter in the hash as well as in the transformer data, which should unify the implementation of that across the board?

ext/TensorKitGPUArraysExt.jl
kshyatt: Yeah, I think it would be possible although I'm not sure if we have any fusion trees that actually do have such a case? Can you think of any?

kshyatt:
I'll split out the Abelian case into a new PR but keep this open to let others keep using it w PEPSKit

ext/TensorKitGPUArraysExt.jl
lkdvos: anything SU(3) :)

kshyatt: That's what I get for testing old reliable SU(2).

lkdvos: how dare you not test on something that is used by virtually no-one!

PR #510, opened 2026-08-19
[WIP] cuTENSOR blocksparse implementation

lkdvos:
This is an initial implementation of calling into cuTENSOR's blocksparse contraction, which should be able to speed up our `UniqueFusion` code. There is still a bit of annoyance for supporting conjugation flags, and non-trivial F or R symbol coefficients, where this has to be handled in pre/postprocessing, but it seems at least from my initial local benchmarks that this is still competitive and outperforms our naive implementation.

Still requires quite a bit of cleanup, and there is one design question that I am still unsure about, which considers whether or not we should keep the "BlockSparseStructure" as something cuTENSOR specific moved into the extension, or as a generic thing backends can hook into.
On the one hand nothing about that is cuTENSOR specific, and there is a case to be made that we could even on CPU do this strategy which merges permutation and multiplication, (maybe someone has a RUST library somewhere that he is working on 😉) but unfortunately the annoying thing is that cuTENSOR is really particular about what integer types have to be where.

Anyways, I'll keep working on this but already wanted to open this up to let everyone know this is in progress.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -31,6 +31,7 @@ EnzymeTestUtils = "12d8515a-0907-448a-8884-5fe00fdf1c5a"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 GPUArrays = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
 Mooncake = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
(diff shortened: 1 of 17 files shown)


kshyatt:
> maybe someone has a RUST library somewhere that he is working on 😉

Crab gang rise up!!

kshyatt:
With the Rust library perhaps superseded, what's the status on this?

PR #511, opened 2026-08-21
Specialise `GradedSpace` functions based on storage type

borisdevos:
This is somewhat connected to what I was doing in QuantumKitHub/TensorKitSectors.jl#106, but beneficial for all sector types. Since I've firsthand experienced how my code transitioned from being unusable to performing well by going from `NTuple` storage to `SectorDict` storage, I thought it was about time to look at these storage paths.

I had two options going into this. The one I'm still looking into is seeing whether there's a cutoff (or range) where `NTuple` storage severely starts underperforming. The other approach I'm taking in this PR is to specialise `GradedSpace` functions and constructors based on their storage type. The overarching problems I tried to fix were the following:
- the `NTuple` constructor could take unboundedly long to compile for sector types with many sectors
- the `NTuple` versions of these functions were not making use of the fact that the sector types match, so there's efficient ways to accessing the sectors, knowing always how many there are as well.
- the `SectorDict` versions of these functions were doing more per-pair dictionary work than the algorithm actually needs (extra hash lookups, double work, etc)

Summary of changes I made:
- `GradedSpace{I, NTuple{N,Int}}` constructor: the old constructor built up the dims tuple via `TupleTools.setindex`, which fell back to Base's `ntuple(f, Val(N))`. This requires compiling this for every distinct `N`, which I found to scale terribly with `N`. I first tried building into a vector and then annotating the splat into a tuple, but it turns out that it has a cost that scales with `N`, which dominated for large enough `N`. So now I directly convert the vector through a Base iterator-to-tuple constructor which Julia specialised to make faster depending on N. The `NTuple` `fuse` and `truncate_space` make use of this as well.
- `dim` specialisation: in general I tried avoiding constructing `sectors(V)` where possible, and just directly checking the `NTuple` directly (through `values(I)`) or the pairs in `SectorDict`.
- `⊕`, `⊖`, `infimum`, `supremum` specialisations:
  - `NTuple` storage: the two spaces here are always of the same sector type, so their tuples are aligned. I could just do the appropriate `map` without looking up sectors. Again `sectors(V)` is the plague.
  - `SectorDict` storage: I made use of how the keys are sorted here to merge them in an appropriate way depending on the what the function actually wanted to achieve. These structurally looked the same, so I refactored them into `_sortedmerge`. This outperforms having to work directly with a `SectorDict`.
- `fuse` `SectorDict` path: previously a bunch of `get`s and `setindex!`s were done in the double for-loop on the `SectorDict`, which accumulated inefficiently due to lookup cost for this type of dictionary. Plain `Dict`s don't have this, so I just do the accumulation in this and then sort at the end. The complexity hasn't changed since there's still two for-loops, but there's a speedup.
- `truncate_space` `SectorDict` path: same structure as `fuse` for `SectorDict`s, but now with vectors because you don't have to look up anything along the way.
- Restored binary search in `SectorDict`'s `_searchsortedfirst`: I looked into when this was implemented, and goes back to 2019 back when product sectors didn't even exist. So I guess back then `N` was always fairly small, and the linear search was more efficient. However, it seems now that's not particularly the case, so I took the liberty of having it default to Base's method.

## Benchmarks

I tested Julia 1.10.10 (LTS) and 1.12.6 (stable) since I think those are the two versions most people are on. For the `NTuple` storage sector types, I tested `N = 2 / 8 / 64 / 256 / 1296` with `Z2Irrep / ZNIrrep{8} / Z4Irrep⊠^3 / Z4Irrep⊠^4 / ZNIrrep{6}⊠^6`. I also tested `N = 15625` with `ZNIrrep{5}⊠^6` where possible, which is important to mention. For the `SectorDict` I tested `U1Irrep` with charges `-6:6`, `-50:50`, and `-200:200` (13/101/401 sectors).

And here the many many numbers. I spared my sanity by having a robot friend write this in markdown.

### Constructor compile time (the `Val` effect)

<details>

| Julia | before | after | speedup |
|---|---|---|---|
| 1.10.10 | 104 ms / 144 ms / 697 ms / 1.55 s / 14.3 s | 158 ms / 157 ms / 274 ms / 342 ms / 135 ms | 0.66x / 0.91x / 2.5x / 4.5x / 106x |
| 1.12.6 | 170 ms / 170 ms / 1.01 s / 1.64 s / 11.2 s | 102 ms / 152 ms / 247 ms / 239 ms / 136 ms | 1.67x / 1.12x / 4.1x / 6.9x / 83x |

`N=15625` before does not complete (at least within 5 minutes on my laptop) on either version. After: 973 ms (1.10.10), 1.00 s (1.12.6).

</details>

### NTuple storage (after above compilation time)

<details>

| operation | Julia | before | after | speedup |
|---|---|---|---|---|
| constructor | 1.10.10 | 468 ns / 2.67 µs / 465 µs / 3.83 ms / 99.2 ms | 1.03 µs / 5.57 µs / 41.7 µs / 375 µs / 5.30 ms | 0.45x / 0.48x / 11.1x / 10.2x / 18.7x |
| constructor | 1.12.6 | 466 ns / 2.47 µs / 177 µs / 2.78 ms / 72.4 ms | 538 ns / 2.50 µs / 32.2 µs / 184 µs / 2.59 ms | 0.87x / 0.99x / 5.5x / 15.1x / 28.0x |
| dim | 1.10.10 | 19.0 ns / 45.3 ns / 21.3 µs / 239 µs / 4.43 ms | 10.8 ns / 11.8 ns / 22.8 µs / 306 µs / 5.22 ms | 1.8x / 3.8x / 0.93x / 0.78x / 0.85x |
| dim | 1.12.6 | 19.5 ns / 47.9 ns / 12.7 µs / 260 µs / 4.76 ms | 5.7 ns / 5.1 ns / 7.41 µs / 148 µs / 2.19 ms | 3.4x / 9.4x / 1.7x / 1.8x / 2.2x |
| ⊕ | 1.10.10 | 520 ns / 1.04 µs / 259 µs / 4.45 ms / 102 ms | 10.0 ns / 10.8 ns / 4.56 µs / 16.7 µs / 112 µs | 52.0x / 96.4x / 56.8x / 267x / 909x |
| ⊕ | 1.12.6 | 305 ns / 843 ns / 220 µs / 3.31 ms / 94.9 ms | 5.1 ns / 5.6 ns / 3.35 µs / 16.2 µs / 107 µs | 59.7x / 151x / 65.7x / 205x / 886x |
| ⊖ | 1.10.10 | 138 ns / 384 ns / 260 µs / 5.24 ms / 106 ms | 202 ns / 328 ns / 48.1 µs / 651 µs / 11.1 ms | 0.68x / 1.17x / 5.40x / 8.05x / 9.56x |
| ⊖ | 1.12.6 | 76.3 ns / 222 ns / 205 µs / 3.37 ms / 84.9 ms | 56.2 ns / 131 ns / 21.1 µs / 296 µs / 5.18 ms | 1.4x / 1.7x / 9.7x / 11.4x / 16.4x |
| infimum | 1.10.10 | 542 ns / 1.54 µs / 287 µs / 4.76 ms / 102 ms | 10.8 ns / 11.3 ns / 4.57 µs / 17.3 µs / 125 µs | 50.2x / 136x / 62.9x / 275x / 820x |
| infimum | 1.12.6 | 326 ns / 907 ns / 218 µs / 3.41 ms / 87.6 ms | 5.1 ns / 5.7 ns / 3.35 µs / 12.6 µs / 106 µs | 63.9x / 159x / 65.2x / 270x / 825x |
| supremum | 1.10.10 | 311 ns / 688 ns / 265 µs / 4.48 ms / 99.1 ms | 10.8 ns / 11.3 ns / 4.59 µs / 17.2 µs / 114 µs | 28.7x / 60.8x / 57.8x / 261x / 866x |
| supremum | 1.12.6 | 219 ns / 556 ns / 216 µs / 3.30 ms / 83.1 ms | 5.6 ns / 7.3 ns / 3.39 µs / 12.4 µs / 105 µs | 39.0x / 76.2x / 63.7x / 267x / 790x |
| fuse | 1.10.10 | 268 ns / 1.86 µs / 2.86 ms / 280 ms / 25.9 s | 111 ns / 281 ns / 2.56 ms / 135 ms / 12.4 s | 2.4x / 6.6x / 1.1x / 2.1x / 2.1x |
| fuse | 1.12.6 | 173 ns / 1.69 µs / 2.45 ms / 286 ms / 13.2 s | 16.2 ns / 65.7 ns / 1.16 ms / 135 ms / 12.6 s | 10.7x / 25.7x / 2.1x / 2.1x / 1.05x |
| truncate_space | 1.10.10 | 41.9 ns / 300 ns / 247 µs / 4.52 ms / 230 ms | 145 ns / 217 ns / 32.9 µs / 525 µs / 9.68 ms | 0.29x / 1.4x / 7.5x / 8.6x / 23.8x |
| truncate_space | 1.12.6 | 40.1 ns / 130 ns / 199 µs / 3.20 ms / 91.3 ms | 46.8 ns / 80.4 ns / 18.4 µs / 251 µs / 4.50 ms | 0.86x / 1.6x / 10.8x / 12.7x / 20.3x |

And now just the `N=15625` case separately, also just after only as before doesn't finish:

| operation | 1.10.10 | 1.12.6 |
|---|---|---|
| constructor (warm) | 533 ms | 526 ms |
| dim | 554 ms | 580 ms |
| ⊕ | 1.08 ms | 1.28 ms |
| ⊖ | 1.14 s | 290.7 s cold (see note below) |
| infimum | 937 µs | 1.32 ms |
| supremum | 869 µs | 1.01 ms |
| truncate_space | 972 ms | 1.17 s |
| fuse | not run (O(N^2)) |

On the `⊖`/1.12.6 number: the first call at this N takes ~290-353s on 1.12.6 specifically (reproduced 3×), but the second call in the same session takes ~1.2s, matching 1.10.10's steady-state ~1.1s for the same op at the same N. Only compiling the whole `⊖` method together on 1.12.6 is this slow. I didn't look deeper into this, also since its use-case is extremely limited for this large N.
</details>

### SectorDict U1Irrep

<details>

| operation | Julia | N=13 before→after | N=101 before→after | N=401 before→after |
|---|---|---|---|---|
| constructor | 1.10.10 | 7.38 µs → 7.02 µs (1.05x) | 56.0 µs → 45.2 µs (1.24x) | 344 µs → 203 µs (1.69x) |
| constructor | 1.12.6 | 5.29 µs → 4.27 µs (1.24x) | 37.3 µs → 27.3 µs (1.37x) | 199 µs → 112 µs (1.78x) |
| dim | 1.10.10 | 279 ns → 21.6 ns (12.9x) | 12.7 µs → 45.1 ns (282x) | 170 µs → 87.5 ns (1941x) |
| dim | 1.12.6 | 129 ns → 12.3 ns (10.5x) | 6.65 µs → 74.1 ns (89.8x) | 75.3 µs → 222 ns (340x) |
| ⊕ | 1.10.10 | 5.67 µs → 872 ns (6.50x) | 91.0 µs → 4.98 µs (18.3x) | 967 µs → 21.8 µs (44.4x) |
| ⊕ | 1.12.6 | 2.43 µs → 201 ns (12.1x) | 46.8 µs → 1.01 µs (46.1x) | 435 µs → 3.71 µs (117x) |
| ⊖ | 1.10.10 | 2.83 µs → 1.52 µs (1.86x) | 66.4 µs → 11.3 µs (5.88x) | 774 µs → 54.7 µs (14.1x) |
| ⊖ | 1.12.6 | 1.53 µs → 566 ns (2.71x) | 29.9 µs → 4.77 µs (6.27x) | 375 µs → 21.8 µs (17.2x) |
| infimum | 1.10.10 | 6.25 µs → 866 ns (7.22x) | 80.1 µs → 4.51 µs (17.7x) | 816 µs → 22.3 µs (36.6x) |
| infimum | 1.12.6 | 4.99 µs → 206 ns (24.2x) | 38.4 µs → 963 ns (39.9x) | 381 µs → 4.49 µs (84.8x) |
| supremum | 1.10.10 | 4.86 µs → 856 ns (5.68x) | 71.5 µs → 4.47 µs (16.0x) | 767 µs → 22.2 µs (34.6x) |
| supremum | 1.12.6 | 3.06 µs → 198 ns (15.5x) | 32.4 µs → 997 ns (32.5x) | 344 µs → 3.62 µs (94.9x) |
| fuse | 1.10.10 | 20.9 µs → 13.2 µs (1.58x) | 6.50 ms → 694 µs (9.36x) | 335 ms → 11.1 ms (30.3x) |
| fuse | 1.12.6 | 16.1 µs → 5.87 µs (2.74x) | 3.65 ms → 333 µs (11.0x) | 171 ms → 5.62 ms (30.5x) |
| truncate_space | 1.10.10 | 2.44 µs → 2.31 µs (1.06x) | 36.4 µs → 17.8 µs (2.05x) | 441 µs → 108 µs (4.07x) |
| truncate_space | 1.12.6 | 1.07 µs → 890 ns (1.21x) | 18.3 µs → 7.68 µs (2.39x) | 206 µs → 35.6 µs (5.80x) |

</details>

(Negative) conclusions/remarks from the benchmarks:
- For very small `N` the constructor is still somewhat slower (0.45x-0.99x), but the alternative, per the first table, is to make large `N` impossible to compile in reasonable time.
- `truncate_space` is still slightly slower at `N=2` (0.29x/0.86x) but wins from `N=8` up.
- Why `NTuple` `fuse` scales the way it does with `N` I have no clue, but it's at least better for reasonable ranges.
- `⊖` on 1.12.6 is a one-time compile bottleneck, but afterwards does fine (see note above).
- `SectorDict` wins scale with sector count.

All in all, these are improvements, notably the `SectorDict`. So it makes you wonder if there's in fact some cutoff `N` above which you want to say `SizeUnknown` to the `SectorValues`' length 🤔


src/auxiliary/dicts.jl
```diff
@@ -89,14 +89,14 @@ end
...
-# _searchsortedfirst(v::Vector, k) = searchsortedfirst(v, k)
-function _searchsortedfirst(v::Vector, k)
-    i = 1
-    @inbounds while i <= length(v) && isless(v[i], k)
-        i += 1
-    end
-    return i
-end
+_searchsortedfirst(v::Vector, k) = searchsortedfirst(v, k)
+# function _searchsortedfirst(v::Vector, k)
```
kshyatt: Can this be gotten rid of?

src/factorizations/truncation.jl
```diff
@@ -37,6 +37,31 @@ _blocklength(ax::Base.OneTo, ind::AbstractVector{Bool}) = count(ind)
...
     return spacetype(V)(c => _blocklength(dim(V, c), ind) for (c, ind) in pairs(inds))
 end
+function truncate_space(V::GradedSpace{I, NTuple{N, Int}}, inds) where {I <: Sector, N}
+    vals = values(I)
+    dualV = isdual(V)
+    newdims = zeros(Int, N)
+    for (c, ind) in pairs(inds)
+        n_read = findindex(vals, dualV ? dual(c) : c) # dual-adjusted index for reading V.dims
+        n_write = findindex(vals, c) # output is never dual, so c is fine as-is
+        newdims[n_write] = _blocklength(V.dims[n_read], ind) # dim(c) = dim(dual(c))
```
lkdvos:
```suggestion
        d = dim(V, c)
        n_write = findindex(vals, c) # output is never dual, so c is fine as-is
        newdims[n_write] = _blocklength(d, ind) # dim(c) = dim(dual(c))
```

I don't think there's a benefit for not using the generic codepath for `n_read` here.

I also am slightly confused by the comments: I know that we are mostly using this function for the factorizations in which case the space is non-dual, but then I don't know why the input could be dual? I also don't really understnd why `dim(c) = dim(dual(c))` is relevant here.

src/spaces/gradedspace.jl
```diff
@@ -126,54 +137,101 @@ function unitspace(S::Type{<:GradedSpace{I}}) where {I <: Sector}
...
 zerospace(S::Type{<:GradedSpace}) = S()
 
-# TODO: the following methods can probably be implemented more efficiently for
-# `FiniteGradedSpace`, but we don't expect them to be used often in hot loops, so
-# these generic definitions (which are still quite efficient) are good for now.
-function ⊕(V₁::GradedSpace{I}, V₂::GradedSpace{I}) where {I <: Sector}
+function ⊕(V₁::GradedSpace{I, <:SectorDict}, V₂::GradedSpace{I, <:SectorDict}) where {I <: Sector}
+    dual1 = isdual(V₁)
+    dual1 == isdual(V₂) || throw(SpaceMismatch("Direct sum of a vector space and a dual space does not exist"))
+    ks, vs = _sortedmerge(V₁.dims.keys, V₁.dims.values, V₂.dims.keys, V₂.dims.values, +, identity, identity)
```
lkdvos: Can we route this through `Base.mergewith(+, V₁.dims, V₂.dims)`, and then implement a specialized version of that for the sorted vector dict? I do like the idea of keeping this a bit closer to the `AbstractDict` functionality, such that if we ever end up switching out the dictionary type this is not too much work.

src/spaces/gradedspace.jl
```diff
@@ -126,54 +137,101 @@ function unitspace(S::Type{<:GradedSpace{I}}) where {I <: Sector}
...
+function ⊖(V::GradedSpace{I, <:SectorDict}, W::GradedSpace{I, <:SectorDict}) where {I <: Sector}
+    dualV = isdual(V)
+    V ≿ W && dualV == isdual(W) || throw(SpaceMismatch("$(W) is not a subspace of $(V)"))
+    ks, vs = _sortedmerge(V.dims.keys, V.dims.values, W.dims.keys, W.dims.values, -, identity, nothing)
+    return typeof(V)(SectorDict{I, Int}(ks, vs), dualV)
+end
+
+function fuse(V₁::GradedSpace{I, <:SectorDict}, V₂::GradedSpace{I, <:SectorDict}) where {I <: Sector}
+    dual1, dual2 = isdual(V₁), isdual(V₂)
+    acc = Dict{I, Int}() # SectorDict `get` within the double for loop accumulates O(N^2) ` findindex` calls -> sort afterwards
```
lkdvos: Is the main purpose of this specialization bypassing the bad scaling of repeated insertion into a `SectorDict` being slow, or is the point to avoid calling `dim(V, c)`, or is it the combination?

For the latter and readability, it might actually make sense to introduce a new function, similar to `Base.pairs` that produces something that iterates like "`zip(sectors(V), dims(V))`" without having to rehash a bunch of things. (Which looks like it might be useful in a bunch of these functions).

For the former, do you think we could get away with a specialized `SectorDict(generator)` constructor that does precisely this, or alternatively try out the approach where we just accumulate everything into vectors, and then "sort+merge"?

src/spaces/gradedspace.jl
```diff
@@ -126,54 +137,101 @@ function unitspace(S::Type{<:GradedSpace{I}}) where {I <: Sector}
...
-            dims[c] = get(dims, c, 0) + Nsymbol(a, b, c) * dim(V₁, a) * dim(V₂, b)
+    ks0 = collect(keys(acc))
+    vs0 = collect(values(acc))
+    perm = sortperm(ks0)
+    return typeof(V₁)(SectorDict{I, Int}(ks0[perm], vs0[perm]), false)
+end
+function fuse(V₁::GradedSpace{I, NTuple{N, Int}}, V₂::GradedSpace{I, NTuple{N, Int}}) where {I <: Sector, N}
+    vals = values(I)
+    dual1, dual2 = isdual(V₁), isdual(V₂)
+    newdims = zeros(Int, N)
```
lkdvos, line 223: I do think that probably here we want to restrict to tuple as well, and use `Base.setindex` instead. (worst case scenario, there is also this: https://github.com/Jutho/TupleTools.jl/blob/a57ef2a1189604bba10ac08b2e2bae02d29d1a3e/src/TupleTools.jl#L37-L43

src/auxiliary/dicts.jl
lkdvos: This is definitely a non-trivial change here though that might actually require a bit of benchmarking in various cases, as the question is basically "do we use linear or binary search".
I seem to recall @Jutho saying that he measured this and found that the linear search is overall faster, although I can definitely see how this has to depend on the length of the vector.

I do however agree with the point that it could be great to just get rid of a bunch of code in general, and if we find something that actually maintains a competitive dictionary type it would be wonderful if we could outsource this in its entirety. I can understand that for standard MPS it is probably more relevant to focus on the small length cases, while for tensors with more legs or symmetries with more sectors it probably ends up flipping, and at some point there might actually be a case for just switching to either `Dict` or `Dictionary`, deleting this code here. Unfortunately it is really hard to measure the total effect this will have since there is soo many different usecases and regimes 😢

lkdvos (commented):
Thanks for starting to have a look at this!

I am trying to go over this in slightly more detail, but let me put some global remarks/thoughts I have here:

The first thing relates to the `NTuple` case for large `N`. I absolutely agree that the compile time of that is untenable, and should be addressed, and while I would probably do something similar to you, there is something that looks a bit off to first constructing a vector, only to then immediately convert it to a tuple. I would definitely prefer to keep the small `N` values non-allocating and fully specialized, if at all possible, since otherwise this is a lot of extra code to maintain and we might also consider just keeping everything `SectorDict` and optimizing that.

In some sense, if we are already allocating a vector, we might as well just store the result as a vector and see if the extra pointer indirections actually matter, which I'm not expecting them to do.
So basically, I think it might be useful to just have a `GradedSpace{I, Vector{Int}}` implementation that has the exact same semantics as the tuple version, but avoids the compile-time issues.

```julia
@assume_effects :foldable function sectorstoragetype(::Type{I}) where {I <: Sector}
    if Base.IteratorSize(values(I)) isa Union{HasLength, HasShape}
        N = length(values(I))
        return N <= 10 ? NTuple{N, Int} : Vector{Int}
    else
        return SectorDict{I, Int}
    end
end

Base.getindex(::SpaceTable, I::Type{<:Sector}) = GradedSpace{I, sectorstoragetype(I)}
```

A different thing that could be relevant is that in principle we could also reduce the compile time issues for the smaller `N` values by "blocking" the compiled types, e.g. `NTuple{N, Int}` for the next value in the set `N = 1,2,4,8,16,32` which basically trades some compilation for storage efficiency (see e.g. the packages `SmallCollections.jl` or similar. I do however think that this probably hints at the `Vector{Int}` approach being more appropriate anyways.

A final comment is that I think one of the inefficiencies of the implementations e.g. for truncated factorizations is probably that we are always constructing these as `SectorDict`s, which is a bit wasteful in the case of `NTuple` storage. There might be additional optimizations in that realm as well.

src/auxiliary/dicts.jl
borisdevos: I admit I wasn't very clear in my benchmarks, and I don't have tests literally showing the benefit to switching back to binary search, but it was noticeable. I'll see to showing this more transparently. I'm predicting a crossover at fairly small number of keys though where linear search starts better, but then binary takes over, but I believe even with the smallest case I tested of 13, binary already outperformed.

At some point I was indeed looking at starting from `Dict` and then converting to `SectorDict`, which ended up working in e.g. `fuse`. For things like this, I did decide based on global improvement, so no cutoff where something else started working better. Indeed, at some point I was losing my mind about all these cutoffs, so I just stuck to what always worked best, not necessarily the best case depending on the scenario.

Jutho: If we any way are going to restrict tuples to very short size, and use vectors otherwise, can we choose that cutoff value to coincide with switching from linear search (for tuples) to binary search (for vectors)?

src/factorizations/truncation.jl
```diff
@@ -37,6 +37,31 @@ _blocklength(ax::Base.OneTo, ind::AbstractVector{Bool}) = count(ind)
...
+        d = get(V.dims, dualV ? dual(c) : c, 0)
+        len = _blocklength(d, ind)
+        if !iszero(len)
+            push!(ks, c)
+            push!(vs, len)
+        end
+    end
+    perm = sortperm(ks)
+    return typeof(V)(SectorDict{I, Int}(ks[perm], vs[perm]), false)
+end
```
Jutho, line 67: Is all of this code purely for more performance? The original `truncate_space` (L37-39) works with general `GradedSpace` objects, right?

src/spaces/gradedspace.jl
```diff
@@ -30,17 +30,17 @@ end
...
+        dc < 0 && throw(ArgumentError(lazy"Sector $k has negative dimension $dc"))
         isset[i] && throw(ArgumentError(lazy"Sector $c appears multiple times"))
-        isset = TupleTools.setindex(isset, true, i)
-        d = TupleTools.setindex(d, dc, i)
+        isset[i] = true
+        d[i] = dc
     end
-    return GradedSpace{I, NTuple{N, Int}}(d, dual)
+    return GradedSpace{I, NTuple{N, Int}}(NTuple{N, Int}(d), dual)
 end
```
Jutho, line 56: Do we really want to allocate here, if we are anyway going to restrict to short tuples?

src/spaces/gradedspace.jl
```diff
@@ -89,9 +89,20 @@ GradedSpace(g::AbstractDict; dual::Bool = false) = GradedSpace(g...; dual = dual
...
+end
+function dim(V::GradedSpace{I, NTuple{N, Int}}) where {I <: Sector, N}
+    init = zero(dimscalartype(I))
+    D = init
+    vals = values(I)
+    @inbounds for n in 1:N
+        d = V.dims[n]
+        iszero(d) && continue
+        D += dim(vals[n]) * d # dim(c) = dim(dual(c))
+    end
```
Jutho: Is the `iszero` check really helping to speed up things? By the time you checked `iszero(d)`, you've probably done the multiplication and addition?

Also, how about


```suggestion
    @inbounds for (c, d) in zip(values(I), V.dims)
        D += dim(c) * d
    end
```

src/auxiliary/dicts.jl
lkdvos: I'd also like to see the performance compared to a regular `Dict` with hashing to be honest, it seems to me that for more complicated product types like the ones @borisdevos is using the actual operations are quite a bit more involved than simply getting a hash function

borisdevos: At least within `SectorDict`s, the crossover from linear to binary outperforming is somewhere in the 16-32 key range, which is not too unexpected complexity-wise. Functions that care about this do equally good within 5% for small N, where for N = 16 they're about equal, and for N>16 get steadily outperformed by binary search. So I think at least within the context of the changes of this PR, binary search is better.

> If we any way are going to restrict tuples to very short size, and use vectors otherwise, can we choose that cutoff value to coincide with switching from linear search (for tuples) to binary search (for vectors)?

I'm a little confused though, because we're never really "densely" searching through `NTuple`s, and wouldn't also through `Vector`s, no? Every lookup within this is done with TensorKitSectors' `findindex`, which specialises for every sector type to be instant (the fallback is indeed linear, but the docs label this method as "required"). So this cutoff would be dependent on other things, like allocation and compile time, and other factors I have yet to discover.

src/factorizations/truncation.jl
borisdevos: I honestly just didn't know/realise the input was always non-dual 🫠

src/factorizations/truncation.jl
borisdevos: Yes, admittedly quite modest (1-3x speedup) and N-dependent, but it's never a regression.

src/factorizations/truncation.jl
lkdvos: Well, to be completely honest me neither, but I deduced this from the implementation which would have been wrong if it is 😆 might be reasonable to add an `@assert` for something like that though, both for code reading and accidental use

borisdevos:
Short summary of the recent commits:
- addressed the code review on `truncate_space` non-dual spaces and `dim` checking a zero contribution
- `blockdims` is the thing that iterates as `zip(sectors(V), dims(V))` without rehashing
- `Base.mergewith` for the sorted merge, but doesn't quite work for the intersect case of `infimum`
- `sectorstoragetype` defined as the name suggests, with a cutoff now above which `SectorDict` is used even for finite sector types.
- Truncation code is now also somewhat rewritten to not always work with `SectorDict`s. I'll be honest, it took me a while to get this working (it's unfamiliar code to me), and the gain is like 2x-3x on my machine, so I don't know if it's worth the maintenance of the code I wrote there. I'm happy to revert this.

Some comments of the review I didn't immediately address, and why:
- The `NTuple` graded space constructor and `NTuple` `fuse` still allocate a vector, and then convert to an `NTuple` at the end. I kept this because the gain in this on my machine is significant, namely factor of the order 20-60 within the range of the `NTuple` cutoff. So I thought that's merit enough to actually not restrict to a tuple there.
- [Julia internals](https://github.com/JuliaLang/julia/blob/15346901f0039751c5488744f1f62de7d87510a8/base/tuple.jl#L363) make tuples up till size 32 very efficient, and then there's a notable drop in performance, so the cutoff is put there. This means that product sectors could quickly enter this regime. For that reason I also kept the binary search, as `N` can get big now for `SectorDict`s.

There was a suggestion to run some DMRG code with these changes which I haven't done yet. I believe since that will be dominated by LAPACK, the only thing I'd have to look for is no notable regression. So if things remain in the same ballpark (or just compile at all, since that was an issue), then I think things are fine.

src/auxiliary/dicts.jl
```diff
@@ -89,14 +89,14 @@ end
...
+#     i = 1
+#     @inbounds while i <= length(v) && isless(v[i], k)
+#         i += 1
+#     end
+#     return i
+# end
```
lkdvos: Probably fair to just delete this and replace the callsites of `_searchsortedfirst` directly.

```suggestion
```

src/spaces/gradedspace.jl
```diff
@@ -89,9 +89,18 @@ GradedSpace(g::AbstractDict; dual::Bool = false) = GradedSpace(g...; dual = dual
 field(::Type{<:GradedSpace}) = ℂ
 InnerProductStyle(::Type{<:GradedSpace}) = EuclideanInnerProduct()
 
-function dim(V::GradedSpace)
-    init = 0 * dim(first(allunits(sectortype(V))))
-    return sum(c -> dim(c) * dim(V, c), sectors(V); init = init)
+function dim(V::GradedSpace{I, <:AbstractDict}) where {I <: Sector}
+    init = zero(dimscalartype(I))
+    return sum(((c, d),) -> dim(c) * d, V.dims; init)
```
lkdvos:
```suggestion
    return sum(((c, d),) -> dim(c) * d, blockdims(V); init)
```

I think this should now also be the correct implementation for both storagetypes, so I think you can remove the other function

src/factorizations/truncation.jl
```diff
@@ -34,32 +34,101 @@ _blocklength(ax, ind) = length(ax[ind])
...
+# fallbacks to catch SectorVector/SectorDict, even for NTuple sectorstoragetype
+_denseget(v, ::Type{I}, c::I) where {I <: Sector} = get(v, c, nothing)
+_densekeys(v, ::Type{I}) where {I <: Sector} = keys(v)
+_densepairs(v, ::Type{I}) where {I <: Sector} = pairs(v)
+
+# builds either a dense Vector or SectorDict based on sectorstoragetype
+# mapping each (c, v) pair's sector c to f(c, v)
+# so every `findtruncated` method shares one output-construction path
+# `pairsiter` are c => v pairs, can be c => nothing for NoTruncation/TruncationIntersection/TruncationUnion
+function _builddensemap(f, ::Type{D}, ::Type{I}, pairsiter, ::Type{V}) where {D <: Tuple, I <: Sector, V}
```
lkdvos: I think previously this was all just going through the `SectorDict` implementations always, because the tuple approach only works for `Int`, is there any indication we need to speed this up by using this vector-based set? If we don't have to do this, that simplifies quite a bit of the code, and makes the `Vector{Int}` implementation simpler as well

borisdevos: I just chose vector storage here because I knew it was better in the tuple case to work with vectors as long as possible. I had all the freedom to make this also dictionary-based, which I put as a todo somewhere. At the end of the day, I added all this mapping machinery this way because I first had if-elses everywhere based on storage type, which looked ugly, so I just came up with something that would dispatch nicely.

Honestly, I think the main speedups that happen here aren't immediately due to this mapping as a vector, but just that `truncate_space` itself is storage type-specialised, and where possible there's type stability.

lkdvos:
pushed some small changes to the sortmerge implementation, removing some branches and duplicate reads/writes, on my machine it measures around 30% faster and it's slightly more compact.
I've restricted the signatures for the specialized implementation to `Integer`, mostly because the `iszero(d)` guard for writing does actually assume that.

For the other comments, I think I didn't really explain properly what I meant. For the truncated spaces, the logic was **always** constructing a `SortedVectorDict`, and never using the `Tuple`, because the tuple approach never had a dictionary interface and was somewhat integer-exclusive (again for the same reason: you need to have 0 as the unoccupied value for that, which is reflecting in the need for the `Union{Nothing, V}` in your vector approach). My main question is, here you have changed this to now switch to `SortedVectorDict` or `Vector{Union{Nothing, V}}` based on the sectortype, and I wonder if this is required, as it does introduce quite a bit of extra code. If it is, I would definitely argue that we should introduce a `FullVectorDict <: AbstractDict` or something similar to encapsulate that, as the `_denseget` etc code looks like it is trying to reinvent most of that interface. (I'm not asking about the `Vector` vs `Tuple` use here).

borisdevos:
> My main question is, here you have changed this to now switch to `SortedVectorDict` or `Vector{Union{Nothing, V}}` based on the sectortype, and I wonder if this is required, as it does introduce quite a bit of extra code. If it is, I would definitely argue that we should introduce a `FullVectorDict <: AbstractDict` or something similar to encapsulate that, as the `_denseget` etc code looks like it is trying to reinvent most of that interface. (I'm not asking about the `Vector` vs `Tuple` use here).

Ignoring `Vector` vs `Tuple`, this code indeed basically mimics some special dictionary purely so that the functions dispatch cleanly based on storage type. But I agree, these index maps can all be homogenised into some different dictionary type. I'm not sure how much less code that would be though.

lkdvos:
> I'm not sure how much less code that would be though.

Isn't this most of the changes to the truncation.jl file, which is a substantial amount of code?

lkdvos (approved):
I've pushed some small changes for things I still wanted to clean up slightly but couldn't really explain since I had to try around a bit myself, feel free to still disagree with some of these. I think all other comments have been resolved, so good to go for me. @Jutho, do you want to have another look, or shall we just merge?

Jutho:
I'll also take a look tonight.

src/auxiliary/dicts.jl
```diff
@@ -186,6 +177,170 @@ function Base.:(==)(d1::SortedVectorDict, d2::SortedVectorDict)
...
+    end
+    resize!(ks, n)
+    resize!(vs, n)
+    return SortedVectorDict{K, V}(ks, vs)
+end
+# write into slot `n + 1` and only advance the length when the value is nonzero
+@inline function _mergestore!(ks, vs, n, k, d)
+    @inbounds ks[n + 1] = k
+    @inbounds vs[n + 1] = d
+    return n + !iszero(d)
```
Jutho, line 227: Is there a specific reason to write when `iszero(d)`? Why not

```suggestion
    if !iszero(d)
        @inbounds ks[n + 1] = k
        @inbounds vs[n + 1] = d
        return n+1
    else
        return n
    end
```

Is this a performance thing?

src/spaces/gradedspace.jl
```diff
@@ -280,13 +373,24 @@ specify `D`.
 const Vect = SpaceTable()
 Base.getindex(::SpaceTable) = ComplexSpace
 Base.getindex(::SpaceTable, ::Type{Trivial}) = ComplexSpace
-function Base.getindex(::SpaceTable, I::Type{<:Sector})
+Base.getindex(::SpaceTable, I::Type{<:Sector}) = GradedSpace{I, sectorstoragetype(I)}
+
+# based on Julia tuple unrolling range
+const _ntuple_storage_threshold = 32
```
Jutho, line 379: Are we using the style for such package constants to have an all caps name?

src/spaces/gradedspace.jl
```diff
@@ -139,54 +152,116 @@ function unitspace(S::Type{<:GradedSpace{I}}) where {I <: Sector}
...
+            n = _mergestore!(ks, vs, n, a, v1[i])
+            i += 1
+        elseif isless(b, a)
+            _throw_not_subspace(V, W) # sector of `W` absent from `V`
+        else
+            v1[i] < v2[j] && _throw_not_subspace(V, W)
+            n = _mergestore!(ks, vs, n, a, v1[i] - v2[j])
+            i += 1
+            j += 1
+        end
```
Jutho, line 195: This seems to be just replicating the code of `mergewith -> _sortedmerge` for `SortedVectorDict`. Why is this not simply calling `mergewith` using `combine(v1,v2) = v1 < v2 ? _throw_not_subspace(V, W) : v1 - v2` ?

src/auxiliary/dicts.jl
lkdvos: Yes, this is branchless and on smaller cases about 10% faster

src/spaces/gradedspace.jl
lkdvos: Mergewith only calls the combiner on elements that are in both sets, so doing this naively would just add the sectors of the second set that aren't in the first. I tried a bit to migrate the two implementations but couldn't really get this to work all that cleanly...

src/auxiliary/dicts.jl
Jutho: Ok. I am a bit surprised, since this is not called from a loop that is itself full of branches with the `isless` structure, but fair enough. And it is even shorter codewise, so no issues there 😄 .

Jutho:
My apologies for the slow review. I find the current code quite difficult to properly review, as I am not entirely convinced of the structure.

I do like the modifications to `GradedSpace` with a cutoff based on length, which could be even smaller for all I am concerned (e.g. 4 or 8 instead of 32). Also the use of `mergewith` as a paradigm, and its specialized implementation for `SortedVectorDict`, is very clever.

However, I am less convinced by the `FullVectorDict` implementation in auxiliary.jl and the associated `sectormap` paradigm. As far as I can tell, this seems to be used only to support the truncation. One issue is that I view `auxiliary.jl` as a place for some very basic auxiliary tools that do not depend on the rest of the package, but are used at various places throughout the package. I.e., it would be code that could be moved to its own package if there were a more general use for it. This seems somewhat orthogonal to the `FullVectorDict` and `sectormap` paradigm, which are only used in `truncation.jl` and which depend strongly on methods from the `Sector` interface, (even though `FullVectorDict` looks like a generic `AbstractDict` subtype, aside from the restriction on the key type parameter `K`). 

I still have to read through truncation.jl to see if there is anything better I could come up with, which is what I will do next. The reason for bringing this up and being difficult is that I want to ensure that the code remains logically structured and therefore easier to maintain.

Jutho:
In particular, is all the code of `FullVectorDict` worth the performance gain instead of just using `SectorDict = SortedVectorDict` for all sectors when applying `sectormap`, irrespective of whether the `GradedSpace` uses `NTuple` or not?

src/factorizations/truncation.jl
```diff
@@ -35,13 +35,41 @@ _blocklength(ax::Base.OneTo, ind::AbstractVector{<:Integer}) = length(ind)
 _blocklength(ax::Base.OneTo, ind::AbstractVector{Bool}) = count(ind)
 
 function truncate_space(V::ElementarySpace, inds)
+    @assert !isdual(V)
+    I = sectortype(V)
+    @assert I == Trivial
```
Jutho, line 40: Is this just to assert that the specialized `GradedSpace` implementations below are used, so that only `ComplexSpace` (or `CartesianSpace`) goes through here?

src/factorizations/truncation.jl
```diff
@@ -35,13 +35,41 @@ _blocklength(ax::Base.OneTo, ind::AbstractVector{<:Integer}) = length(ind)
...
+    return typeof(V)(SectorDict{I, Int}(ks[perm], vs[perm]), false)
+end
 
 function truncate_domain!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap, inds)
     for (c, b) in blocks(tdst)
         I = get(inds, c, nothing)
-        @assert !isnothing(I)
+        @assert !isnothing(I) # kept for safety, but should be guaranteed by pairs(inds) covering every block sector
         b′ = block(tsrc, c)
         b .= view(b′, :, I)
```
Jutho, line 74: This is unrelated to this PR, but can we also call `I -> ind` here, which is more in line with the naming elsewhere. `I` is typically also used for sectortype in this package.

I would make it as a Github suggestion, but the button doesn't seem to work on multiline selection today (and if I force it manually, it typically ends up not being mergeable anyway).

src/factorizations/truncation.jl
```diff
@@ -50,7 +78,7 @@ end
 function truncate_codomain!(tdst::AbstractTensorMap, tsrc::AbstractTensorMap, inds)
     for (c, b) in blocks(tdst)
         I = get(inds, c, nothing)
-        @assert !isnothing(I)
+        @assert !isnothing(I) # kept for safety, but should be guaranteed by pairs(inds) covering every block sector
         b′ = block(tsrc, c)
         b .= view(b′, I, :)
```
Jutho, line 83: Same comment here.

src/factorizations/truncation.jl
```diff
@@ -59,7 +87,7 @@ end
 function truncate_diagonal!(Ddst::DiagonalTensorMap, Dsrc::DiagonalTensorMap, inds)
     for (c, b) in blocks(Ddst)
         I = get(inds, c, nothing)
-        @assert !isnothing(I)
+        @assert !isnothing(I) # kept for safety, but should be guaranteed by pairs(inds) covering every block sector
         diagview(b) .= view(diagview(block(Dsrc, c)), I)
```
Jutho, line 91: Same comment

src/factorizations/truncation.jl
```diff
@@ -114,7 +142,9 @@ end
 function MAK.truncate(
         ::typeof(left_null!), (U, S)::NTuple{2, AbstractTensorMap}, strategy::NoTruncation
     )
-    ind = SectorDict(c => (size(b, 2) + 1):size(b, 1) for (c, b) in blocks(S))
+    ind = sectormap(blocks(S)) do _, b
+        (size(b, 2) + 1):size(b, 1)
+    end
     V_truncated = truncate_space(space(S, 1), ind)
     Ũ = similar(U, codomain(U) ← V_truncated)
     truncate_domain!(Ũ, U, ind)
```
Jutho, line 150: It seems the old code was not very consistent with naming variables
```suggestion
    inds = sectormap(blocks(S)) do _, b
        (size(b, 2) + 1):size(b, 1)
    end
    V_truncated = truncate_space(space(S, 1), inds)
    Ũ = similar(U, codomain(U) ← V_truncated)
    truncate_domain!(Ũ, U, inds)
```

src/factorizations/truncation.jl
```diff
@@ -123,7 +153,9 @@ end
 function MAK.truncate(
         ::typeof(right_null!), (S, Vᴴ)::NTuple{2, AbstractTensorMap}, strategy::NoTruncation
     )
-    ind = SectorDict(c => (size(b, 1) + 1):size(b, 2) for (c, b) in blocks(S))
+    ind = sectormap(blocks(S)) do _, b
+        (size(b, 1) + 1):size(b, 2)
+    end
     V_truncated = truncate_space(dual(space(S, 2)), ind)
     Ṽᴴ = similar(Vᴴ, V_truncated ← domain(Vᴴ))
     truncate_codomain!(Ṽᴴ, Vᴴ, ind)
```
Jutho, line 161:
```suggestion
    inds = sectormap(blocks(S)) do _, b
        (size(b, 1) + 1):size(b, 2)
    end
    V_truncated = truncate_space(dual(space(S, 2)), inds)
    Ṽᴴ = similar(Vᴴ, V_truncated ← domain(Vᴴ))
    truncate_codomain!(Ṽᴴ, Vᴴ, inds)
```

src/factorizations/truncation.jl
```diff
@@ -206,18 +238,24 @@ MAK.findtruncated_svd(values::SectorVector, strategy::TruncationByOrder) =
     MAK.findtruncated(values, strategy)
 
 function MAK.findtruncated(values::SectorVector, strategy::TruncationByFilter)
-    return SectorDict(c => findall(strategy.filter, d) for (c, d) in pairs(values))
+    return sectormap(pairs(values)) do _, v
+        findall(strategy.filter, v)
+    end
```
Jutho, line 243: Are you sure that any of those are actually faster? They all contain the call `pairs(values)`, which has the following definition:

```julia
Base.pairs(v::SectorVector) = SectorDict(c => v[c] for c in keys(v))
```

and so it anyway creates a `SectorDict`. None of this has anything to do with tuples, so I am not sure why these changes are in this PR, and not having this would allow to discard all of the `sectormap` code.

If there really is some overhead in these calls, it might be worth investigating whether we can improve the `pairs(v::SectorVector)` call, e.g. as

```julia
Base.pairs(v::SectorVector) = SectorDict((c, s) => v[s] for (c, s) in pairs(v.structure))
```

lkdvos:
I think this is ready for another round of review. I've incorporated the code style suggestions and variable names, removed the `FullVectorDict` and slightly improved the `pairs(::SectorVector)` iterator to compensate for that. It would probably be good to remeasure performance and re-evaluate after merging this PR, but since there are already quite a few useful changes here I'd vote in favor of merging this first.

Small thing I noted, the type alias `ZNSpace{N}` is no longer compatible with the split sectorstoragetype based on the sectortype alone, as this will give a wrong result as soon as `N > tuple_threshold`. I think it's fine to simply deprecate this, the only other thing I can come up with is to change this to `ZNSpace(N)(args...)`, which also looks a bit off to me.

docs/src/lib/spaces.md
```diff
@@ -37,13 +37,18 @@ Vect
...
 In this respect, there are also a number of type aliases for the `GradedSpace` types associated with the most common sectors, namely
 
 ```julia
-const ZNSpace{N} = Vect[ZNIrrep{N}]
-const Z2Space = ZNSpace{2}
-const Z3Space = ZNSpace{3}
-const Z4Space = ZNSpace{4}
+const Z2Space = Vect[ZNIrrep{2}]
+const Z3Space = Vect[ZNIrrep{3}]
+const Z4Space = Vect[ZNIrrep{4}]
```
Jutho:
```suggestion
const Z2Space = Rep[ℤ{2}]
const Z3Space = Rep[ℤ{3}]
const Z4Space = Rep[ℤ{4}]
```

docs/src/Changelog.md
```diff
@@ -26,12 +26,14 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 - For sector types with `GenericUnit` such that colorings are not unique, `GradedSpace`, `ProductSpace` and `HomSpace` now check for this compatibility. In particular, this prevents the construction of `TensorMap`s with incompatible colorings, which previously either errored or produced empty tensors inconsistently. ([#515](https://github.com/QuantumKitHub/TensorKit.jl/pull/515))
 
 ### Deprecated
+- The type alias `ZNSpace{N}` is deprecated in favour of `Vect[ZNIrrep{N}]`: a type alias cannot compute the storage type from `N`, so the two only agree for small `N`. ([#511](https://github.com/QuantumKitHub/TensorKit.jl/pull/511))
```
Jutho:
```suggestion
- The type alias `ZNSpace{N}` is deprecated in favour of `Vect[ZNIrrep{N}]` or `Rep[ℤ{N}]`: a type alias cannot compute the storage type from `N`, so the two only agree for small `N`. ([#511](https://github.com/QuantumKitHub/TensorKit.jl/pull/511))
```

src/factorizations/factorizations.jl
```diff
@@ -18,6 +19,8 @@ using LinearAlgebra: LinearAlgebra, BlasFloat, Diagonal,
 
 using TensorOperations: Index2Tuple
 
+using TupleTools: MutableNTuple, StaticLength
+
```
Jutho: Is this still used?

Jutho: Nevermind, seems it is.

src/factorizations/truncation.jl
```diff
@@ -35,32 +35,58 @@ _blocklength(ax::Base.OneTo, ind::AbstractVector{<:Integer}) = length(ind)
...
 end
+function truncate_space(V::TupleGradedSpace{I, N}, inds) where {I <: Sector, N}
+    @assert !isdual(V)
+    vals = values(I)
+    newdims = MutableNTuple(ntuple(_ -> 0, StaticLength(N)))
+    for (c, ind) in pairs(inds)
+        d = dim(V, c)
+        n_write = findindex(vals, c)
+        @inbounds newdims[n_write] = _blocklength(d, ind)
+    end
```
Jutho: Do we need this construction via `MutableNTuple`? How about


```suggestion
    newdims = ntuple(N) do n
        c = vals[n]
        d = V.dims[n]
        ind = inds[c]
        return _blocklength(d, ind)
    end
```

or thus as onliner

```suggestion
    newdims = ntuple(n->_blocklength(V.dims[n], inds[vals[n]]), N)
```

So we are then not iterating the `inds` dictionary, but since we anyway know that `N` is pretty small in this case, I do wonder how severe the key lookups are.

(We are also not needing to do `findindex(vals, c)` anymore, which might sometimes be a simple calculation, but sometimes also an inefficient lookup).

src/spaces/gradedspace.jl
```diff
@@ -12,13 +15,13 @@ and pivotal (pre-, multi-) fusion category.
...
+If the result is of type `HasLength` or `HasShape` and `N = length(values(I))` does not exceed
+`$_NTUPLE_STORAGE_THRESHOLD`, `dims` will be stored in a `NTuple{N,Int}`. This requires that a
+sector `s::I` can be transformed into an index via `s == getindex(values(I), i)` and
+`i == findindex(values(I), s)`. Otherwise, i.e. for more sectors than that or if
+`Base.IteratorSize(values(I))` results `IsInfinite()` or `SizeUnknown()`, a `SectorDict{I,Int}`
+is used to store the non-zero degeneracy dimensions with the corresponding sector as key.
+The parameter `D` is hidden from the user and should typically be of no concern.
 
 The concrete type `GradedSpace{I,D}` with correct `D` can be obtained as `Vect[I]`, or if
 `I == Irrep[G]` for some `G<:Group`, as `Rep[G]`.
```
```suggestion
The concrete type `GradedSpace{I,D}` with correct `D` can be obtained as `Vect[I]`,
or as `Rep[G]` if `I == Irrep[G]` for some `G<:Group`.
```

src/spaces/gradedspace.jl
```diff
@@ -39,18 +46,20 @@ function _check_unit_homogeneity(::Type{I}, sectors) where {I <: Sector}
 end
 
 function GradedSpace{I, NTuple{N, Int}}(dims; dual::Bool = false) where {I, N}
-    d = ntuple(n -> 0, N)
-    isset = ntuple(n -> false, N)
+    @assert N <= 64 "the `UInt64` bitmask tracking which sectors have been set holds 64 bits"
+    d = TupleTools.MutableNTuple(ntuple(Returns(0), StaticLength(N)))
```
Jutho: `StaticLength(N)` is an old TupleTools artefact from long before constant propagation, I am not sure if that is still necessary, especially in a function that has `N` as a parameter.

src/factorizations/truncation.jl
lkdvos: Good call, although it requires a little more care because not all `inds[vals[n]]` will work, but I think I can actually work around this because we always have either `V.dims[n] == 0` or the key is present

src/factorizations/factorizations.jl
Jutho: But not with my suggestion below, at least not in the factorizations submodule.

Jutho (commented):
I left some smaller final comments, but otherwise approved.

Jutho:
I have not actually approved since I hadn't checked wether automerge was on, and wanted to give you time to look at my suggestions. However, feel free to merge after having looked at them.

src/spaces/gradedspace.jl
lkdvos: removed

Issue #514, opened 2026-08-26
Inconsistent handling of empty tensors when using multifusion sectors


leburgel:
There is some inconsistency in how tensor maps without any valid fusion channels are handled when using multifusion sectors. Depending on how the spaces of such a tensor are partitioned between the domain and codomain, construction and manipulation eithers returns an empty tensor with norm zero, or throws an `ArgumentError`.

For example:

```julia
using TensorKit, TensorKitSectors

C0, C1 = IsingBimodule(1, 1, 0), IsingBimodule(1, 1, 1)
D0, D1 = IsingBimodule(2, 2, 0), IsingBimodule(2, 2, 1)
M, Mop = IsingBimodule(1, 2, 0), IsingBimodule(2, 1, 0)

V = (
    Vect[IsingBimodule](C0 => 2, C1 => 1), Vect[IsingBimodule](Mop => 1)',
    Vect[IsingBimodule](D0 => 2, D1 => 2), Vect[IsingBimodule](M => 2)',
    Vect[IsingBimodule](C0 => 3, C1 => 2),
)
t = randn(Float64, V[1] ⊗ V[2] ← V[3] ⊗ V[4] ⊗ V[5])
# t is fine: norm(t) == 0.0, blocksectors(t) == IsingBimodule[]
@show norm(t)
@show blocksectors(t)

repartition(t, 0)   # ok, norm = 0.0
repartition(t, 1)   # ok, norm = 0.0
repartition(t, 2)   # ok, norm = 0.0
repartition(t, 3)   # ArgumentError: invalid fusion channel
repartition(t, 4)   # ArgumentError: invalid fusion channel
repartition(t, 5)   # ok, norm = 0.0
```

The issue seems to be that `blocksectors` calls corresponding to an empty set sometimes return and empty set and sometimes throw, depening on the specific arguments:

```julia
codom = Vect[IsingBimodule](C0 => 1) ⊗ Vect[IsingBimodule](C0 => 1)
blocksectors(codom ← Vect[IsingBimodule](D0 => 1))   # ArgumentError: invalid fusion channel
blocksectors(codom ← Vect[IsingBimodule](C1 => 1))   # IsingBimodule[]
```

I'm not sure whether it's better to always throw or never throw. @lkdvos, @borisdevos.

borisdevos:
[...]

PR #515, opened 2026-08-26
Prevent construction of multifusion tensors with incompatible coloring

borisdevos:
Deals with #514, at least partially.
Now we have
```julia
julia> Wis
(Vect[IsingBimodule]((1, 1, 0) => 1) ⊗ Vect[IsingBimodule]((1, 1, 0) => 1)) ← Vect[IsingBimodule]((2, 2, 0) => 1)

julia> blocksectors(Wis)
0-element Dictionaries.Indices{IsingBimodule}
```
which behaves the same way as 
```julia
julia> Wu1
Rep[U₁](1 => 1) ← (Rep[U₁](0 => 1) ⊗ Rep[U₁](0 => 1))

julia> blocksectors(Wu1)
0-element Dictionaries.Indices{U1Irrep}
```

Edit: decided to actually prevent construction of these kinds of tensors at the level of the product space and hom space.
Since this was incompatible with the previous implementation of `unitspace`, this now errors for `GenericUnit` sector types.


lkdvos (commented):
I do think this is probably quite a big performance hit right in the hot path of the fusiontree constructors though, I don't think there is typically a fast implementation of this. Additionally, I was kind of expecting to convert more things to `Nsymbol`, to actually make them error if the fusion is disallowed?

src/spaces/gradedspace.jl
```diff
@@ -122,6 +122,8 @@ function flip(V::GradedSpace{I}) where {I <: Sector}
 end
 
 function unitspace(S::Type{<:GradedSpace{I}}) where {I <: Sector}
+    UnitStyle(I) isa GenericUnit &&
+        throw(ArgumentError("Cannot construct unit space for sector types with semisimple unit structure."))
     return S(unit => 1 for unit in allunits(I))
```
lkdvos, line 138: Would it also work to have this call `S(unit(I) => 1)` and depend on the `unit` function erroring, or is that error message not very helpful?

src/spaces/productspace.jl
```diff
@@ -7,7 +7,22 @@ Only tensor products between [`ElementarySpace`](@ref) objects of the same type
...
+end
+
+# check that every product of elementary spaces is compatible color-wise
+function _check_unit_compatibility(::Type{S}, spaces::NTuple{N, S}) where {N, S <: ElementarySpace}
+    UnitStyle(sectortype(S)) isa GenericUnit || return nothing
+    N == 0 && return nothing
+
+    @inbounds for i in 1:(N - 1)
+        _rightunit(spaces[i]) == _leftunit(spaces[i + 1]) ||
+            throw(ArgumentError(lazy"spaces $(i) and $(i + 1) have incompatible coloring"))
```
lkdvos: Do we want this to be `SpaceMismatch`, or should we just introduce a new error type for this? Do you also want to print which spaces are the offending ones?
This comment also holds for the `HomSpace` checks

src/spaces/productspace.jl
```diff
@@ -64,7 +79,7 @@ Base.axes(P::ProductSpace) = map(axes, P)
 Base.axes(P::ProductSpace, n::Int) = axes(P[n])
 
 dual(P::ProductSpace{<:ElementarySpace, 0}) = P
-dual(P::ProductSpace) = ProductSpace(map(dual, reverse(P)))
+dual(P::ProductSpace) = ProductSpace(map(dual, reverse(P.spaces)))
```
lkdvos:
```suggestion
dual(P::ProductSpace) = ProductSpace(reverse(map(dual, P))
```
Slight preference for not accessing the fields 😉

src/spaces/homspace.jl
```diff
@@ -9,6 +9,43 @@ to denote categories and their objects, and keep `HomSpace` distinct.
...
+    function HomSpace{S, P1, P2}(codomain::P1, domain::P2) where {S <: ElementarySpace, P1 <: CompositeSpace{S}, P2 <: CompositeSpace{S}}
+        _check_unit_compatibility(codomain, domain)
+        return new{S, P1, P2}(codomain, domain)
+    end
+end
+function HomSpace(codomain::P1, domain::P2) where {S, P1 <: CompositeSpace{S}, P2 <: CompositeSpace{S}}
+    return HomSpace{S, P1, P2}(codomain, domain)
+end
+
+function _check_unit_compatibility(codomain::ProductSpace{S}, domain::ProductSpace{S}) where {S <: ElementarySpace}
```
lkdvos: I think there are some missing edge cases that are not really handled correctly/consistently here, mostly for the case where `leftunit` or `rightunit` cannot be inferred. In principle it is allowed (?) to create 0-dimensional `ElementarySpace` objects, which in this implementation will always error even though no violation is detected. (In some sense this is because here you check for validity and error if it cant be proven, while you probably want to error if invalidity can be proven).

The only reason I can think of why this might be important is that if you have an SVD that accidentally kills all sectors, you end up with a 0-dim leg on the S which will now error on the space constructions, even though nothing really went wrong

src/spaces/vectorspaces.jl
```diff
@@ -192,22 +193,24 @@ function rightunitspace(V::ElementarySpace)
...
+
 """
     isunitspace(V::S) where {S <: ElementarySpace} -> Bool
 
 Return whether the elementary space `V` is a unit space, i.e. is isomorphic to the
 trivial one-dimensional space. For vector spaces of type `GradedSpace{I}` where `Sector` `I` has a
-semi-simple unit structure, this returns `true` if `V` is isomorphic to either the left, right or
-semi-simple unit space.
+semisimple unit structure, this returns `true` if `V` is isomorphic to either the left, right or
+semisimple unit space.
```
lkdvos: https://en.wikipedia.org/wiki/Semi-simplicity

[edit] it seems like wikipedia isn't even consistent with itself... (https://en.wikipedia.org/wiki/Semisimple_module)

lkdvos (commented):
Other than the comments, looks good, thanks for looking into this!

src/spaces/gradedspace.jl
borisdevos: This error is more informative in this context, I'd keep it this way.

src/spaces/productspace.jl
borisdevos: Changed this to spacemismatch and return the spaces in both cases

src/spaces/homspace.jl
borisdevos: You're right, I was also not dealing with length-1 product spaces correctly. My fix for the 0d ones fixed both automatically. Importantly though, I determined that when one of the spaces is 0d, there's no need to check colorings. This is based on the sectors of the total product space always being empty, regardless of valid or invalid coloring, and things like blocksectors and fusiontrees are built on this.

src/spaces/vectorspaces.jl
borisdevos: [The Bible](https://[math.mit.edu/~etingof/egnobookfinal.pdf](https://math.mit.edu/~etingof/egnobookfinal.pdf)) doesn't hyphenate it, so I'll follow suit :)

src/spaces/gradedspace.jl
```diff
@@ -122,6 +122,8 @@ function flip(V::GradedSpace{I}) where {I <: Sector}
 end
 
 function unitspace(S::Type{<:GradedSpace{I}}) where {I <: Sector}
+    UnitStyle(I) isa GenericUnit &&
+        throw(ArgumentError("Cannot construct unit space for sector types with semisimple unit structure."))
```
Jutho, line 137: Would it be useful to actually check if `length(allunits(I)) > 1` before erroring? That way simply declaring a normal fusion category as `GenericUnit` would not immediately trigger the error.

borisdevos: I think instead of this, we should add a test in the TKS testsuite checking that `GenericUnit` actually agrees with having `length(allunits(I)) > 1`, precisely to prevent normal fusion categories from having `GenericUnit`. I don't see why we would want that in the first place.

lkdvos: I feel like I don't really agree with this, since a lot of TensorKit is structured around symmetry-agnostic code I feel like as a downstream library it is good to decide to either support multifusion, or not, which restricts the available functions. In the case where you decide to make things multifusion compatible, it's definitely more convenient to have a loud error immediately. I think I also recall that checking the length is actually the default implementation of the trait to begin with?

borisdevos: The unit style is indeed based on checking the length of allunits, so this is effectively doing that but bypassing the style check, since there was an argument to be able to artificially set some custom fusion category to have this multifusion property. As of now such a custom sector type would then be consistent everywhere except in calling `unitspace`.

Where exactly should this loud error then occur for downstream packages which don't support multifusion? With the current state of this PR (but even before, in general with multifusion) `unitspace` and braids are the only functions that cannot be used.

Jutho: Ok I can see the erroring argument, so I am fine either way.

I think that I am still curious to what extent we can support mixed objects, i.e. objects having different colorings on one or two sides, and thus to not error so quickly. But I can see that it would be much more subtle and tedious to work out all the possible ways where things can go wrong, and if this is not necessary anyway, it is probably not worth the effort.

lkdvos: reverted this for now, and we can revisit the mixed case later.

lkdvos (approved):

src/spaces/vectorspaces.jl
```diff
@@ -192,22 +192,45 @@ function rightunitspace(V::ElementarySpace)
...
 
+function _rightunit(V::ElementarySpace)
+    u = _units(V)
+    isnothing(u) && throw(ArgumentError("Cannot determine the right unit of an empty space"))
+    return _rightunitof(u)
+end
+
+# Return the `(leftunit, rightunit)` shared by all sectors of `V`, or `nothing` for the zero
+# space, whose coloring is unconstrained and thus acts as a wildcard in unit checks.
+function _units(V::ElementarySpace)
```
Jutho: I would be more explicit and call this function `_leftrightunit`

Jutho: Also, since we are now checking a lot, I think the question is whether this check should already appear in the `GradedSpace` constructors. Currently, I can still construct something like `GradedSpace(c=>1 for c in allunits(I))`, but as soon as I take tensor products with this space, I get errors? That does not sound like the correct behaviour.

src/spaces/vectorspaces.jl
```diff
@@ -192,22 +192,45 @@ function rightunitspace(V::ElementarySpace)
...
+    l, r = leftunit(first(s)), rightunit(first(s))
+    all(c -> leftunit(c) == l && rightunit(c) == r, s) ||
+        throw(SpaceMismatch(lazy"sectors of $V do not share a single left and right unit"))
+    return (l, r)
+end
+
+_leftunitof(u::Tuple{I, I}) where {I <: Sector} = u[1]
+_rightunitof(u::Tuple{I, I}) where {I <: Sector} = u[2]
+_leftunitof(::Nothing) = nothing
+_rightunitof(::Nothing) = nothing
```
Jutho: I find the current code structure quite confusing and verbose. There is `_leftunitof`, which always needs to be combined with `_units` in order to deal with the nothing case, and there is `_leftunit` which explicitly checks against the `nothing` case. Some call sites then use `_leftunit` whereas others use the `_leftunitof(_units(...))` paradigm.

Can't we just have a `_leftunit` and `_rightunit` that returns the respective unit of a space, or `nothing` if the space is empty. And then check explicitly against `nothing` in those call sites where we don't want to allow this. I don't know if `_units/_leftrightunit` should still exist, but if it does, it should simply return `(nothing, nothing)` in the empty case.

src/spaces/vectorspaces.jl
lkdvos: implemented. This was slightly annoying since our tests were not fully compatible with that, but I think I managed to bypass this.

src/spaces/vectorspaces.jl
lkdvos: I gave it another go, hopefully now a bit more transparent?

Jutho: Thanks, I'll take a look. I also didn't mind giving this a go, but it's only now that I got some coding time.

src/spaces/gradedspace.jl
```diff
@@ -29,6 +29,15 @@ struct GradedSpace{I <: Sector, D} <: ElementarySpace
 end
 sectortype(::Type{<:GradedSpace{I}}) where {I <: Sector} = I
 
+# elementary spaces are homogeneously colored: all sectors share a left and a right unit.
+function _check_unit_homogeneity(::Type{I}, sectors) where {I <: Sector}
+    (!(UnitStyle(I) isa GenericUnit) || isempty(sectors)) && return nothing
```
```suggestion
    (UnitStyle(I) isa SimpleUnit || isempty(sectors)) && return nothing
```

src/spaces/vectorspaces.jl
borisdevos: Just to motivate why I didn't do it at the `GradedSpace` level: I treat `GradedSpace` as some semi-simple object of a sector type. Only from `ProductSpace` onwards is it really seen as "a leg of a tensor". But it's fine as-is

src/spaces/vectorspaces.jl
```diff
@@ -134,9 +134,8 @@ Always returns `false` for spaces where `V == conj(V)`, i.e. vector spaces over
 
 Return the corresponding vector space of type `S` that represents the trivial
 one-dimensional space, i.e. the space that is isomorphic to the corresponding field.
-For vector spaces where `I = sectortype(S)` has a semi-simple unit structure
-(`UnitStyle(I) == GenericUnit()`), this returns a multi-dimensional space corresponding to all unit sectors:
-`dim(unitspace(V), s) == 1` for all `s in allunits(I)`. 
+For vector spaces where `I = sectortype(S)` has a semisimple unit structure
+(`UnitStyle(I) == GenericUnit()`), this errors.
```
Jutho: I think `SimpleUnit` is also semisimple, whereas `GenericUnit` is semisimple but not simple. (We probably want to stay away from categories with a non-semisimple unit alltogether (if that is even a thing).)

```suggestion
For vector spaces where `I = sectortype(S)` has a non-simple unit structure
(`UnitStyle(I) == GenericUnit()`), this errors. 
```

Jutho (approved):

src/spaces/vectorspaces.jl
borisdevos: At least in this context, semisimplicity is specific to the unit, so it's fine this way no? But in general yes, fusion categories are semisimple by definition, namely every object can be written as a direct sum of simple objects. I'm not exactly sure what non-simple means here, though.

Jutho: I think (semi)simplicity refers to the structure of the algebra `Hom(V,V)` with `V` the `unitspace` object. But the simple case is just a special case of semisimple, no?
Screenshot 2026-09-10 at 17 41 46

Issue #516, opened 2026-08-27
Permuting an `AdjointTensorMap` seems much slower than it should be


leburgel:
It seems the in-place `permute!` of an `AdjointTensorMap` with a plain `TensorMap` parent is much slower than it should be compared to the time and allocations it takes to `permute!` a plain `TensorMap`. This issue is specific to the in-place method, since an out-of-place `permute` seems to just unwrap the parent and follow the same path as a `TensorMap` `permute`. It also seems that the slowdown is much worse for `TensorMap`s with non-Abelian sectortypes, becoming increasingly severe with increasing complexity of the symmetry.

I first encountered this on TensorKit.jl  v0.16.5, but it can be reproduced on the current `main` (v0.17.1). Running the following script,

<details>
<summary>Reproducer script</summary>

```julia
using TensorKit, BenchmarkTools
using TensorKit.VectorInterface: One, Zero

p = ((2, 3), (4, 1))

function compare(name, V)
    t   = randn(ComplexF64, V ⊗ V ← V ⊗ V)
    adj = t'         # AdjointTensorMap
    mat = copy(t')   # identical space and contents, but a genuine TensorMap

    Vdst    = TensorKit.permute(space(mat), p)
    dst_adj = similar(mat, Vdst)
    dst_mat = similar(mat, Vdst)

    # NB: permute(::AdjointTensorMap, p) unwraps the adjoint and is therefore NOT affected.
    # The slow path is the in-place one, which is what `tensoradd!`/`blas_contract!` use.
    f_adj() = permute!(dst_adj, adj, p, One(), Zero())
    f_mat() = permute!(dst_mat, mat, p, One(), Zero())
    f_adj(); f_mat()
    @assert convert(Array, dst_adj) ≈ convert(Array, dst_mat)

    b_adj = @benchmark $f_adj() samples = 500 seconds = 10
    b_mat = @benchmark $f_mat() samples = 500 seconds = 10
    println(rpad(name, 14),
            " adjoint ", round(minimum(b_adj).time / 1e6; digits = 4), " ms (", b_adj.allocs, " allocs)",
            "  tensormap ", round(minimum(b_mat).time / 1e6; digits = 4), " ms (", b_mat.allocs, " allocs)",
            "  ratio ", round(minimum(b_adj).time / minimum(b_mat).time; digits = 2), "x")
end

compare("fZ2",    Vect[FermionParity](0 => 8, 1 => 8))
[...]

PR #520, opened 2026-08-27
Don't materialize an `AdjointTensorMap` in `TO.tensoradd!` when `conjA = true`

leburgel:
A second option for a proper fix for #516, which avoids materializing an `AdjointTensorMap` in `TO.tensoradd!` altogether, but rather passes the `conj` flag along to the array-level `tensoradd!`

Basically the same underlying idea as #519, but since I don't exactly know what I'm doing here I'm posting both, mostly just for inspiration.

Some benchmark timings based on the reproducer of #516:

sector type | old | patched | plain TensorMap | old ratio | patched ratio | speedup
-- | -- | -- | -- | -- | -- | --
fℤ₂ ⊠ U(1) | 0.1245 ms | 0.0375 ms | 0.0336 ms | 3.70× | 1.12× | 3.32×
SU(2) | 2.8247 ms | 0.3739 ms | 0.3448 ms | 8.19× | 1.08× | 7.55×
fℤ₂ ⊠ U(1) ⊠ SU(3) | 4.0868 ms | 0.0510 ms | 0.0512 ms | 79.80× | 1.00× | 80.08×

### DIFF
--- a/src/spaces/homspace.jl
+++ b/src/spaces/homspace.jl
@@ -170,6 +170,30 @@ See also [`degeneracystructure`](@ref), [`blockstructure`](@ref).
 subblockstructure(W::HomSpace) = Dictionary(fusiontrees(W), degeneracystructure(W).subblockstructure)
 
 """
+    adjoint_subblockstructure(W::HomSpace) -> Dictionary
+
+Subblock structure of `W'`, expressed as strides into a buffer laid out for `W`.
+
+This mirrors the relation used by `subblock(::AdjointTensorMap, ...)`: the adjoint shares its
+parent's data, with the tree pair swapped and sizes and strides permuted by
+`(domainind..., codomainind...)`. Permuting `adjoint(t)` can therefore read `t`'s own buffer
+directly, instead of materialising an `AdjointTensorMap` and falling back to the uncached
+tree-transformation path.
+
+See also [`subblockstructure`](@ref).
+"""
+function adjoint_subblockstructure(W::HomSpace)
+    N₁, N₂ = numout(W), numin(W)
+    swap = (ntuple(i -> N₁ + i, N₂)..., ntuple(identity, N₁)...)
+    structs = subblockstructure(W)
+    newkeys = map(((f₁, f₂),) -> (f₂, f₁), collect(keys(structs)))
+    newvals = map(collect(values(structs))) do (sz, str, off)
+        return (TupleTools.getindices(sz, swap), TupleTools.getindices(str, swap), off)
+    end
+    return Dictionary(newkeys, newvals)
+end
+
+"""
     fusionblocks(W::HomSpace)
 
 Return the [`FusionTreeBlock`](@ref)s corresponding to all valid fusion channels of a given `HomSpace`,
--- a/src/tensors/indexmanipulations.jl
+++ b/src/tensors/indexmanipulations.jl
@@ -579,6 +579,29 @@ Base.@deprecate(
     return tdst
 end
 
+"""
+    add_conj_transform!(tdst, tsrc, p, transformer, α, β, backend, allocator) -> tdst
+
+Compute `tdst = β * tdst + α * permute(adjoint(tsrc), p)` while reading `tsrc`'s own data
+buffer, without materialising an `AdjointTensorMap`.
+
+`transformer` must come from [`conj_treebraider`](@ref), whose source strides address the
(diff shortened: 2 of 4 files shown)


src/tensors/indexmanipulations.jl
```diff
@@ -587,14 +610,15 @@ function use_threaded_transform(t::AbstractTensorMap, transformer)
 end
 
 function add_transform_kernel!(
-        tdst, tsrc, p, transformer, α, β, backend, allocator, scheduler
+        tdst, tsrc, p, transformer, α, β, backend, allocator, scheduler;
+        conjsrc::Bool = false
```
lkdvos, line 614:
```suggestion
        tdst, tsrc, p, conjsrc, transformer, α, β, backend, allocator, scheduler
```
little bit of a nitpick, but I'd be inclined to just make this a mandatory (positional) argument instead, which is more in line with the remainder of these functions. Obviously doens't matter, just for consistency.

src/tensors/tensoroperations.jl
```diff
@@ -52,9 +52,23 @@ function TO.tensoradd!(
         return C
     end
     if conjA
-        A′ = adjoint(A)
         pA′ = adjointtensorindices(A, _canonicalize(pA, C))
-        permute!(C, A′, pA′, α, β, backend, allocator)
+        if C isa TensorMap && A isa TensorMap
```
lkdvos, line 56: It somehow feels like this specialization might make more sense on the `permute!`-level directly, where we could intercept `TensorMap`-`AdjointTensorMap` combinations directly? That would also immediately catch manual calls to `permute!(tdst, tsrc', ...)` in the same way.

To achieve this, I'm wondering if it makes sense to just have `treebraider(Vdst, Vsrc, p, levels, conj)` as the entrypoint, which then also fixes the caching keys being distinct? There's some subtleties left with this, mostly with respect to other combinations like `permute!(tdst', tsrc, ...)`, which we would then probably want to map to the other case, but that might be reasonable?

src/spaces/homspace.jl
```diff
@@ -169,6 +169,30 @@ See also [`degeneracystructure`](@ref), [`blockstructure`](@ref).
...
+
+This mirrors the relation used by `subblock(::AdjointTensorMap, ...)`: the adjoint shares its
+parent's data, with the tree pair swapped and sizes and strides permuted by
+`(domainind..., codomainind...)`. Permuting `adjoint(t)` can therefore read `t`'s own buffer
+directly, instead of materialising an `AdjointTensorMap` and falling back to the uncached
+tree-transformation path.
+
+See also [`subblockstructure`](@ref).
+"""
+function adjoint_subblockstructure(W::HomSpace)
```
lkdvos, line 185: code-organization wise, would it be easier to make this `subblockstructure(W::HomSpace, conjW::Bool=false)`?

src/spaces/homspace.jl
```diff
@@ -169,6 +169,30 @@ See also [`degeneracystructure`](@ref), [`blockstructure`](@ref).
...
+    N₁, N₂ = numout(W), numin(W)
+    swap = (ntuple(i -> N₁ + i, N₂)..., ntuple(identity, N₁)...)
+    structs = subblockstructure(W)
+    newkeys = map(((f₁, f₂),) -> (f₂, f₁), collect(keys(structs)))
+    newvals = map(collect(values(structs))) do (sz, str, off)
```
lkdvos, line 190: it feels like it should be possible to avoid the `collect` calls here, but I'd have to dig through the Dictionaries interface(s) to see what is possible?

lkdvos (commented):
Overall this looks great, it's definitely the right call to also specialize the adjointtensor, I've just been too lazy and putting it off, hoping that the actual runtime fraction is not too high. I somehow didn't realize that the structure mapping in the current state is actually not that complicated, so this does indeed open up many optimization possibilities.

---

On a larger scale comment, I'm not sure if this is really a route we want to take, but it does look like we could conceivably reduce the cache sizes by simply making the "treetransformers" only store which tokens go to which other tokens, and therefore decouple that from the actual subblockstructure dictionaries that can then fill out which data is where. This might be a bit larger of a refactor though, so I'll see what I can do next week and if this really makes sense

src/tensors/tensoroperations.jl
leburgel: I think the first part of this comment is more similar to what is sketched in #519, as opposed to the `tensoradd!`-level fix that is sketched here.

Issue #522, opened 2026-08-31
Rework planar_trace hack to enable invariant-based hash/== for FusionTreeBlock


lkdvos:
## Context

`FusionTreeBlock` is conceptually fully determined by its outer charges and duality flags: the canonical constructor `FusionTreeBlock{I}(uncoupled, isdual)` deterministically enumerates all compatible tree pairs (sorted coupled sectors, deterministic tree iterators), and all blocks on the main code path (`fusionblocks(::HomSpace)` and the duality/braiding manipulations) are built this way. Ideally, both `hash` and `==` would exploit this invariant and operate on `(uncoupled, isdual)` only, making dictionary/LRU-cache operations on block keys O(1) instead of comparing full `trees` vectors.

## The obstacle

`planar_trace` for non-`UniqueFusion` sectors contains a workaround (marked `TODO: this is a bit of a hack` in `src/fusiontrees/duality_manipulations.jl`) that wraps individual tree pairs in ad-hoc single-pair blocks:

```julia
src = FusionTreeBlock([(f₁, f₂)])
...
src′ = FusionTreeBlock([(f₁′′, f₂′)])
dst′, U′ = transpose(src′, (p₁′, p₂′))
```

These blocks violate the invariant: they share `(uncoupled, isdual)` with each other and with the canonical block, while being genuinely different objects. They also flow into the cached `fstranspose` as keys. As a consequence:

- `==` cannot be defined on `(uncoupled, isdual)` — distinct single-pair blocks would compare equal and the `fstranspose` cache would return wrong results.
- `hash` currently includes the first tree pair (`hash(first(fusiontrees(b)), h)`) rather than just the defining data, specifically to keep these single-pair blocks from all colliding in the cache.

## Proposal

Rework the `planar_trace` implementation so that it no longer constructs non-canonical single-pair blocks (or at least no longer feeds them into the cached `fstranspose` path). Once every `FusionTreeBlock` in circulation is canonical, both `hash` and `==` can be redefined purely in terms of `(uncoupled, isdual)`, making cache lookups on block keys O(1).

Related: #518 introduced the current `hash` implementation.

PR #525, opened 2026-09-02
Add TimerOutputs-based timers for internal kernels

### LINKED ISSUE #523: Support TimerOutputs.jl instrumentation

lkdvos:
Instruments the index manipulation, contraction and factorization kernels with `@timeit_debug` sections, compiled away by default and enabled via `TensorKit.enable_timers!()`.

Section labels carry a category prefix (`symmetry:` / `bookkeeping:` / `alloc:` / `dense:`), and `TensorKit.timer_summary()` aggregates exclusive times into per-category totals that sum to the total measured time. The `@cached` macro times lookups separately from miss-path construction. While timing, `get_num_*_threads()` return 1 so that the (single-task) timer is never touched concurrently; the guard const-folds away when disabled, leaving production paths unchanged (verified via benchmark against main).

Adds TimerOutputs (1.x, for recursive `enable_debug_timings`) as a dependency, a "Profiling and timers" manual page, and tests.


leburgel (approved):
This is pretty amazing.

src/auxiliary/timers.jl
```diff
@@ -0,0 +1,163 @@
...
+    i === nothing && return nothing
+    prefix = Symbol(label[1:prevind(label, i)])
+    return prefix in TIMER_CATEGORIES ? prefix : nothing
+end
+
+# category of the miss-path (construction) section of an `@cached` function
+function _cached_category(fname::Symbol)
+    return fname in (:fsbraid, :fstranspose, :treebraider, :treetransposer) ?
+        "symmetry" : "bookkeeping"
+end
```
Jutho: Might be useful to have this definition closer to where it is used in the `@cached` macro.

src/factorizations/matrixalgebrakit.jl
```diff
@@ -38,13 +40,17 @@ for f! in (
...
     @eval function MAK.$f!(t::AbstractTensorMap, F, alg::AbstractAlgorithm)
         $(f! in (:eig_full!, :eigh_full!) && :(LinearAlgebra.checksquare(t)))
-        foreachblock(t, F...) do _, (tblock, Fblocks...)
-            Fblocks′ = $f!(tblock, Fblocks, alg)
-            # deal with the case where the output is not in-place
-            for (b′, b) in zip(Fblocks′, Fblocks)
-                b === b′ || copy!(b, b′)
+        @timeit_debug GLOBAL_TIMER $(string(f!)) begin
+            foreachblock(t, F...) do _, (tblock, Fblocks...)
+                @timeit_debug GLOBAL_TIMER "dense: lapack" begin
```
Jutho: Should we call this lapack if it can be other drivers? Just `MatrixAlgebraKit`?

src/factorizations/matrixalgebrakit.jl
```diff
@@ -59,11 +65,15 @@ for f! in (
...
     @eval function MAK.$f!(t::AbstractTensorMap, N, alg::AbstractAlgorithm)
         $(f! in (:eig_vals!, :eigh_vals!, :project_hermitian!, :project_antihermitian!, :exponential!) && :(LinearAlgebra.checksquare(t)))
-        foreachblock(t, N) do _, (tblock, Nblock)
-            Nblock′ = $f!(tblock, Nblock, alg)
-            # deal with the case where the output is not the same as the input
-            Nblock === Nblock′ || copy!(Nblock, Nblock′)
-            return nothing
+        @timeit_debug GLOBAL_TIMER $(string(f!)) begin
+            foreachblock(t, N) do _, (tblock, Nblock)
+                @timeit_debug GLOBAL_TIMER "dense: lapack" begin
```
Jutho: Same comment

src/factorizations/matrixalgebrakit.jl
```diff
@@ -72,11 +82,15 @@ end
...
 function MAK.exponential!((τ, t)::Tuple{E, T}, N, alg::AbstractAlgorithm) where {E <: Number, T <: AbstractTensorMap}
     LinearAlgebra.checksquare(t)
-    foreachblock(t, N) do _, (tblock, Nblock)
-        Nblock′ = exponential!((τ, tblock), Nblock, alg)
-        # deal with the case where the output is not the same as the input
-        Nblock === Nblock′ || copy!(Nblock, Nblock′)
-        return nothing
+    @timeit_debug GLOBAL_TIMER "exponential!" begin
+        foreachblock(t, N) do _, (tblock, Nblock)
+            @timeit_debug GLOBAL_TIMER "dense: lapack" begin
```
Jutho: "

src/planar/planaroperations.jl
```diff
@@ -91,20 +91,22 @@ function planartrace!(
...
+        β′ = One()
+        @timeit_debug GLOBAL_TIMER "dense: trace" for (f₁, f₂) in fusiontrees(A)
+            for ((f₁′, f₂′), coeff) in planar_trace((f₁, f₂), (p₁, p₂), (q₁, q₂))
+                TO.tensortrace!(
+                    C[f₁′, f₂′],
+                    A[f₁, f₂], (p₁, p₂), (q₁, q₂), false,
+                    α * coeff, β′,
+                    backend, allocator
+                )
+            end
```
Jutho: Would it not make more sense for the `"dense: trace"` timer to be inside the `for` loop?

src/planar/planaroperations.jl
```diff
@@ -159,35 +161,36 @@ function planarcontract!(
...
 
+        if cindB == codB && oindB == domB
+            B′ = B
+        else
+            B′ = @timeit_debug GLOBAL_TIMER "alloc: buffers" TensorOperations.tensoralloc_add(
+                scalartype(B), B, (cindB, oindB), false, Val(true), allocator
+            )
+            transpose!(B′, B, (cindB, oindB), One(), Zero(), backend, allocator)
+        end
+        mul!(C, A′, B′, α, β)
```
Jutho, line 190: Don't we want a `"dense"` timer around this `mul!` call?

src/tensors/indexmanipulations.jl
```diff
@@ -221,7 +221,9 @@ See also [`permute`](@ref) for creating a new tensor.
     )
     @boundscheck spacecheck_transform(permute, tdst, tsrc, p)
     levels = ntuple(identity, numind(tsrc))
-    return @inbounds braid!(tdst, tsrc, p, levels, α, β, backend, allocator)
+    return @timeit_debug GLOBAL_TIMER "permute!" @inbounds braid!(
```
Jutho: Is it useful to have a separate `permute!` timer entry, if it is anyway directly lowered to `braid!` and that one has its own timer entry?

src/tensors/indexmanipulations.jl
```diff
@@ -307,14 +309,16 @@ See also [`braid`](@ref) for creating a new tensor.
         backend::AbstractBackend = TO.DefaultBackend(), allocator = TO.DefaultAllocator()
     )
     @boundscheck spacecheck_transform(braid, tdst, tsrc, p, levels)
-    if has_array_view(tdst) && has_array_view(tsrc)
-        TO.tensoradd!(tdst[], tsrc[], p, false, α, β, backend, allocator)
-        return tdst
+    @timeit_debug GLOBAL_TIMER "braid!" begin
```
Jutho: Maybe we can label this timer entry `"permute!/braid!"`.

src/tensors/indexmanipulations.jl
```diff
@@ -307,14 +309,16 @@ See also [`braid`](@ref) for creating a new tensor.
...
+        if has_array_view(tdst) && has_array_view(tsrc)
+            TO.tensoradd!(tdst[], tsrc[], p, false, α, β, backend, allocator)
```
Jutho: Do we want a `"dense"` timer around this call?

src/tensors/indexmanipulations.jl
```diff
@@ -383,12 +387,14 @@ end
         backend::AbstractBackend = TO.DefaultBackend(), allocator = TO.DefaultAllocator()
     )
     @boundscheck spacecheck_transform(transpose, tdst, tsrc, p)
-    if has_array_view(tdst) && has_array_view(tsrc)
-        TO.tensoradd!(tdst[], tsrc[], p, false, α, β, backend, allocator)
-        return tdst
+    @timeit_debug GLOBAL_TIMER "transpose!" begin
+        if has_array_view(tdst) && has_array_view(tsrc)
+            TO.tensoradd!(tdst[], tsrc[], p, false, α, β, backend, allocator)
```
Jutho: Same comment

src/tensors/indexmanipulations.jl
Jutho: `add_transform!` below has such a timer around the analoguous code block

src/tensors/indexmanipulations.jl
```diff
@@ -590,23 +598,27 @@ function add_transform_kernel!(
...
-            @inbounds TO.tensoradd!(
-                tdst[f₁′, f₂′], tsrc[f₁, f₂], p, false, α * coeff, β, backend, allocator
-            )
+        @timeit_debug GLOBAL_TIMER "dense: tensoradd" begin
+            taskforeach(fusiontrees(tsrc), ntasks) do (f₁, f₂)
+                (f₁′, f₂′), coeff = transformer((f₁, f₂))
+                @inbounds TO.tensoradd!(
+                    tdst[f₁′, f₂′], tsrc[f₁, f₂], p, false, α * coeff, β, backend, allocator
+                )
+            end
```
Jutho, line 610: Should this timer be inside the `for` loop?

src/tensors/indexmanipulations.jl
```diff
@@ -626,11 +638,13 @@ function add_transform_kernel!(
...
-            StridedView(data_dst, struct_dst...), StridedView(data_src, struct_src...),
-            p, false, α * coeff, β, backend, allocator
-        )
+    @timeit_debug GLOBAL_TIMER "dense: tensoradd" begin
+        taskforeach(transformer.data, ntasks) do (coeff, struct_dst, struct_src)
+            TO.tensoradd!(
+                StridedView(data_dst, struct_dst...), StridedView(data_src, struct_src...),
+                p, false, α * coeff, β, backend, allocator
+            )
+        end
```
Jutho, line 650: Timer inside `taskforeach` block?

src/tensors/tensoroperations.jl
```diff
@@ -325,33 +327,37 @@ function contract!(
     length(pA[2]) == length(pB[1]) ||
         throw(IndexError("number of contracted indices does not match"))
 
-    # find optimal contraction scheme by checking the following options:
-    # - sorting the contracted inds of A or B to avoid permutations
-    # - contracting B with A instead to avoid permutations
-    pA′, pB′, pA″, pB″, pAB′ = _contract_candidates(pA, pB, pAB)
+    @timeit_debug GLOBAL_TIMER "contract!" begin
+        @timeit_debug GLOBAL_TIMER "bookkeeping: planning" begin
```
Jutho: Will be interesting to see if this ever comes up as something non-negligible (and thus, is it worth timing this?).

src/tensors/tensoroperations.jl
```diff
@@ -325,33 +327,37 @@ function contract!(
...
+            else
+                return blas_contract!(C, A, pA″, B, pB″, pAB, α, β, backend, allocator)
+            end
         else
-            return blas_contract!(C, B, reverse(pB″), A, reverse(pA″), pAB′, α, β, backend, allocator)
+            if memcost3 <= memcost4
+                return blas_contract!(C, B, reverse(pB′), A, reverse(pA′), pAB′, α, β, backend, allocator)
+            else
+                return blas_contract!(C, B, reverse(pB″), A, reverse(pA″), pAB′, α, β, backend, allocator)
+            end
```
Jutho, line 358: Are these `blas_contract!` calls inside the scope of the `"planning!"` timer?

src/tensors/tensoroperations.jl
```diff
@@ -446,7 +452,7 @@ function blas_contract!(
     copyC = !TO.isblasdestination(C, ipAB)
 
     if copyC
-        Cnew = TO.tensoralloc_add(TC, C, ipAB, false, Val(true), allocator)
+        Cnew = @timeit_debug GLOBAL_TIMER "alloc: buffers" TO.tensoralloc_add(TC, C, ipAB, false, Val(true), allocator)
         mul!(Cnew, Anew, Bnew)
```
Jutho, line 454: Does this `mul!` call deserve a `"dense: matmul"` timer wrapper?

src/auxiliary/timers.jl
```diff
@@ -0,0 +1,163 @@
+# define the `@timeit_debug` switch explicitly (instead of letting the first macro
+# expansion do it) so that ordinary code can branch on it, see `timers_enabled`
```
Jutho: I don't understand this comment.

src/auxiliary/timers.jl
```diff
@@ -0,0 +1,163 @@
...
+    TensorKit.timers_enabled() -> Bool
+
+Return whether the `@timeit_debug` timer sections of TensorKit are currently compiled in,
+i.e. whether [`enable_timers!`](@ref) has been called.
+
+This is used internally to force serial execution of parallel regions while timing, since
+a `TimerOutput` may only be manipulated from a single task. When timers are disabled this
+check const-folds to `false`, so it has no runtime cost.
+"""
+timers_enabled() = timeit_debug_enabled()
```
Jutho, line 38: Is there a reason to have both `timers_enabled()` and `timeit_debug_enabled()`? Are they the same? Is it because you do not want to redefine a publicly documented function?

Jutho: Ok, after reading the TimerOutputs.jl documentation section on packages/modules, I understand. Could you add a comment to the `timeit_debug_enabled()` function that this is the one that is relied up and being redefined by the TimerOutputs.jl functionality. That is probably what the comment above that I didn't understand was saying, but I found it very unclear.

src/planar/planaroperations.jl
Jutho: I now realise `mul!` has a timer inside. So I guess it depends on how granular we want the call tree to be. I agree it wouldn't make sense to add a timer here, if the timer in `mul!` is outside the for loop and thus encompasses almost the entire function body. A more granular approach could be to have a timer here, and then move the `"dense"` timer to around the inner matmul call in the loop of the `mul!` implementation. But maybe then the overhead of the timers becomes too significant?

src/auxiliary/timers.jl
lkdvos: Done — moved it right next to the `@cached` macro in `caches.jl`.

src/factorizations/matrixalgebrakit.jl
lkdvos: Good point — renamed to `"dense: MatrixAlgebraKit"` (here and in the other two drivers).

src/factorizations/matrixalgebrakit.jl
lkdvos: Done (renamed to `"dense: MatrixAlgebraKit"`).

src/factorizations/matrixalgebrakit.jl
lkdvos: Done.

src/planar/planaroperations.jl
lkdvos: Agreed — moved it inside, around the `tensortrace!` call, so the `planar_trace` coefficient computation is no longer attributed to `dense` (it now lands in the `planartrace!` section's exclusive time, i.e. `:other`).

src/tensors/indexmanipulations.jl
lkdvos: Agreed — removed the `permute!` wrapper and renamed the `braid!` section to `"permute!/braid!"` as suggested below.

src/tensors/indexmanipulations.jl
lkdvos: Done.

src/tensors/indexmanipulations.jl
lkdvos: Done, same as in `braid!`.

src/tensors/tensoroperations.jl
lkdvos: Removed after all — the planning cost was negligible enough that the section wasn't earning its place; it now simply falls into `contract!`'s exclusive time.

src/tensors/tensoroperations.jl
lkdvos: No — the `"bookkeeping: planning"` block closes right after the four `_contract_memcost` evaluations, so the `blas_contract!` calls are inside `"contract!"` but outside `"planning"`.

src/tensors/tensoroperations.jl
lkdvos: It is already covered: `LinearAlgebra.mul!` for `AbstractTensorMap` carries the `"dense: matmul"` section around its block loop, which nests under `"contract!"` at this call site.

src/auxiliary/timers.jl
lkdvos: Rewrote the comment — it now states explicitly that this is the switch `@timeit_debug` checks and that `TimerOutputs.enable_debug_timings` redefines (see also the thread below).

src/tensors/indexmanipulations.jl
lkdvos: Done — added `"dense: tensoradd"` around the fast path here and in `transpose!`.

src/tensors/indexmanipulations.jl
lkdvos: I kept it outside deliberately: while timers are enabled the region runs serially anyway (so the totals are identical), and a section per tree adds ~150ns each, which is significant for abelian-like workloads with many small blocks. The symmetry work inside the loop (the uncached `fsbraid` call via the transformer) carries its own nested section, and `timer_summary` attributes by *exclusive* time, so it is still correctly counted as `symmetry` rather than `dense`.

src/tensors/indexmanipulations.jl
lkdvos: Same reasoning as above — here the loop body is purely dense work (the coefficient is precomputed in the transformer), so the outer wrap measures exactly the dense kernels plus negligible iteration overhead, without paying a timer entry per tree.

src/planar/planaroperations.jl
lkdvos: That was my thinking as well — the `"dense: matmul"` section inside `mul!` nests under `"planarcontract!"` here, so the call tree keeps the context, and its exclusive time only mislabels the (cheap) block-iteration bookkeeping. Moving the section to the per-block `mul!` call would be more precise but pays a timer entry per coupled sector, which starts to matter exactly in the many-small-blocks regime one typically wants to profile. Happy to revisit if the block-loop overhead ever shows up as non-negligible.

src/auxiliary/timers.jl
lkdvos: I've expanded the comments a bit to make this hopefully clearer

src/planar/planaroperations.jl
lkdvos: Yeah, I think that kind of overhead is better measured with an actual profiler rather than the timers like this. I realize the current implementation is not hugely consistent, but I did try and make a judgement call on most of the cases

src/planar/planaroperations.jl
lkdvos: postedit: this should probably be reworked after #522 gets properly handled

Jutho (approved):
Thanks, this looks very useful and the timer structuring decisions all seem very reasonable to me, and can be finetuned further at some later point if need be.

PR #518, opened 2026-08-27
Add `hash` for `FusionTreeBlock` to fix cacheing behind `AdjointTensorMap` `permute`s

leburgel:
Potential first step to addressing #516.

The cache underlying `permute`s of `AdjointTensorMap`s (` GLOBAL_FSBRAID_CACHE`) never gets a single hit, making it effectively useless. This cache uses keys generated from a `FusionTreeBlock`, but `Base.hash` wasn't defined for this type causing literally all keys to miss. Adding the hash improves the performance of permuting `AdjointTensorMap`s with non-Abelian sector types.

Running the reproduction script from #516 gives:

<details>
<summary>Reproduction script</summary>

```julia
using TensorKit, BenchmarkTools
using TensorKit.VectorInterface: One, Zero

p = ((2, 3), (4, 1))

function compare(name, V)
    t   = randn(ComplexF64, V ⊗ V ← V ⊗ V)
    adj = t'         # AdjointTensorMap
    mat = copy(t')   # identical space and contents, but a genuine TensorMap

    Vdst    = TensorKit.permute(space(mat), p)
    dst_adj = similar(mat, Vdst)
    dst_mat = similar(mat, Vdst)

    # NB: permute(::AdjointTensorMap, p) unwraps the adjoint and is therefore NOT affected.
    # The slow path is the in-place one, which is what `tensoradd!`/`blas_contract!` use.
    f_adj() = permute!(dst_adj, adj, p, One(), Zero())
    f_mat() = permute!(dst_mat, mat, p, One(), Zero())
    f_adj(); f_mat()
    @assert convert(Array, dst_adj) ≈ convert(Array, dst_mat)

    b_adj = @benchmark $f_adj() samples = 500 seconds = 10
    b_mat = @benchmark $f_mat() samples = 500 seconds = 10
    println(rpad(name, 14),
            " adjoint ", round(minimum(b_adj).time / 1e6; digits = 4), " ms (", b_adj.allocs, " allocs)",
            "  tensormap ", round(minimum(b_mat).time / 1e6; digits = 4), " ms (", b_mat.allocs, " allocs)",
            "  ratio ", round(minimum(b_adj).time / minimum(b_mat).time; digits = 2), "x")
end

compare("fZ2",    Vect[FermionParity](0 => 8, 1 => 8))
compare("fZ2xU1", Vect[FermionParity ⊠ U1Irrep]((0,0) => 4, (1,1) => 4, (0,2) => 3, (1,-1) => 3))
[...]

PR #526, opened 2026-09-03
Refactor index manipulation kernels around position-indexed subblocks

### LINKED ISSUE #516: Permuting an `AdjointTensorMap` seems much slower than it should be

lkdvos:
Structural fix for #516 (adjoint `permute!` up to 80× slower than the plain path), superseding #519 and #520 and building on #518/#521.

**What changes**
- `StridedSubblocks`: sector-independent, integer-indexed views into the flat data of a `TensorMap` in canonical fusion-tree order, with an optional lazy conjugation (`op = identity`/`conj` as type parameter). `TreeSubblocks` is the generic counterpart for any `AbstractTensorMap`, going through `subblock`.
- `TreeTransformer`s store only the mapping between subblock positions and recoupling coefficients (plus the two subblock structures), and are cached for every tensor type; the closure-based fallback and `TrivialTreeTransformer` are gone.
- One kernel serves abelian and generic transformers, `TensorMap`s and other tensor types.
- `permute!`/`braid!`/`transpose!` and `TO.tensoradd!` fold `AdjointTensorMap` sources *and* destinations (and `conjA`) into a `conjsrc::Bool`, relabeled `p`/`levels` and conjugated `α`/`β`; the flag is resolved into the view type only at the kernel entry, so everything stays type-stable.
- `BraidingTensor` sources are converted before the transformer is chosen (the old overload could reach an untyped kernel).

**Numbers** (issue reproducer, adjoint vs plain `permute!`): fℤ₂ 1.30× → 1.09×, fℤ₂⊠U₁ 3.57× → 1.05×, SU₂ 4.44× → 1.01×, U₁ 2.23× → 1.11×; plain path unchanged within noise (+32 B from the extra `Bool` in the cache key).

**Tests**: adjoint source/destination/both for `permute!`/`transpose!`/`braid!` with accumulation, `@tensor conj`, `BraidingTensor` source, and a `dot`-based isometry check that would catch a wrongly conjugated recoupling matrix for complex sector scalar types. Benchmark suite gained `adjoint = true` permute variants.

Follow-ups (not here): keying transformers on sector structure only; passing `conj` into `trace_permute!`.


src/tensors/blockiterators.jl
```diff
@@ -171,3 +171,73 @@ function Base.show(io::IO, mime::MIME"text/plain", iter::SubblockIterator)
     show_subblocks(io, mime, iter)
     return nothing
 end
+
```
Jutho, line 174: It's probably becoming time to rename this file from "blockiterator.jl" to "tensoriterators.jl" or something of this kind.

Jutho: Or "blockiterators.jl" at least.

src/tensors/indexmanipulations.jl
```diff
@@ -220,8 +220,17 @@ See also [`permute`](@ref) for creating a new tensor.
...
     @boundscheck spacecheck_transform(permute, tdst, tsrc, p)
-    levels = ntuple(identity, numind(tsrc))
-    return @inbounds braid!(tdst, tsrc, p, levels, α, β, backend, allocator)
+    @timeit_debug GLOBAL_TIMER "permute!/braid!" begin
+        if has_array_view(tdst) && has_array_view(tsrc)
+            @timeit_debug GLOBAL_TIMER "dense: tensoradd" TO.tensoradd!(
+                tdst[], tsrc[], p, false, α, β, backend, allocator
+            )
+            return tdst
+        end
```
Jutho: Is this addition useful. Is this superior to unwrapping adjoints and then passing to `tensoradd!` in the corresponding `_braid!` call?

src/tensors/blockiterators.jl
lkdvos: agreed, done!

src/tensors/indexmanipulations.jl
lkdvos: It's not really an addition in the sense that this was already there, it just got moved around a bit. The main thing is that it should go before the `treebraider` or `treetransposer` call to bypass the LRU lookup overhead, which I've now done, and meanwhile refactored this into a reused function so the code got slightly condensed as well.

Sidenote; I realize that I could also bypass the `treebraider` and `treetransposer` functions themselves, but I feel like that introduces quite a lot of code so this might not be worth it

benchmark/TensorKitBenchmarks/indexmanipulations/IndexManipulationBenchmarks.jl
```diff
@@ -34,9 +36,10 @@ function benchmark_permute!(bench; sigmas = nothing, T = "Float64", I = "Trivial
 
     codomain = mapreduce(Base.Fix1(getindex, Vs), ⊗, p_[1]; init = one(eltype(Vs)))
     domain = mapreduce(Base.Fix1(getindex, Vs), ⊗, p_[2]; init = one(eltype(Vs)))
-    init() = init_permute_tensors(T_, codomain ← domain, p_)
+    init() = init_permute_tensors(T_, codomain ← domain, p_, adjoint)
 
-    bench[T, I, dims, sigmas, p] = @benchmarkable permute!(C, A, $p_) setup = ((C, A) = $init())
+    key = adjoint ? (T, I, dims, sigmas, p, "adjoint") : (T, I, dims, sigmas, p)
```
Jutho, line 41: heterogeneous key choice? Does this only affect how the results are displayed?

src/tensors/blockiterators.jl
```diff
@@ -171,3 +171,73 @@ function Base.show(io::IO, mime::MIME"text/plain", iter::SubblockIterator)
...
+const SubblockOp = Union{typeof(identity), typeof(conj)}
+struct StridedSubblocks{A <: DenseVector, N, F <: SubblockOp}
+    data::A
+    structure::Vector{StridedStructure{N}}
+    op::F
+end
+Base.length(s::StridedSubblocks) = length(s.structure)
+Base.firstindex(s::StridedSubblocks) = 1
+Base.lastindex(s::StridedSubblocks) = length(s)
+Base.eltype(::Type{S}) where {S <: StridedSubblocks} = Core.Compiler.return_type(getindex, Tuple{S, Int})
```
Jutho, line 197: Isn't this a bit lazy? Why not:

```suggestion
Base.eltype(::Type{StridedSubblocks{A, N, F}) where {A, N, F} = StridedView{eltype(A), N, A, F}
```

src/tensors/indexmanipulations.jl
```diff
@@ -220,8 +220,11 @@ See also [`permute`](@ref) for creating a new tensor.
         backend::AbstractBackend = TO.DefaultBackend(), allocator = TO.DefaultAllocator()
     )
     @boundscheck spacecheck_transform(permute, tdst, tsrc, p)
-    levels = ntuple(identity, numind(tsrc))
-    return @inbounds braid!(tdst, tsrc, p, levels, α, β, backend, allocator)
+    @timeit_debug GLOBAL_TIMER "permute!/braid!" begin
+        tdst′, tsrc′, p′, _, conjsrc, α′, β′ = unwrap_adjoints(tdst, tsrc, p, nothing, false, α, β)
+        @inbounds _braid!(tdst′, tsrc′, p′, conjsrc, allind(tsrc′), α′, β′, backend, allocator)
+    end
+    return tdst
```
Jutho, line 227: I guess this needs a rebase on the merged TimerOutputs PR?

src/tensors/indexmanipulations.jl
```diff
@@ -388,15 +384,10 @@ end
...
-        if has_array_view(tdst) && has_array_view(tsrc)
-            @timeit_debug GLOBAL_TIMER "dense: tensoradd" TO.tensoradd!(
-                tdst[], tsrc[], p, false, α, β, backend, allocator
-            )
-            return tdst
-        end
-        transformer = treetransposer(tdst, tsrc, p)
-        return @inbounds add_transform!(tdst, tsrc, p, transformer, α, β, backend, allocator)
+        tdst′, tsrc′, p′, _, conjsrc, α′, β′ = unwrap_adjoints(tdst, tsrc, p, nothing, false, α, β)
+        @inbounds _transpose!(tdst′, tsrc′, p′, conjsrc, α′, β′, backend, allocator)
```
Jutho, line 388: Even though there is no primary version of `conjsrc`, I would still like it primed for consistency 😄 .

```suggestion
        tdst′, tsrc′, p′, _, conjsrc′, α′, β′ = unwrap_adjoints(tdst, tsrc, p, nothing, false, α, β)
        @inbounds _transpose!(tdst′, tsrc′, p′, conjsrc′, α′, β′, backend, allocator)
```

src/tensors/indexmanipulations.jl
```diff
@@ -308,17 +311,10 @@ See also [`braid`](@ref) for creating a new tensor.
...
-                tdst[], tsrc[], p, false, α, β, backend, allocator
-            )
-            return tdst
-        end
-        levels1 = TupleTools.getindices(levels, codomainind(tsrc))
-        levels2 = TupleTools.getindices(levels, domainind(tsrc))
-        transformer = treebraider(tdst, tsrc, p, (levels1, levels2))
-        return @inbounds add_transform!(tdst, tsrc, p, transformer, α, β, backend, allocator)
+        tdst′, tsrc′, p′, levels′, conjsrc, α′, β′ = unwrap_adjoints(tdst, tsrc, p, levels, false, α, β)
+        @inbounds _braid!(tdst′, tsrc′, p′, conjsrc, levels′, α′, β′, backend, allocator)
```
Jutho, line 315:
```suggestion
        tdst′, tsrc′, p′, levels′, conjsrc′, α′, β′ = unwrap_adjoints(tdst, tsrc, p, levels, false, α, β)
        @inbounds _braid!(tdst′, tsrc′, p′, conjsrc′, levels′, α′, β′, backend, allocator)
```

src/tensors/indexmanipulations.jl
```diff
@@ -220,8 +220,11 @@ See also [`permute`](@ref) for creating a new tensor.
         backend::AbstractBackend = TO.DefaultBackend(), allocator = TO.DefaultAllocator()
     )
     @boundscheck spacecheck_transform(permute, tdst, tsrc, p)
-    levels = ntuple(identity, numind(tsrc))
-    return @inbounds braid!(tdst, tsrc, p, levels, α, β, backend, allocator)
+    @timeit_debug GLOBAL_TIMER "permute!/braid!" begin
+        tdst′, tsrc′, p′, _, conjsrc, α′, β′ = unwrap_adjoints(tdst, tsrc, p, nothing, false, α, β)
+        @inbounds _braid!(tdst′, tsrc′, p′, conjsrc, allind(tsrc′), α′, β′, backend, allocator)
```
Jutho, line 225:
```suggestion
        tdst′, tsrc′, p′, _, conjsrc′, α′, β′ = unwrap_adjoints(tdst, tsrc, p, nothing, false, α, β)
        @inbounds _braid!(tdst′, tsrc′, p′, conjsrc′, allind(tsrc′), α′, β′, backend, allocator)
```

benchmark/TensorKitBenchmarks/indexmanipulations/IndexManipulationBenchmarks.jl
lkdvos: this is indeed only display, I can change it to a boolean though!
[edit] left it as-is to keep this backwards-compatible, as this makes the keys of previous runs unchanged...

src/tensors/blockiterators.jl
lkdvos: I think the annoying part here is that stridedviews "normalize" their parent arrays, so the `Vector{T}` becomes a `Memory{T}` I think. I'll see if I can figure this out in a better way.

src/tensors/indexmanipulations.jl
lkdvos: I don't think so, why?

src/tensors/blockiterators.jl
lkdvos: edit: fixed by immediately normalizing the parent array of the stridedsubblocks as well

src/tensors/indexmanipulations.jl
Jutho: Never mind, I was confused that there was no timer in the old code, but the call lowering has changed which triggered this new timer entry point.

src/tensors/indexmanipulations.jl
```diff
@@ -562,209 +553,183 @@ Base.@deprecate(
...
-    @timeit_debug GLOBAL_TIMER "alloc: buffers" buffers = [
-        TO.tensoralloc(typeof(data_dst), bufsize, Val(true), allocator)
-            for _ in 1:clamp(length(transformer.data), 1, ntasks)
-    ]
-    taskforeach(transformer.data, buffers) do subtransformer, buffer
-        _add_transform_block!(
-            data_dst, data_src, p, subtransformer, buffer, α, β, backend, allocator
-        )
+    if !conjsrc && p[1] === codomainind(tsrc) && p[2] === domainind(tsrc)
+        add!(tdst, tsrc, α, β)
```
Jutho, line 630: If we consider `add_transform!` a public method, and not just as a helper method for `permute!`, `braid!`, `transpose!`, then I wonder if there could ever be a `transformer` that even assigns a non-trivial transformation to the trivial permutation `p[1] === codomainind(tsrc) && p[2] === domainind(tsrc)`. But I guess that is very much a hypothetical question.

src/tensors/indexmanipulations.jl
```diff
@@ -562,209 +553,183 @@ Base.@deprecate(
...
-        @timeit_debug GLOBAL_TIMER "dense: pack" @inbounds for (i, (f₁, f₂)) in enumerate(fusiontrees(src))
-            TO.tensoradd!(
-                sreshape(view(buffer_src, :, i), sz_src), tsrc[f₁, f₂],
-                ptriv, false, One(), Zero(), backend, allocator
-            )
-        end
+use_threaded_transform(t::TensorMap, transformer) =
+    get_num_transformer_threads() > 1 && length(t.data) > Strided.MINTHREADLENGTH
+use_threaded_transform(t::AbstractTensorMap, transformer) =
+    get_num_transformer_threads() > 1 && dim(space(t)) > Strided.MINTHREADLENGTH
```
Jutho: Is the `get_num_transformer_threads() > 1` check useful? If it is not here, and the second check evaluates true, then the corresponding line `ntasks = use_threaded_transform(tdst, transformer) ? get_num_transformer_threads() : 1` anyway evaluates to one in both the `true` and `false` cases.

lkdvos: Simplified, now just use `get_num_transformer_threads` everywhere since indeed we check for `1`, not for `true/false`.

src/tensors/indexmanipulations.jl
lkdvos: I definitely think you are right, in the sense that for example braiding two legs around eachother fully would fall in that category, but I also don't really think I consider `add_transform` a public method :)

Jutho: That is a good example indeed, and might be important if we ever consider composing fusion tree transformations. If `add_transform!` is not public, I do wonder if we still need the `spacecheck_transform` at the top. This is already called in `permute!/braid!/tranpose!` and then disabled here via `@inbounds`.

If the `@inbounds/@boundscheck` mechanism works correctly, it shouldn't hurt, but it is something that crossed my mind.

src/tensors/blockiterators.jl
Jutho: This sentence is a bit strange and seems broken (but not really part of this PR).

src/tensors/tensor.jl
```diff
@@ -496,6 +496,26 @@ end
...
+    return sectortype(t) === Trivial ? SubblockIterator(t, fusiontrees(t)) :
+        SubblockIterator(t, StridedSubblocks(t))
+end
+function Base.iterate(iter::SubblockIterator{<:TensorMap, <:StridedSubblocks}, i::Int = 1)
+    i > length(iter.structure) && return nothing
+    @inbounds begin
+        f = gettokenvalue(fusiontrees(iter.t), i)
+        return f => iter.structure[i], i + 1
+    end
+end
```
Jutho, line 517: I find this definition a bit strange here, and would have expected it near SubblockIterator`. But I guess it is using internals of both `TensorMap` (not really though), `SubblockIterator` and `StridedSubblocks`, which is maybe an indication that this is not the best design for maintainability. I will first read the rest of the PR though, before casting a final judgement.

src/tensors/tensoroperations.jl
```diff
@@ -39,6 +39,7 @@ has_array_view(t) = has_array_view(typeof(t))
 has_array_view(::Type) = false
 has_array_view(::Type{T}) where {T <: TensorMap} = sectortype(T) === Trivial
 has_array_view(::Type{T}) where {T <: AdjointTensorMap} = has_array_view(parenttype(T))
+has_array_view(t, ts...) = has_array_view(t) && has_array_view(ts...)
```
Jutho: Do we need this, can't we use `all(has_array_view, (t, ts...))` on the places where this is called?

src/tensors/indexmanipulations.jl
lkdvos: alright, removed this, and changed the docstring to a comment to highlight the non-publicness a bit more.

src/tensors/tensor.jl
lkdvos: It is somewhat consistent though, the `BlockIterator` has the same shape, where the "generic" codepaths are in the blockiterators.jl file, while the specialized ones are in the files of the respective tensor types.

The main reason both still exist and I didn't manage to join them is that one is indexed by `Int`, and the other by `FusionTreePair`. In principle I can make the latter indexable by `Int` as well, but the issue is that I can't support indexing by fusiontrees in a way that has no type that depends on the symmetry. This design is chosen specifically to make the kernel code symmetry-independent, such that only a single instance has to be compiled...

src/tensors/indexmanipulations.jl
```diff
@@ -562,209 +553,176 @@ Base.@deprecate(
...
+            _add_transform_block!(dst, src, p, U, inds_dst, inds_src, nothing, α, β, backend, allocator)
+        end
+    else
+        # One max-sized workspace per task (a single one that is reused by all blocks when
+        # serial), allocated on the calling thread before any task spawns, so that also
+        # allocators that are not thread-safe can be used.
+        cp = TO.allocator_checkpoint!(allocator)
+        @timeit_debug GLOBAL_TIMER "alloc: buffers" buffers = [
+            TO.tensoralloc(storagetype(dst), bufsize, Val(true), allocator)
+                for _ in 1:clamp(length(transformer.data), 1, ntasks)
```
Jutho: I don't have the habit of using `clamp` so I always have to look up the syntax. Is the lower bound of `1` necessary and even correct here? Can there be a case where there `transformer.data` has length zero, and we simply don't want to do any work nor allocate any buffers?


```suggestion
                for _ in 1:min(length(transformer.data), ntasks)
```

lkdvos: will include this with the next round of comments/rebase, I think this is a remnant of a previous implementation I was playing around with, but indeed there is no reason to clamp this as `1:0` already clamps, and this should never be reached if the data has zero length anyways

ext/TensorKitEnzymeExt/utility.jl
```diff
@@ -24,6 +24,7 @@ pullback_dC!(ΔC, β::Number) = scale!(ΔC, conj(β))
 @inline EnzymeRules.inactive_type(::Type{<:TensorKit.FusionTreeDict}) = true
 @inline EnzymeRules.inactive_type(::Type{<:TensorKit.FusionTreeBlock}) = true
 @inline EnzymeRules.inactive_type(::Type{<:TensorKit.GenericTreeTransformer}) = true
+@inline EnzymeRules.inactive_type(::Type{<:TensorKit.AbelianTreeTransformer}) = true
```
kshyatt, line 27: whoops

src/tensors/blockiterators.jl
```diff
@@ -171,3 +171,83 @@ function Base.show(io::IO, mime::MIME"text/plain", iter::SubblockIterator)
...
     return nothing
 end
+
+const SubblockOp = Union{typeof(identity), typeof(conj)}
+
+"""
+    struct StridedSubblocks{A <: DenseVector, N, F}
+    StridedSubblocks(t::TensorMap, [op = identity])
+
+Sector-independent, integer-indexable collection of the subblocks of a `TensorMap`, as
```
kshyatt: Should this docstring explain what `N` and `F` are?

src/tensors/treetransformers.jl
```diff
@@ -2,215 +2,202 @@
     TreeTransformer
 
 Supertype for structures containing the data for a tree transformation.
+
+The transformers only store how subblocks map onto each other in terms of their positions in
+[`StridedSubblocks`](@ref) (the canonical order of `fusiontrees`), together with the recoupling
```
kshyatt: should `fusiontrees` get a ref here?

src/tensors/tensor.jl
```diff
@@ -496,6 +496,26 @@ end
 
 # Getting and setting the data at the subblock level
 # --------------------------------------------------
+function StridedSubblocks(t::TensorMap, op::SubblockOp = identity)
+    return StridedSubblocks(t, degeneracystructure(space(t)).subblockstructure, op)
```
Jutho, line 500: I think this is correct?

```suggestion
    return StridedSubblocks(t, subblockstructure(space(t)), op)
```

Jutho: Never mind, it is not. That maps it back into a dictionary with fusion trees as keys.

src/tensors/treetransformers.jl
```diff
@@ -2,215 +2,202 @@
...
+    structure_src::Vector{StridedStructure{N}}
 end
 
-function AbelianTreeTransformer(transform, p, Vdst, Vsrc)
+# Space and permutation as seen by the tree manipulations: for a conjugated source these are
+# those of the adjoint space, whose tree pairs `(f₁, f₂)` address the subblock `(f₂, f₁)` of the source.
+function transform_source(Vsrc::HomSpace, p::Index2Tuple, conjsrc::Bool)
+    return conjsrc ? (adjoint(Vsrc), adjointtensorindices(Vsrc, p)) : (Vsrc, p)
+end
+source_tree((f₁, f₂)::FusionTreePair, conjsrc::Bool) = conjsrc ? (f₂, f₁) : (f₁, f₂)
```
lkdvos: inline these functions

src/tensors/treetransformers.jl
```diff
@@ -2,215 +2,202 @@
...
+function AbelianTreeTransformer(transform, p, Vdst, Vsrc, conjsrc::Bool)
     t₀ = Base.time()
-    permute(Vsrc, p) == Vdst || throw(SpaceMismatch("Incompatible spaces for permuting."))
-    fts_src = subblockstructure(Vsrc)
-    fts_dst = subblockstructure(Vdst)
-    L = length(fts_src)
+    Vsrc′, p′ = transform_source(Vsrc, p, conjsrc)
+    spacecheck_transform(permute, Vdst, Vsrc′, p′)
+    srcindices = fusiontreeindices(Vsrc)
+    dstindices = fusiontreeindices(Vdst)
```
lkdvos: look at if we can simplify `fusiontreeindices` and if that can be removed. Note `enumerate` gives same order!

lkdvos:
All comments should now be addressed again, I think I also managed to simplify slightly more because of @Jutho's comments, in particular the abelian case now no longer needs to refer to the "adjoint spaces" and simply loops over the source fusiontrees and exchanges them on the fly, which at least reads a lot easier.

For the non-abelian case I still am doing it that way because swapping splitting and fusiontree in the `FusionBlock` requires rebuilding the block array and then sorting it (in order to have a unique cached value for the unitary generated there), so there's a lot more subtle annoyances to get rid of this. I think in this case that is just complexity that is required... :(

src/tensors/tensor.jl
Jutho: I guess I am wondering to what extent this specialization is necessary. As far as I can tell, `subblocks` is not really used in any performance critical code, they all go via `StridedSubblocks` or `TreeSubblocks` directly. The `subblocks` function and the associated `SubblockIterator` seem mostly user convenience functions, for getting and setting data in the tensor. So I don't know if we really need to make this more complicated for negligible performance gain>

src/tensors/treetransformers.jl
```diff
@@ -2,215 +2,201 @@
...
 
-const AbelianTransformerData{T, N} = Tuple{T, StridedStructure{N}, StridedStructure{N}}
+"""
+    AbelianTreeTransformer{T, N} <: TreeTransformer
 
+Tree transformation for `UniqueFusion` sectors, where every source subblock maps onto a single
+destination subblock with a scalar coefficient, stored as `(coeff, idst, isrc)`. The subblock
+structures of the destination and source spaces are kept alongside, such that the
+[`StridedSubblocks`](@ref) of both tensors can be created without further lookups.
+"""
```
Jutho, line 25: Thanks for the very clear explanation/motivation.

src/tensors/treetransformers.jl
```diff
@@ -2,215 +2,201 @@
...
 end
 
-
 function treetransformertype(Vdst, Vsrc)
     I = sectortype(Vdst)
-    I === Trivial && return TrivialTreeTransformer
-
     T = sectorscalartype(I)
     N = numind(Vdst)
     return FusionStyle(I) == UniqueFusion() ? AbelianTreeTransformer{T, N} : GenericTreeTransformer{T, N}
```
Jutho, line 152: Like with the `FusionStyle`, the meaning of abelian can be interpreted in two different ways. It can refer to the fusion of two sectors having a unique sector as result (true for Vect[G] for any group G, or Rep[G] for abelian G), or it can refer to the commutativity of the fusion rules. (true for Rep[G] for any G and some other categories, or for Vect[G] for abelian G only). Should they be called `UniqueTreeTransformer`?

src/tensors/indexmanipulations.jl
```diff
@@ -562,209 +560,170 @@ Base.@deprecate(
...
         buffer_src = StridedView(buffer, (blocksize, cols), (1, blocksize), blocksize * rows)
 
         # 1. Extract: copy each source block into column i of buffer_src as a flat vector,
         #    using a trivial permutation so the layout is canonical before the matmul.
-        @timeit_debug GLOBAL_TIMER "dense: pack" @inbounds for (i, struct_src_i) in enumerate(structs_src)
+        @timeit_debug GLOBAL_TIMER "dense: pack" @inbounds for (i, isrc) in enumerate(inds_src)
             TO.tensoradd!(
-                sreshape(view(buffer_src, :, i), sz_src), StridedView(data_src, sz_src, struct_src_i...),
+                sreshape(view(buffer_src, :, i), sz_src), src[isrc],
                 ptriv, false, One(), Zero(), backend, allocator
```
Jutho, line 711: I think I finally manage to grasp the full logic in my mind. Would it make sense to not have the `conj` as an `op` in the `TrasnformSubblocks` and just use the argument in this `tensoradd!` call, i.e. `false -> conjsrc` for argument 4 here, or in line 696? I think that is all that is required (which of course requires to pass the boolean flag `conjsrc` through `add_transform_kernel!` into `_add_transform_block!`, but would still reduce or postpone the compilation of different methods.

src/tensors/tensor.jl
Jutho, line 517: Reposting as a new comment, so that it does not get lost:

I guess I am wondering to what extent this specialization is necessary. As far as I can tell, subblocks is not really used in any performance critical code, they all go via StridedSubblocks or TreeSubblocks directly. The subblocks function and the associated SubblockIterator seem mostly user convenience functions, for getting and setting data in the tensor. So I don't know if we really need to make this more complicated for negligible performance gain.

Jutho:
Ok, I finally managed to make my way through. Looks really great. I have four final questions or suggestions, but otherwise fully approve.

lkdvos:
Thanks for the careful review, I think all of these comments were helpful to further simplify and improve the implementation, which I've hopefully managed to achieve here.

In particular, your comment about the `SubblockIterator` specialization made me realize that the `TreeSubblocks` were functionally the same structure as that, and so I managed to merge these implementations, just adding the "token-based" access to the original. This has further reduced some code at otherwise zero additional cost, which is always nice 😄.

As for whether the TensorMap specialization is still worth it, I would argue yes, mostly because there are still other parts of the code that iterate fusiontrees and look at the data, such as `flip` and `twist`, and while I don't want to delay this PR further, it would definitely be useful to rewrite these in terms of the `subblocks` as well, further bypassing some symmetry overhead.

I'm hoping all tests pass, and if so might merge this Monday morning (ET) if no further comments, or unless someone merges this first.

src/tensors/tensor.jl
lkdvos: To address this more precisely, this is almost true, in the sense that for example operations that mix diagonal tensors and regular tensors would still end up with `subblocks` being called, and I do see future work benefiting from the knowledge that `subblocks` is a performant primitive to build around, e.g. for `twist` or `flip` implementations.

src/tensors/tensor.jl
```diff
@@ -496,32 +496,36 @@ end
...
-# iterate the subblock views in canonical order alongside the fusion trees, without hashing
-function subblocks(t::TensorMap)
-    return sectortype(t) === Trivial ? SubblockIterator(t, fusiontrees(t)) :
-        SubblockIterator(t, StridedSubblocks(t))
+# the throws are kept out of line: mentioning `iter` in the error would make it escape,
+# which costs the caller scalar replacement of the subblock views
+@noinline _throw_subblock_bounds(iter, i) = throw(BoundsError(iter, i))
+@noinline _throw_subblock_missing(f) = throw(SectorMismatch(lazy"fusion tree pair $f is not present"))
+
+@propagate_inbounds function Base.getindex(iter::SubblockIterator{<:TensorMap}, i::Int)
```
Jutho, line 511: This is now relying on `iter` being constructed as `SubblockIterator(t)`. What happens if someone constructs `SubblockIterator(t::TensorMap, fusiontrees(t))`. This method is still being called and will fail, no?

lkdvos: Yes, although that is also true for someone calling `SubblockIterator(t::AbstractTensorMap, nothing)`, in the sense that you really should not be doing that? I can try and restrict the method even further, but since this should all be internals I don't know if that makes too much sense 😄

Jutho: Ok, fair enough.

Jutho (approved):

Jutho:
By the way, did you rerun the benchmarks from the original issue with this latest iteration of the PR?

lkdvos:
## Issue #516 reproducer, rerun against the final state of this PR

`main` @ 7a187c42 vs `ld-adjoint` @ 6b2dd79f, rusty `rome` node, julia 1.12.7, single-threaded, `@benchmark samples=500 seconds=10`. The ratio is in-place adjoint `permute!` over plain `TensorMap` `permute!` *within the same revision*, so 1.00 means an adjoint source costs nothing extra.

| sector type | ratio on `main` | ratio here | adjoint allocs, `main` → here |
|:--|--:|--:|--:|
| `fℤ₂` | 1.31× | **1.09×** | 134 → 68 |
| `fℤ₂ ⊠ U₁` | 3.82× | **1.01×** | 675 → 320 |
| `U₁` | 2.01× | **1.11×** | 299 → 145 |
| `SU₂` | 3.92× | **1.01×** | 7096 → 1926 |
| `fℤ₂ ⊠ U₁ ⊠ SU₃` | 19.06× | **0.99×** | 7873 → 486 |

In every case the adjoint path now allocates exactly what the plain path allocates and runs at the same speed. The plain path itself did not move (`SU₂` 0.183 → 0.187 ms, `fℤ₂ ⊠ U₁ ⊠ SU₃` 0.0296 → 0.0301 ms), so this is the adjoint side coming down to the baseline rather than the baseline drifting up. In absolute terms the adjoint column goes `SU₂` 0.718 → 0.189 ms and `fℤ₂ ⊠ U₁ ⊠ SU₃` 0.564 → 0.030 ms.

Two notes on the baseline: `main` already carries #518 and #521, so the comparison is against the post-hash-fix numbers from the issue thread, not the original 82×. And the `U₁` row is not in the issue script — I added it with a guessed space so the `U₁` figure quoted in the PR description has a counterpart.

Full suite (`linalg` + `indexmanipulations` + `tensornetworks`) on both revisions is still running; I'll post the table when it lands.

PR #529, opened 2026-09-10
 Insert allocator checkpoints for `@planar` calls

leburgel:
`@planar` blocks were missing allocator checkpoint and resets. For buffer allocators this means they just moved the offset until the buffer is full, rendering the allocator effectively useless.


Jutho (approved):
Looks good.

lkdvos (approved):

PR #532, opened 2026-09-11
Fix planar index reordering and remove `_contractedspace`

lkdvos:
Fixes the root cause that #531 works around, and supersedes it.

### Problem

`@planar` allocates through `TensorOperations.tensoralloc_contract` with the raw index tuples of the planar decomposition, in which `pA` and `pB` need not be planar partitions by themselves — only `planarcontract!` rotated them into shape, and only when `pAB` could be absorbed into those rotations. Since #515 added the coloring checks to the `ProductSpace`/`HomSpace` constructors, computing the output space as `permute(compose(permute(VA, pA), permute(VB, pB)), pAB)` throws for `GenericUnit` sectors, which is why #515 replaced it by the leg-wise `_contractedspace`. That workaround pins both operands to a single spacetype, which breaks contracting operands whose `spacetype` parameters differ (a `BlockTensorMap` with a `TensorMap`, as reported from MPSKit in #531); the fallback added there fixes the `MethodError` but reintroduces exactly the intermediate spaces that #515 removed.

### Changes

- `planarcontract!` takes an arbitrary cyclic `pAB`, like the non-planar `contract!`: a residual permutation is applied with `transpose!` through an intermediate, mirroring what `blas_contract!` does with `permute!`. The remaining difference between planar and non-planar contraction is `transpose` vs `permute`, which is also what the AD rules of #124 need.
- `reorder_indices` is replaced by `planar_contract_indices(A, pA, B, pB, pAB) -> pA′, pB′, pAB′`: it rotates each operand's index cycle so the contracted indices form a contiguous arc and remaps `pAB` instead of absorbing it. It is shared with the two `BraidingTensor` methods and works on `HomSpace`s as well as on tensors.
- `@planar` canonicalizes the index tuples it emits at macro-expansion time: `planar_contract_indices` only needs the index partitions of both operands, and those are written in the expression and already checked against the actual tensors by `_extract_tensormap_objects`. Both the contraction and its allocation therefore see planar partitions, and the allocation needs no planar-specific entry point. Hand-written planar kernels should canonicalize with `planar_contract_indices` before allocating with `tensoralloc_contract`; `planarcontract!` keeps canonicalizing at runtime, so the kernel itself is safe either way.
- `⊗` allocates its destination directly; its `pA = ((all indices), ())` intermediate is non-planar by construction.
- `tensorcontract_structure` composes the intermediate spaces again and `_contractedspace` is gone, which restores mixed `spacetype` support. For `GenericUnit` sectors, `pA` and `pB` passed to it must now be planar partitions.

### Tests

New multifusion coverage (there was no `@planar` binary contraction on multifusion spaces): a planar contraction whose partitions are non-planar by themselves, `⊗` with more than one index in the domain, a non-trivial and a non-cyclic `pAB`, `planar_contract_indices` unit tests including the fully contracted and non-planar cases, and a check on the index tuples of the macro expansion.

Verified on the full space list for `test/tensors/{contractions,braidingtensor,planar,linalg,indexmanipulations}.jl` and `test/symmetries/spaces.jl`. The mixed-spacetype case has no in-repo coverage (it needs a downstream space type), so it is worth a run against BlockTensorKit + MPSKit before merging.


docs/src/man/contractions.md
```diff
@@ -289,6 +289,10 @@ Finally, the name `τ` is reserved for the braiding tensor: every literal crossi
 The `BraidingTensor` itself does not need to be constructed by the user; the macro figures out the appropriate spaces from the surrounding contraction.
 Any layout the macro cannot identify as planar is rejected at parse time with `ArgumentError("not a planar diagram expression: ...")`.
 
+For users writing their own planar kernels, it is important to note that the index tuples `pA = (oindA, cindA)` and `pB = (cindB, oindB)` handed to `planarcontract!` need not individually be planar partitions of the operands, even when the overall diagram is planar; only a cyclic rotation of them is.
```
Jutho: The pattern and structure of this sentence, namely "A is not generally true; only B is" is now littering several of the recent preprints I tried to read 😄 . I take less offence to it in this particular instance though.

src/planar/planaroperations.jl
```diff
@@ -179,76 +175,114 @@ function planarcontract!(
...
-    end
-    for j in 2:N₂
-        @assert indB[end + 1 - j] == oindB2[j]
+# whether a contraction with `N₁` open indices on `A` directly yields the destination
+function _isdirectoutput(pAB::Index2Tuple, N₁::Int)
+    return length(pAB[1]) == N₁ && pAB == TO.trivialpermutation(pAB)
+end
+
+# rotate `t` such that `x` comes first
+_rotate_to(t::Tuple, x) = TupleTools.circshift(t, 1 - something(findfirst(==(x), t)))
```
Jutho, line 213: It seems that `Base.circshift` also works on tuples and actually outputs more efficient code (at the llvm/native level).

src/planar/planaroperations.jl
```diff
@@ -179,76 +175,114 @@ function planarcontract!(
...
+    # if all indices are contracted, fix the residual rotation using the other tensor
+    if isempty(oindA′) && !isempty(cindA)
+        cindA′ = _rotate_to(cindA′, cindA[something(findfirst(==(first(cindB′)), cindB))])
     end
-    if isempty(oindB2) && !isempty(cindB)
-        hit = cindB[findfirst(==(first(cindA2)), cindA)]
-        while hit != first(cindB2)
-            cindB2 = _cyclicpermute(cindB2)
-        end
+    if isempty(oindB′) && !isempty(cindB)
```
Jutho: Could this be an `elseif`. I think it is still fine currently. If both `oindA` and `oindB` are empty, then `cindB′ == indB`, `cindA′` will be rotated to match this order, and `cindB[something(findfirst(==(first(cindA′)), cindA))]` should correspond to the first index value in `cindB′`, so no further rotation is happening.

src/planar/planaroperations.jl
lkdvos: This got introduced in Julia 1.12 apparently, just checked this. I will update the TupleTools implementation to link to Julia's on the newer versions.

(https://github.com/Jutho/TupleTools.jl/pull/25)

docs/src/man/contractions.md
lkdvos: Reworded it anyways, being non-offensive is of utmost importance 😄

src/planar/macros.jl
```diff
@@ -91,6 +94,7 @@ function planarparser(planarexpr, kwargs...)
     parser.contractioncostcheck = nothing
     push!(parser.preprocessors, ex -> _check_planarity(ex))
     push!(parser.preprocessors, ex -> _decompose_planar_contractions(ex, temporaries))
+    push!(parser.preprocessors, ex -> _index_partitions!(partitions, ex))
```
Jutho, line 97: For consistency, can we let this preprocessor start with a verb, i.e.

`_index_partitions` -> `_record_index_partitions!`

For further consistency, I would also like `ex` to be the first argument of the preprocessor. 

It also seems I was not consistently using the exclamation mark convention in these preprocessor names, as `_decompose_planar_contractions` does not have an exclamation mark. Here, consistency would require either removing the exlamation mark in `_record_index_partitions` or adding one to `_decompose_planar_contractions`.

test/tensors/contractions.jl
```diff
@@ -57,6 +57,25 @@ for V in spacelist
...
+            t1 = transpose(t, ((1,), (4, 5, 3, 2)))
+            t2 = transpose(t', ((1, 2, 5, 4), (3,)))
+            @test ρ ≈ t1 * t2
+            if BraidingStyle(I) isa Bosonic
+                # `@planar` and `@tensor` only agree for bosonic braiding
+                @tensor ρ2[a; b] := t[a c d; e f] * conj(t[b c d; e f])
+                @test ρ ≈ ρ2
+            end
+            # the intermediate result is allocated as a temporary
+            @planar ρ3[a; b] := t[a c d; e f] * t'[e f; g c d] * ρ[g; b]
```
Jutho, line 76: How is this contraction decomposed. Is think it is left to right, so `(t[a c d; e f] * t'[e f; g c d]) * ρ[g; b]? Would it make sense to have



```suggestion
            @planar ρ3[a; b] := t[a c d; e f] * (t'[e f; g c d] * ρ[g; b])
```

To force a somewhat less trivial temporary?

Jutho (approved):
Up to small suggestions, very much approved.

PR #533, opened 2026-09-15
KA kernel for add_transform for Abelian symmetries

kshyatt:
Split off the simpler Abelian component from https://github.com/QuantumKitHub/TensorKit.jl/pull/509. I've kept the device cache logic for now as it should still apply in this case (if I remember 3 weeks ago correctly...)


lkdvos:
Update here: I'm trying to work out a way to make the whole `TreeTransformer` that ends up being cached also depend on the storagetype. This has the added benefit that we can store complex unitaries for complex tensors, therefore enabling this to go through BLAS, while also having the ability to return custom types for host vs device implementations. I'm hoping this will make this PR consist only of the actual kernel implementation

PR #536, opened 2026-09-16
Docs: add an example page on constructing physical (charged) operators

ZongYongyue:
## Summary

This PR adds a new appendix page, *An example for constructing physical operators*, which works through the construction of the electron creation operator as a symmetric `TensorMap` for three common symmetry choices:

1. U(1) × SU(2) 
2. U(1) × U(1) 
3.  ℤ₂ × SU(2) 

This page complements the manual section on [constructing tensors](https://quantumkithub.github.io/TensorKit.jl/latest/man/tensors/#s_tensors) by showing how the block data of a physically meaningful operator is derived. These derivations were worked out while I was writing my PhD thesis, and I thought they were worth sharing here. As a TensorKit user who has benefited a lot from this package, I hope this example can be helpful to other users as well.


Yue-Zhengyuan:
This may be better suited for https://github.com/QuantumKitHub/TensorKitTensors.jl. One thing we didn't do there is constructing a single creation operator (they always come in pairs there).

ZongYongyue:
@Yue-Zhengyuan  Here I want to give a small case to show what a symmetric tensor looks like specifically after one read the section [Constructing tensors and the TensorMap type](https://quantumkithub.github.io/TensorKit.jl/latest/man/tensors/#s_tensors). The section gives the general form 
截屏2026-09-20 12 38 07
and I want to give a case after it, like this:
```math
C^{\dagger}_{a;(b,e)}
=
\left(I_{D_a}\otimes X_c^a\right)
\circ
\left(P^c_{a,(b,e)}\otimes I_{V_c}\right)
\circ
\left(I_{D_b\otimes D_e}\otimes X_c^{be}\right)^{\dagger},
```

```math
C^\dagger_{\mathbb Z_2\times\mathrm{SU(2)}}
=\begin{pmatrix}1&0\end{pmatrix}\otimes
\left(X_{(-,\frac12)}^{(+,0),(-,\frac12)}\right)^\dagger
+\begin{pmatrix}0\\ \sqrt{2}\end{pmatrix}\otimes
\left(X_{(+,0)}^{(-,\frac12),(-,\frac12)}\right)^\dagger.
```
The motivation is when the first time I have read this section [Constructing tensors and the TensorMap type], I can't wait to seek a case in physics that I'm familiar with, which can be written in this form. What I do here is starting from the reading experience of the readers, maybe not all of them have a tensor background at the beginning, just like me, so you can see I did not just give the last forms of the three kinds of creation operators, but give a whole process of thinking. 

In short, leaving this PR open or close is both ok for me and one who want to see it will always find it.

docs/src/appendix/chargedoperators.md
```diff
@@ -0,0 +1,594 @@
...
+There, the data itself was random, read off from a given dense array, or fixed by simple eigenvalue considerations; what was left open is how the block data of a physically meaningful operator should be determined in the first place.
+This page works through such a case: the construction of fermionic creation operators, the elementary building blocks of many-body Hamiltonians such as the Hubbard model, as symmetric tensor maps.
+
+The example is instructive because a creation operator is not symmetric in the naive sense: acting with it on a state changes the quantum numbers, so it cannot be represented as a symmetric map from the local Hilbert space to itself.
+Instead, it is an instance of a *charged operator*, a symmetric tensor with an additional incoming leg that supplies the quantum numbers of the added particle.
+We first explain in general how the fusion rules and Clebsch–Gordan coefficients of the symmetry group, together with the physical action of the operator, completely fix the block data of such an operator, and then carry out the construction explicitly for three commonly encountered symmetry choices:
```
kshyatt, line 13: Might be nice to mention real models where these combinations are found?

kshyatt:
Hard to review this just looking at the LaTeX so hopefully the doc build will work!

